PR #551, opened 2020-08-26
Remove dependabot and add a check-dependencies command in make file

glours:
Signed-off-by: Guillaume Lours <guillaume.lours@docker.com>

**What I did**
Because Dependabot did manage correctly `go.mod` and `go.sum` files, we decided to remove it and add a simple command in Makefile.

Now use `make check-dependencies` to get the list of the dependencies to update

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,9 @@ lint: ## run linter(s)
 	--build-arg GIT_TAG=$(GIT_TAG) \
 	--target lint
 
+check-dependencies: ## check dependency updates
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -75,6 +75,9 @@ lint: ## run linter(s)
 	--build-arg GIT_TAG=$(GIT_TAG) \
 	--target lint
 
+check-dependencies: ## check dependency updates
+	go list -u -m -f '{{if not .Indirect}}{{if .Update}}{{.}}{{end}}{{end}}' all
```
crazy-max, line 79: @glours You might be interested by https://github.com/docker/buildx/pull/758

PR #622, opened 2020-09-15
Introduce support for external EFS volumes

ndeloof:
**What I did**
Introduce support for EFS volumes set as extrernal volumes:

```yaml
services:
  foo:
    volumes:
      - "data:/tmp"

volumes:
  data:
    external: true
    name: "fs-1234abcd"
```

**Related issue**
https://github.com/docker/ecs-plugin/issues/254
https://github.com/docker/ecs-plugin/issues/220

image

### DIFF
--- a/ecs/cloudformation.go
+++ b/ecs/cloudformation.go
@@ -56,6 +56,47 @@ func (b *ecsAPIService) Convert(ctx context.Context, project *types.Project) ([]
 	if err != nil {
 		return nil, err
 	}
+
+	// Create a NFS inbound rule on each mount target for volumes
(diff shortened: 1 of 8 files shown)


christophluehr:
Quick note, volumes ca have dashes in their names, with ECS/Cloudformation this leads to errors (if someone else runs into this):

```
    volumes:
      - data-vol:/data

volumes:
  data-vol:
    external: true
    name: "fs-xxxxxx"
```

ValidationError: Template format error: Resource name frontendNFSMountdata-volume is non alphanumeric.
        status code: 400, request id: fb3c1-xxxxxxxx

PR #638, opened 2020-09-22
ECS conversion code cleanup and refactoring

ndeloof:
**What I did**
Refactored and cleanup ECS conversion code for better readability and maintainability

/test-ecs

image

### DIFF
--- a/ecs/awsResources.go
+++ b/ecs/awsResources.go
@@ -1,0 +1,278 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package ecs
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/awslabs/goformation/v4/cloudformation/ec2"
+	"github.com/awslabs/goformation/v4/cloudformation/elasticloadbalancingv2"
+
+	"github.com/awslabs/goformation/v4/cloudformation"
+	"github.com/awslabs/goformation/v4/cloudformation/ecs"
+	"github.com/compose-spec/compose-go/types"
+	"github.com/sirupsen/logrus"
+)
+
+// awsResources hold the AWS component being used or created to support services definition
+type awsResources struct {
+	sdk              sdk
+	vpc              string
+	subnets          []string
+	cluster          string
+	loadBalancer     string
+	loadBalancerType string
+	securityGroups   map[string]string
+}
+
+func (r *awsResources) serviceSecurityGroups(service types.ServiceConfig) []string {
+	var groups []string
+	for net := range service.Networks {
+		groups = append(groups, r.securityGroups[net])
+	}
+	return groups
+}
+
+func (r *awsResources) allSecurityGroups() []string {
+	var securityGroups []string
+	for _, r := range r.securityGroups {
+		securityGroups = append(securityGroups, r)
+	}
+	return securityGroups
+}
+
+// parse look into compose project for configured resource to use, and check they are valid
+func (r *awsResources) parse(ctx context.Context, project *types.Project) error {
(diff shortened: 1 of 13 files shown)


ecs/awsResources.go
```diff
@@ -0,0 +1,278 @@
...
+
+	"github.com/awslabs/goformation/v4/cloudformation"
+	"github.com/awslabs/goformation/v4/cloudformation/ecs"
+	"github.com/compose-spec/compose-go/types"
+	"github.com/sirupsen/logrus"
+)
+
+// awsResources hold the AWS component being used or created to support services definition
+type awsResources struct {
+	sdk              sdk
```
gtardif, line 35: maybe a detail, but it might be nice to separate sdk from all other fields here (have an sdk field, and another field holding all AWS resources that are populated by `awsResources` parsing methods.) 
not required, but my first impression is it would regroup things in a clearer way, and maybe make it more obvious to add unit tests, or see at a glance what structs we need to populate. Sdk seems an outsider here

ecs/cloudformation.go
```diff
@@ -34,22 +34,21 @@ import (
...
 )
 
 func (b *ecsAPIService) Convert(ctx context.Context, project *types.Project) ([]byte, error) {
-	template, networks, err := b.convert(project)
+	err := b.checkCompatibility(project)
+	if err != nil {
+		return nil, err
+	}
+
+	err = b.resources.parse(ctx, project)
```
gtardif, line 46: the sequence feels a bit strange here, in basic situation I would expect to get a result from `b.resource.parse()`, and then pass this result to `b.convert()` to get the template. I suppose this was done this way because `convert()` is using some things from the compose project directly, and other from the awsResource created in the `parse()` execution. 
Would it make sense still to make this a bit more explicit with something like 
```
resources, err = b.resources.parse(ctx, project)
if err != nil {
	return nil, err
}

template, err := b.convert(project, resources)
```

or maybe a better way would be to call `b.resource.parse()` as a first step inside `b.parse(project)` rather than requiring both methods to be called in the right order, to limit the statefullness of this

ecs/cloudformation.go
```diff
@@ -103,103 +73,31 @@ func (b *ecsAPIService) Convert(ctx context.Context, project *types.Project) ([]
...
-			logrus.Error(err.Error())
-		} else {
-			logrus.Warn(err.Error())
-		}
-	}
-	if !compatibility.IsCompatible(checker) {
-		return nil, nil, fmt.Errorf("compose file is incompatible with Amazon ECS")
-	}
-
+func (b *ecsAPIService) convert(project *types.Project) (*cloudformation.Template, error) { //nolint:gocyclo
```
gtardif, line 76: any chance to remove the //nolint now this has beed broken into smaller pieces ?

ecs/cloudformation.go
```diff
@@ -293,10 +184,46 @@ func (b *ecsAPIService) convert(project *types.Project) (*cloudformation.Templat
...
 		}
 	}
-	return template, networks, nil
+	return template, nil
+}
+
+func (b *ecsAPIService) createIngress(service types.ServiceConfig, net string, port types.ServicePortConfig, template *cloudformation.Template) {
+	protocol := strings.ToUpper(port.Protocol)
+	if protocol == "" {
+		protocol = "-1"
```
gtardif, line 193: some strange AWS required defaults ?? :) 
a constant with an explicit name might make this less obscure

ecs/awsResources.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/awslabs/goformation/v4/cloudformation/elasticloadbalancingv2"
+
+	"github.com/awslabs/goformation/v4/cloudformation"
+	"github.com/awslabs/goformation/v4/cloudformation/ecs"
+	"github.com/compose-spec/compose-go/types"
+	"github.com/sirupsen/logrus"
+)
+
+// awsResources hold the AWS component being used or created to support services definition
+type awsResources struct {
```
gtardif, line 34: This whole parsing class is asking for unit tests, or is this all covered by slightly higher level tests in cloudformation_tests.go? 
(if required, maybe adding a ticket to do this as a follow up PR if these tests didn't exist before, and not make the refactoring PR too large)

ecs/cloudformation.go
ndeloof: `parse` indeed populates `awsResources` from compose file (using x-aws-*)
then `ensure` creates all the infrastructure resources
and eventyally `convert` create the higher level components from the compose model

I'll revisit this, I agree `parse` to return awsResources would be a more logical approach

ecs/awsResources.go
ndeloof: parse does not just parse extension, it also uses AWS sdk to check valid IDs, so not so simple to write unit tests

gtardif (approved):

PR #640, opened 2020-09-23
Add gRPC layer to volumes management

ulyssessouza:
**Related issue**
Resolves #604 

**(not mandatory) A picture of a cute animal, if possible in relation with what you did**
remote

Issue #604, opened 2020-09-10
volume gRPC APIs


gtardif:
make volume APIs available in gRPC, after https://github.com/docker/compose-cli/pull/593

PR #683, opened 2020-09-28
Volumes grpc interface

gtardif:
**What I did**
Replace #640 (including commit from @ulyssessouza) to follow up with things we did pair programming and fixes in metrics)

**Related issue**
https://github.com/docker/compose-cli/issues/604


gtardif:
CI workflows need to be fixed to include build happening on fork ([for this PR](https://github.com/gtardif/compose-cli/tree/volumes-grpc))

gtardif:
This will require a follow up PR in node-sdk to add volumes API, and add a volume sample (already tested, working on ACI context)

cli/cmd/serve.go
```diff
@@ -64,6 +65,7 @@ func runServe(ctx context.Context, opts serveOpts) error {
 	p := proxy.New(ctx)
 
 	containersv1.RegisterContainersServer(s, p)
+	volumesv1.RegisterVolumesServer(s, p)
```
chris-crone, line 68: nit: Any reason not to sort these?

protos/volumes/v1/volumes.pb.go
```diff
@@ -0,0 +1,855 @@
+//
+//  Copyright 2020 Docker, Inc.
```
chris-crone, line 2: This can't be Docker. Inc. with just a DCO:

```suggestion
//  Copyright 2020 Docker Compose CLI authors
```

protos/volumes/v1/volumes.proto
```diff
@@ -0,0 +1,63 @@
...
+	rpc VolumesDelete(VolumesDeleteRequest) returns (VolumesDeleteResponse);
+}
+
+message Volume {
+	string id = 1;
+	string description = 2;
+}
+
+message AciVolumeCreateOptions {
+	string storage_account= 1;
```
chris-crone, line 36:
```suggestion
	string storage_account = 1;
```

protos/volumes/v1/volumes.proto
```diff
@@ -0,0 +1,63 @@
...
+message VolumesListRequest {
+}
+
+message VolumesListResponse {
+	repeated Volume volumes = 1;
+}
+
+message VolumesDeleteRequest {
+	string id = 1;
+	repeated google.protobuf.Any options = 2;
```
chris-crone, line 59: Curious what options we can pass here?

server/proxy/volumes.go
```diff
@@ -0,0 +1,85 @@
...
+
+// VolumesCreate creates a volume.
+func (p *proxy) VolumesCreate(ctx context.Context, req *volumesv1.VolumesCreateRequest) (*volumesv1.VolumesCreateResponse, error) {
+	backendClient := Client(ctx)
+	contextType := ""
+	if backendClient != nil {
+		contextType = backendClient.ContextType()
+	}
+
+	if contextType != "aci" {
```
chris-crone, line 36: Do we have consts for the types that we can use?

server/proxy/volumes.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, fmt.Errorf("volume API only available in ACI context")
```
chris-crone, line 37: You can use not implemented from our errdefs:
```suggestion
		return nil, errors.Wrapf(errdefs.ErrNotImplemented, "volume API only available in ACI context")
```

chris-crone (commented):
Generally looks good, only minor issues

cli/cmd/serve.go
gtardif: no, fixed

protos/volumes/v1/volumes.proto
gtardif: This was to potentially pass more info for other backends. but we can add this later as an optional field, this wil not break the API. removing this field

server/proxy/volumes.go
gtardif: Yes, Thx !

server/proxy/volumes.go
gtardif: fixed

server/proxy/volumes.go
```diff
@@ -0,0 +1,91 @@
...
+
+// VolumesCreate creates a volume.
+func (p *proxy) VolumesCreate(ctx context.Context, req *volumesv1.VolumesCreateRequest) (*volumesv1.VolumesCreateResponse, error) {
+	backendClient := Client(ctx)
+	contextType := ""
+	if backendClient != nil {
+		contextType = backendClient.ContextType()
+	}
+
+	if contextType != store.AciContextType {
```
rumpl, line 39: Not sure we really want this here, the backend should return a `not implemented` error instead and then the cli can say that the volumes is not implemented. no ?

gtardif: indeed this is not reached if not ACI

protos/streams/v1/streams.pb.go
```diff
@@ -23,15 +23,16 @@ package v1
 
 import (
 	context "context"
+	reflect "reflect"
```
rumpl, line 26: Why did these change? How did you generate the protos?

server/proxy/volumes.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil, errors.Wrapf(errdefs.ErrNotImplemented, "volume API only available in ACI context")
+	}
+	aciReq := req.GetAciOption()
+	if aciReq == nil {
```
rumpl, line 43: This check should be in the backend too no?

gtardif: Actually we need to read the gRPC param and convert it into params for the backend. We can just pass an empty storage account by default and let the backend return the error, that would avoid different errors for different cases, yes.

protos/streams/v1/streams.pb.go
gtardif: This is me running `goimports -local github.com/docker/compose-cli -w .` after `make protos`. I'll remove this to not change all generated files in this PR, only volume changes required

server/metrics.go
```diff
@@ -36,9 +36,12 @@ var (
 		"/com.docker.api.protos.containers.v1.Containers/Kill":    "kill",
 		"/com.docker.api.protos.containers.v1.Containers/Inspect": "inspect",
 		"/com.docker.api.protos.containers.v1.Containers/Logs":    "logs",
-		"/com.docker.api.protos.streams.v1.Streaming/NewStream":   "",
+		"/com.docker.api.protos.streams.v1.Streaming/NewStream":   "streaming",
```
rumpl, line 39: Why this change?

gtardif: I just didn't want to let this empty, if anyone calls this, we'll see something we can follow in the metrics

rumpl (approved):
LGTM

Issue #607, opened 2020-09-11
Cli command json output


gtardif:
Not all CLI commands support json output at the moment. 

Also, we want to have a `--format` flag (rather than --json), supporting for the moment `--format json` and making it easy to allow more formats
Not proposing go template formatting in order to simplify usage (users should not have to type things like `--format {{ json . }}`), and if filtering is needed, jq is here to help scripting this.

- [x] `docker context ls`
- [x] `docker ps`
- [x] `docker compose ls`
- [x] `docker compose ps`
- [x] `docker secret ls`
- [x] `docker volume ls`
- [x] `docker version` (needs special attention due to retro-compatibility with the moby CLI and `--format` flag)

These commands should allow `--format json` to output json results

PR #689, opened 2020-09-28
Add json output format to several commands

ulyssessouza:
**What I did**
Add json output format to several commands

- [x] docker context ls - To make the `--json` hidden and implement `--format json`
- [x] docker ps
- [x] docker compose ls
- [x] docker compose ps
- [x] docker secret ls
- [x] docker volume ls
- [x] docker version - To clarify retro compatibility issues
- [x] Add tests

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/607

format


go.mod
```diff
@@ -62,5 +62,6 @@ require (
 	google.golang.org/protobuf v1.25.0
 	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
 	gopkg.in/ini.v1 v1.61.0
+	gotest.tools v2.2.0+incompatible
```
ulyssessouza, line 65: This is the result of a `go mod tidy`

cli/cmd/version.go
```diff
@@ -18,43 +18,94 @@ package cmd
...
 	// Still, technically the [] byte versionResult could be nil, just let the original command display what it has to display
 	if versionResult == nil {
 		mobycli.Exec(cmd.Root())
-		return nil
+		return ""
+	}
+	return string(versionResult)
+}
+
+func fixedPrettyArgs(oArgs []string) []string {
```
ulyssessouza, line 80: Not very happy with this, but works fine

cli/cmd/version.go
```diff
@@ -18,43 +18,94 @@ package cmd
...
+			(strings.ToLower(oArgs[i+1]) == formatter.PRETTY || oArgs[i+1] == "") {
+			i++
+			continue
+		}
+		args = append(args, oArgs[i])
+	}
+	return args
+}
+
+func fixedJsonArgs(oArgs []string) []string {
```
ulyssessouza, line 94: Not very happy with this, but works fine

cli/cmd/context/ls.go
```diff
@@ -93,35 +95,41 @@ func runList(cmd *cobra.Command, opts lsOpts) error {
...
+	return printContextLsFormatted(opts.format, currentContext, os.Stdout, contexts)
+}
 
-	for _, c := range contexts {
-		contextName := c.Name
-		if c.Name == currentContext {
-			contextName += " *"
+func printContextLsFormatted(format string, currContext string, out io.Writer, contexts []*store.DockerContext) error {
+	var err error
+	switch strings.ToLower(format) {
```
rumpl, line 106: What about
```golang
err = formatter.Print(context, format, func(w io.Writer) {
	for _, c := range contexts {
		contextName := c.Name
		if c.Name == currContext {
			contextName += " *"
		}
		fmt.Fprintf(w, "%s\t%s\t%s\t%s\t%s\t%s\n",
			contextName,
			c.Type(),
			c.Metadata.Description,
			getEndpoint("docker", c.Endpoints),
			getEndpoint("kubernetes", c.Endpoints),
			c.Metadata.StackOrchestrator)
	}
} , "NAME", "TYPE", "DESCRIPTION", "DOCKER ENDPOINT", "KUBERNETES ENDPOINT", "ORCHESTRATOR")
```

And we let the formatter switch the format and call the `ToStandardJSON` when needed?
This way this can be reused everywhere.

gtardif (commented):
We should return `[]` and not null when there are no elements : 
```
$ ./bin/docker --context aci  ps --format json
null
```

cli/cmd/compose/ps.go
```diff
@@ -59,17 +61,26 @@ func runPs(ctx context.Context, opts composeOptions) error {
...
+	var err error
+	switch strings.ToLower(format) {
+	case formatter.PRETTY, "":
+		err = formatter.PrintPrettySection(out, func(w io.Writer) {
+			for _, service := range serviceList {
+				fmt.Fprintf(w, "%s\t%s\t%d/%d\t%s\n", service.ID, service.Name, service.Replicas, service.Desired, strings.Join(service.Ports, ", "))
+			}
+		}, "ID", "NAME", "REPLICAS", "PORTS")
+	case formatter.JSON:
+		outJSON, err := formatter.ToStandardJSON(serviceList)
```
gtardif, line 77: Same question here, although it does not expose too many additional things. At least we should set json attribut names in the go struct, current field case is mainly driven by public/private fields and go linter, that should not drive the json output

cli/cmd/compose/testdata/compose-list-out-json.golden
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "ID": "123",
+        "Name": "myName123",
+        "Status": "Running"
+    }
+]
```
gtardif, line 7: nit

cli/cmd/context/ls.go
```diff
@@ -58,7 +59,8 @@ func listCommand() *cobra.Command {
 	}
 	cmd.Flags().BoolVarP(&opts.quiet, "quiet", "q", false, "Only show context names")
 	cmd.Flags().BoolVar(&opts.json, "json", false, "Format output as JSON")
-	cmd.Flags().StringVar(&opts.format, "format", "", "Format output as JSON")
+	cmd.Flags().StringVar(&opts.format, "format", "", "Format the output. Values: [pretty | json | go template]. (Default: pretty)")
```
gtardif, line 62: We support go templating here only for backward compatibility, not sure we should mention it in the doc

cli/cmd/context/ls.go
gtardif: +1. This goes with my concerns above regarding exposing our go structs and not returning the same information depending on format json or pretty. Using the formatter to output both json and pretty enforces better choices there

cli/cmd/ps.go
```diff
@@ -59,50 +54,69 @@ func PsCommand() *cobra.Command {
 	cmd.Flags().BoolVarP(&opts.quiet, "quiet", "q", false, "Only display IDs")
 	cmd.Flags().BoolVarP(&opts.all, "all", "a", false, "Show all containers (default shows just running)")
 	cmd.Flags().BoolVar(&opts.json, "json", false, "Format output as JSON")
+	cmd.Flags().StringVar(&opts.format, "format", "", "Format the output. Values: [pretty | json | go template]. (Default: pretty)")
```
gtardif, line 57: Same here, we support got template for backward compatibility reason here, not sure we want to document it for the long run

cli/cmd/ps.go
```diff
@@ -59,50 +54,69 @@ func PsCommand() *cobra.Command {
...
+	_ = cmd.Flags().MarkHidden("json")
```
gtardif, line 58: Maybe a comment mentionning this is used in VSCode Docker extension for the moment

cli/cmd/ps.go
```diff
@@ -59,50 +54,69 @@ func PsCommand() *cobra.Command {
...
+	switch strings.ToLower(format) {
+	case formatter2.PRETTY, "":
+		err = formatter2.PrintPrettySection(out, func(w io.Writer) {
+			for _, c := range containers {
+				fmt.Fprintf(w, "%s\t%s\t%s\t%s\t%s\n", c.ID, c.Image, c.Command, c.Status,
+					strings.Join(formatter.PortsToStrings(c.Ports, fqdn(c)), ", "))
+			}
+		}, "CONTAINER ID", "IMAGE", "COMMAND", "STATUS", "PORTS")
+	case formatter2.JSON:
+		out, err := formatter2.ToStandardJSON(containers)
```
gtardif, line 110: Same here

cli/cmd/version.go
```diff
@@ -18,43 +18,94 @@ package cmd
...
+	var versionString string
+	format := strings.TrimSpace(cmd.Flag(formatOpt).Value.String())
 	displayedVersion := strings.TrimPrefix(version, "v")
-	versionResult, _ := mobycli.ExecSilent(cmd.Context())
+	// Replace is preferred in this case to keep the order.
+	switch format {
+	case formatter.PRETTY, "":
+		versionString = strings.Replace(getOutFromMoby(cmd, fixedPrettyArgs(os.Args[1:])...),
+			"\n Version:", "\n Cloud integration:  "+displayedVersion+"\n Version:", 1)
+	case formatter.JSON, "{{json .}}", "{{json . }}", "{{ json .}}", "{{ json . }}": // Try to catch full JSON formats
```
gtardif, line 62: easier if you replace space first ? (`replace(" ", "")`)

gtardif: and toLowerCase

cli/mobycli/exec.go
```diff
@@ -112,7 +112,10 @@ func IsDefaultContextCommand(dockerCommand string) bool {
 }
 
 // ExecSilent executes a command and do redirect output to stdOut, return output
-func ExecSilent(ctx context.Context) ([]byte, error) {
-	cmd := exec.CommandContext(ctx, ComDockerCli, os.Args[1:]...)
+func ExecSilent(ctx context.Context, args ...string) ([]byte, error) {
+	if len(args) == 0 {
```
gtardif, line 116: How was this change required ? maybe we should change the caller to make sure args include the `docker` command

cli/cmd/compose/testdata/compose-list-out-json.golden
ulyssessouza: That's the good one. It's a golden file

cli/mobycli/exec.go
ulyssessouza: This is used on `getOutFromMoby` when shelling out to moby. The idea is to hack the `format` option before passing it to moby, since it should treat the cases of `pretty` and `json`. Both are inexistent on moby, so I do a little trick to translate it.

cli/cmd/context/ls.go
ulyssessouza: Done

cli/cmd/ps.go
ulyssessouza: Done

cli/cmd/version.go
ulyssessouza: Done

aci/containers.go
```diff
@@ -53,7 +53,7 @@ func (cs *aciContainerService) List(ctx context.Context, all bool) ([]containers
 	if err != nil {
 		return nil, err
 	}
-	var res []containers.Container
+	res := []containers.Container{}
```
rumpl, line 56: Unrelated change is unrelated :D

ulyssessouza: It is related.

The previous results in `nil` in the case of an empty list. The new results in an `empty slice`.

This is related to the `JSON` serialization. It handles differently `nil` and an `empty slice`.

ulyssessouza: Maybe `make([]containers.Cotainer, 0)` could make that more specific.

ulyssessouza:
@gtardif @rumpl PTAL

gtardif (changes requested):
Couple of things: 
* `docker ps --format json` does not display ports properly : `,"Ports":["20.56.234.156:80-\u003e80/tcp"]`
* More important (moby regression): version command is not forwarded properly to moby when we don't know the fomat: 
` ./bin/docker  version --format "{{ yaml . }}" does not display anything and exits with 0
The moby command  says `Template parsing error: template: version:1: function "yaml" not defined`
Same with ` $ ./bin/docker  version --format "{{ json .Client }}"`: empty result.
(`context ls` does not seem to have this issue).

ulyssessouza:
@gtardif PTAL

gtardif (approved):

BuonOmo:
Hey, is it normal that this commit is available in v1.0.X, v2.X, but not versions in between ? :) (source: the tags associated with the commit https://github.com/docker/compose-cli/commit/8961805412ba307255f378680a1af903b19b384c)

Issue #661, opened 2020-09-25
Let user select AMI / machine type


ndeloof:
**Description**
Allow end-users to select the AMI and machine type to be used for EC2 provisioning.
We can rely on `deploy.placement` for this purpose

PR #696, opened 2020-09-29
Allow user to set ami/machine by Deploy.Placement.Constraint

ndeloof:
**What I did**
User can select custom AMI and machine type to configure CapacityProvider by setting `node.ami` and `node.machine` placement constraints

**Related issue**
https://github.com/docker/compose-cli/issues/661

image

### DIFF
--- a/ecs/ec2.go
+++ b/ecs/ec2.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"strings"
 
 	"github.com/awslabs/goformation/v4/cloudformation"
 	"github.com/awslabs/goformation/v4/cloudformation/autoscaling"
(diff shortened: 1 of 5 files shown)


ecs/ec2.go
```diff
@@ -41,14 +53,20 @@ func (b *ecsAPIService) createCapacityProvider(ctx context.Context, project *typ
 		return nil
 	}
 
-	ami, err := b.SDK.GetParameter(ctx, "/aws/service/ecs/optimized-ami/amazon-linux-2/gpu/recommended")
-	if err != nil {
-		return err
+	if ami != "" {
```
aiordache, line 56: At first glance it looks like we discard the user defined `ami` value and get the recommended one, or am I misunderstanding?

ndeloof: indeed, should be `==`. Good catch

chris-crone (approved):
LGTM

Issue #658, opened 2020-06-29
Please support Sovereign clouds like AzureUSGovernment 


actions-devops:
Hello, Please add an option to specify the sovereign cloud option similiar to the AZ CLI command 
```
az cloud set --name AzureUSGovernment
```

Please also use the following endpoints for login (AzureUSGovernment as an example)
```json
{
    "endpoints": {
      "activeDirectory": "https://login.microsoftonline.us",
      "activeDirectoryDataLakeResourceId": null,
      "activeDirectoryGraphResourceId": "https://graph.windows.net/",
      "activeDirectoryResourceId": "https://management.core.usgovcloudapi.net/",
      "appInsightsResourceId": "https://api.applicationinsights.us",
      "appInsightsTelemetryChannelResourceId": "https://dc.applicationinsights.us/v2/track",
      "batchResourceId": "https://batch.core.usgovcloudapi.net/",
      "gallery": "https://gallery.usgovcloudapi.net/",
      "logAnalyticsResourceId": "https://api.loganalytics.us",
      "management": "https://management.core.usgovcloudapi.net/",
      "mediaResourceId": "https://rest.media.usgovcloudapi.net",
      "microsoftGraphResourceId": "https://graph.microsoft.us/",
      "ossrdbmsResourceId": "https://ossrdbms-aad.database.usgovcloudapi.net",
      "resourceManager": "https://management.usgovcloudapi.net/",
      "sqlManagement": "https://management.core.usgovcloudapi.net:8443/",
      "vmImageAliasDoc": "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/arm-compute/quickstart-templates/aliases.json"
    },
    "isActive": true,
    "name": "AzureUSGovernment",
    "profile": "latest",
    "suffixes": {
      "acrLoginServerEndpoint": ".azurecr.us",
      "azureDatalakeAnalyticsCatalogAndJobEndpoint": null,
      "azureDatalakeStoreFileSystemEndpoint": null,
      "keyvaultDns": ".vault.usgovcloudapi.net",
      "mariadbServerEndpoint": ".mariadb.database.usgovcloudapi.net",
      "mysqlServerEndpoint": ".mysql.database.usgovcloudapi.net",
      "postgresqlServerEndpoint": ".postgres.database.usgovcloudapi.net",
      "sqlServerHostname": ".database.usgovcloudapi.net",
      "storageEndpoint": "core.usgovcloudapi.net"
    }
  }
```

[...]

PR #700, opened 2020-09-30
Regroup azure constants that might be things to switch in order to provide access to other azure clouds

gtardif:
See `az cloud list` to list endpoints specific to each azure cloud.
See #658 

**What I did**
* regroup & de-duplicate azure constants

**Related issue**

### DIFF
--- a/aci/login/login.go
+++ b/aci/login/login.go
@@ -38,12 +38,18 @@ import (
 
 //go login process, derived from code sample provided by MS at https://github.com/devigned/go-az-cli-stuff
 const (
-	authorizeFormat = "https://login.microsoftonline.com/organizations/oauth2/v2.0/authorize?response_type=code&client_id=%s&redirect_uri=%s&state=%s&prompt=select_account&response_mode=query&scope=%s"
-	tokenEndpoint   = "https://login.microsoftonline.com/%s/oauth2/v2.0/token"
-	getTenantURL    = "https://management.azure.com/tenants?api-version=2019-11-01"
+	// AcrRegistrySuffix suffix for ACR registry images
+	AcrRegistrySuffix         = ".azurecr.io"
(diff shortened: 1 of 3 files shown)


aci/login/login.go
```diff
@@ -38,12 +38,18 @@ import (
...
 //go login process, derived from code sample provided by MS at https://github.com/devigned/go-az-cli-stuff
 const (
-	authorizeFormat = "https://login.microsoftonline.com/organizations/oauth2/v2.0/authorize?response_type=code&client_id=%s&redirect_uri=%s&state=%s&prompt=select_account&response_mode=query&scope=%s"
-	tokenEndpoint   = "https://login.microsoftonline.com/%s/oauth2/v2.0/token"
-	getTenantURL    = "https://management.azure.com/tenants?api-version=2019-11-01"
+	// AcrRegistrySuffix suffix for ACR registry images
+	AcrRegistrySuffix         = ".azurecr.io"
+	activeDirectoryURL        = "https://login.microsoftonline.com"
+	azureManagementURL        = "https://management.core.windows.net/"
+	azureResouceManagementURL = "https://management.azure.com/"
```
ulyssessouza, line 45:
```suggestion
	azureManagementURL        = "https://management.core.windows.net"
	azureResouceManagementURL = "https://management.azure.com"
```
What about normalizing the end of the constant so it's responsability of the user to specify the "/"?
This makes explicit from the reader that the previous section of the URL is completed and whatever is concatenated is another section.

gtardif: Actually I tried to stick as much as possible with values we get if you try to do `az cloud list` so that if we want to switch clouds we can easily replace these const.

ulyssessouza: OK

ulyssessouza (approved):
LGTM

Issue #656, opened 2020-07-17
Allow passing arguments to the container 


metcalfc:
The way the args are parsed there is no way to pass arguments to the container itself. Local docker I can:

```
docker run  -it --rm --name=iperf3-server -p 5201:5201 networkstatic/iperf3 -s
```

And the `-s` is passed to iperf3. In ACI's run:

```
docker run  -p 5201:5201 networkstatic/iperf3 -s
unknown shorthand flag: 's' in -s
```

gtardif:
Image command can now be overriden in the command line. 
In ACI, overriding command parameters also overrides the entrypoint of the image if one is defined, unlike with the docker local engine. ACI does not have the notion of entrypoint that we could potentially override.
So you need to slightly change your command line for ACI invocation, and add the actual binary invoked, that is defined in the image entrypoint. 
(for images defining a COMMAND but no ENTRYPOINT, there is no difference between ACI and the local docker engine)

So on ACI I managed to invoke: 
```
> docker --context acicontext run --name=iperf3-server -p 5201:5201 networkstatic/iperf3 iperf3 -s -V -d
[+] Running 2/2
 ⠿ Group iperf3-server  Created                  9.0s
 ⠿ iperf3-server        Done                           51.9s
2020-10-07T10:34:26.9208801Z stdout F iperf 3.6
2020-10-07T10:34:26.9208801Z stdout F Linux SandboxHost-637376636434999278 4.19.104-microsoft-standard #1 SMP Fri Feb 21 04:01:19 UTC 2020 x86_64
2020-10-07T10:34:26.9449005Z stderr F warning: this system does not seem to support IPv6 - trying IPv4
```

PR #702, opened 2020-09-30
Add argument support to run command

gtardif:
Signed-off-by: Ulysses Souza <ulyssessouza@gmail.com>

**What I did**
Rebase commits from #542 that is now obsolete (fork was disconnected from compose-cli when the upstream repo was open sourced)

**Related issue**
https://github.com/docker/compose-cli/issues/656


karolz-ms:
What about [entrypoint](https://docs.docker.com/compose/compose-file/#entrypoint)? Should that also be mapped?

One might read in ACI docs that entrypoint can be overridden and try to do the same in Compose file...

gtardif:
> What about [entrypoint](https://docs.docker.com/compose/compose-file/#entrypoint)? Should that also be mapped?
> 
> One might read in ACI docs that entrypoint can be overridden and try to do the same in Compose file...

Indeed, That should ideally be mapped, unfortunately in ACI payload I see only Image and Command fields, but no way to override Entrypoint specifically.
You're right @karolz-ms, we need to document this point specifically, extending the doc started at https://github.com/docker/compose-cli/blob/main/docs/aci-interation.md.

ulyssessouza (approved):
LGTM

PR #708, opened 2020-10-01
Fix ACI e2e : now we support passing commands, set `--domainname` before the image in the cmd

gtardif:

**What I did**
* changed order of parameters in e2e invocation

**Related issue**
https://github.com/docker/compose-cli/runs/1192523711, container logs show explicit error while ACI is trying to restart container in a loop : `starting container process caused "exec: \"--domainname\": executable file not found in $PATH"`

/test-aci


ulyssessouza (approved):
LGTM

Issue #360, opened 2020-07-07
Add run option in grpc API to allow setting environment variables


gtardif:
Follow up on https://github.com/docker/api/pull/358

PR #709, opened 2020-10-01
Add environment variables to grpc run request.

aiordache:
Add EnvVars property to the RunRequest and update the unit tests to check the container.Environment is set.

Closes https://github.com/docker/compose-cli/issues/360

Test with the node-sdk query:
```
import * as grpc from '@grpc/grpc-js';
import { RunRequest } from '../src/containers';
import { Containers } from '../src/index';

const client = new Containers("unix:///tmp/backend.sock");
const request = new RunRequest()
    .setImage('nginx')
    .setId('grpctest')
    .setEnvironmentList(["BLA=BLABLA"])
    .setCpuLimit(1.0)

client.run(
    request,
  (error: grpc.ServiceError | null) => {
    if (error != null) {
      throw error;
    }
  }
);
```

```
$ docker exec -it grpctest sh
# env
...
BLA=BLABLA

```


gtardif (approved):
LGTM, you can rename to field to `environment` indeed

PR #711, opened 2020-10-01
ACI tests : Wait for http status and retry if necessary. 

gtardif:
We were getting 500 errors because database is not ready initially

**What I did**
* do not only retry GET on errors, also retry if status is not the expected one.

**Related issue**
https://github.com/docker/compose-cli/runs/1193172850

/test-aci


ulyssessouza (approved):
LGTM

Issue #686, opened 2020-09-28
Add e2e tests for metrics sending


gtardif:
Code sending metrics is adding more and more detailed features : separating source (CLI/API), separating success/failures/cancel, with different codepaths for executing in default contexts or cloud backends. 
Some things are tested with unit tests, but we need to add e2e tests making sure the right json payload is sent to Docker Desktop for various cases. 
Given we already have a large set of e2e tests, we can add a mock local server listening for metrics, and ensure in some e2e tests that we send metrics properly.

PR #713, opened 2020-10-02
E2e test on metrics

gtardif:

**What I did**
* Mock metrics with a custom socket for tests
* build flag to read test metrics socket from env
* add local e2e tests checking metric on moby context and example context

**Related issue**
https://github.com/docker/compose-cli/issues/686

/test-windows


gtardif:
I will need to adapt the metrics test socket on windows with pipes

gtardif:
> I will need to adapt the metrics test socket on windows with pipes

Actually we can also use unix sockets on windows to send metrics. Using this only for the server mock, not changing the Desktop current setup on windows.

.github/workflows/ci.yml
```diff
@@ -55,12 +55,12 @@ jobs:
...
       - name: Test
         env:
-          BUILD_TAGS: example,local,ecs
+          BUILD_TAGS: example,local
         run: make -f builder.Makefile test
 
       - name: Build
         env:
-          BUILD_TAGS: example,local,ecs
+          BUILD_TAGS: example,local,e2e
```
chris-crone, line 63: Not for this PR but I think we need to rename the steps in the CI workflow. It's a bit weird to do end to end testing in the "Build" step.

I noticed this because we're setting e2e as a build tag here which sounded wrong for build.

ndeloof (approved):

.github/workflows/ci.yml
gtardif: updated name

PR #715, opened 2020-10-02
Fix regression: let users login to local registry (no ‘.’ In registry name)

gtardif:

**What I did**
* Removed login arg check since `azure` is now a subcommand and not a code path in the `login` command. 

**Related issue**
https://github.com/docker/compose-cli/issues/712

/test-aci
/test-windows

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -205,9 +205,30 @@ func TestLoginCommandDelegation(t *testing.T) {
 		})
 	})
 
+	t.Run("localhost registry interactive", func(t *testing.T) {
+		res := c.RunDockerOrExitError("login", "localhost:443")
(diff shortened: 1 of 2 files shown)


tests/e2e/e2e_test.go
```diff
@@ -205,9 +205,22 @@ func TestLoginCommandDelegation(t *testing.T) {
 		})
 	})
 
+	t.Run("localhost registry", func(t *testing.T) {
```
chris-crone, line 208: Would be good to add a test for `localhost` without the port as well:

```console
com.docker.cli login localhost
Username: bob
Password: 
Error response from daemon: Get http://localhost/v2/: dial tcp [::1]:80: connect: connection refused
``

chris-crone (approved):
One more test but then LGTM, thanks!

PR #718, opened 2020-10-02
delete CapacityProvider before we invoke DeleteStack

ndeloof:
**What I did**
this is a workaround for CloudFormation issue to manage CapacityProvider <-> Cluster reverse dependency
`compose down` on ECS first delete the CapacityProvider and AutoscalingGroup (if present) by imperative API call then run `DeleteStack`. This prevent Cluster deletion by CloudFormation to fail

**Related issue**
https://github.com/docker/docker_aws/issues/43

/test-ecs

image

### DIFF
--- a/ecs/sdk.go
+++ b/ecs/sdk.go
@@ -32,6 +32,8 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/autoscaling"
+	"github.com/aws/aws-sdk-go/service/autoscaling/autoscalingiface"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/aws-sdk-go/service/cloudformation/cloudformationiface"
 	"github.com/aws/aws-sdk-go/service/cloudwatchlogs"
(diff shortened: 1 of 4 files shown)


ecs/sdk.go
```diff
@@ -364,7 +368,23 @@ type stackResource struct {
...
-func (s sdk) ListStackResources(ctx context.Context, name string) ([]stackResource, error) {
+type stackResourceFn func(r stackResource) error
+
+type stackResources []stackResource
+
+func (resources stackResources) apply(awsType string, fn stackResourceFn) error {
+	for _, r := range resources {
+		if r.Type == awsType {
+			err := fn(r)
+			if err != nil {
```
gtardif, line 379: we can use multierr here, see for example https://github.com/gtardif/compose-cli/blob/main/cli/cmd/rm.go#L60

ecs/wait.go
```diff
@@ -28,8 +28,12 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 )
 
-func (b *ecsAPIService) WaitStackCompletion(ctx context.Context, name string, operation int) error { //nolint:gocyclo
+func (b *ecsAPIService) WaitStackCompletion(ctx context.Context, name string, operation int, ignored ...string) error { //nolint:gocyclo
 	knownEvents := map[string]struct{}{}
+	for _, id := range ignored {
```
gtardif, line 33: Looking at the code, was this causing issues ? at first sight it seems it would refresh the progress display for these "ignored" events but should redisplay the same things. I guess this was not the case, to require this additional `ignored` params

ecs/sdk.go
ndeloof: Initial usage expect to only match a single resource, but makes sense

ecs/wait.go
ndeloof: issue is we show all resources as "CREATE_COMPLETE" and this override the "DELETE_" status that we just created for capacity provider. Always has been weird to me we first display this "CREATE" or "UPDATE" status for previous stack operation.

gtardif (approved):

Issue #685, opened 2020-09-28
Normalize ECS volume names


chris-crone:
**Description**
Tracking issue raised by @christophluehr [here](https://github.com/docker/compose-cli/pull/622#issuecomment-699615335)

Comment from Nicolas internally:
> We need to invoke `normalizeResourceName` on such name to make it CloudFormation compliant

PR #719, opened 2020-10-02
Normalize volume name

aiordache:
PR https://github.com/docker/compose-cli/pull/717 didn't fix the issue.

This should fix it (I tested it)

Closes https://github.com/docker/compose-cli/issues/685

### DIFF
--- a/ecs/volumes.go
+++ b/ecs/volumes.go
@@ -42,7 +42,7 @@ func (b *ecsAPIService) createNFSmountIngress(securityGroups []string, project *
 				}
 				break
 			}
-			name := fmt.Sprintf("%sNFSMount%s", normalizeResourceName(s.Name), n)
+			name := fmt.Sprintf("%sNFSMount%s", normalizeResourceName(s.Name), normalizeResourceName(n))
 			template.Resources[name] = &ec2.SecurityGroupIngress{
 				Description:           fmt.Sprintf("Allow NFS mount for %s on %s", s.Name, n),
 				GroupId:               target,


gtardif (commented):
Probably worth a unit test. 
BTW (not related to this PR), it seems we can remove the error return type, [we only return nil](https://github.com/docker/compose-cli/pull/719/files#diff-db0d254ee03206549e7d89bd881def1eL58)

ecs/volumes.go
```diff
@@ -42,7 +42,7 @@ func (b *ecsAPIService) createNFSmountIngress(securityGroups []string, project *
 				}
 				break
 			}
-			name := fmt.Sprintf("%sNFSMount%s", normalizeResourceName(s.Name), n)
+			name := fmt.Sprintf("%sNFSMount%s", normalizeResourceName(s.Name), normalizeResourceName(n))
```
thaJeztah, line 45: (not for this PR, but it looks like using ` normalizeResourceName()` can lead to ambiguity; https://github.com/docker/compose-cli/blob/ced0eb3912c8ff18d5245775be04098127e93ec5/ecs/cloudformation.go#L442-L444

For example, . `hello world`, `h e l l o world`, `h$%e^&lloworld`, `hello-world`, `Hello___world`, and `h
	e
	l
	l
	o
	
	
    world!!!!`

Are all considered equal. (see https://play.golang.org/p/Lp3xWkxztpy)

Is there a validation step before we reach this code (and other places where we normalise?); I'm always a bit scary with silently "fixing up" user-mistakes (if possible prefer to produce a clear, but "fatal" error to not hide mistakes).

ndeloof (approved):

thaJeztah:
I'd have appreciated if comments were not ignored; perhaps my comment is of no concern whatsoever, but in that case, just a small comment to confirm that would be good.

thaJeztah:
Opened https://github.com/docker/compose-cli/issues/722 for tracking

Issue #682, opened 2020-09-28
CI red because of ACI size limitation per region


gtardif:
ACI e2e test fail from time to time with `StandardCores' exceeded in region 'eastus2'. Limit: '10'`. 
This happens when github actions run too many builds at the same time (and tests are parallelized so the 10 limit is reached  quite easily). 
All tests are executed in region eastus2 for now, we should set a list of regions and randomly select one when starting tests, to avoid having too many tests in // in the same region.

gtardif:
We should exclude some regions where we already have some user resources in our MSFT subscription, to avoid sharing resources between manual usage and CI: 
* West europe
* France central
* East US
* UK South (@djs55 , maybe if you're not making a huge use of Azure, you can move to West europe to rationalize our usage ?)

PR #720, opened 2020-10-02
Running tests in many regions to avoid per region limit when we run too many // things in the CI

gtardif:

**What I did**
Run tests in random regions, listed from error message when trying non available regions

**Related issue**
https://github.com/docker/compose-cli/issues/682

/test-aci


ndeloof (approved):
¯\_(ツ)_/¯

PR #724, opened 2020-10-05
Allow `docker context ls —formatter {{json.}}` to not delegate to Moby and include context type in the json

gtardif:

**What I did**
* `docker context ls` reads format `{{json.}}`as `json` 

**Related issue**

### DIFF
--- a/cli/cmd/context/ls.go
+++ b/cli/cmd/context/ls.go
@@ -69,7 +69,8 @@ func runList(cmd *cobra.Command, opts lsOpts) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


cli/cmd/context/ls.go
```diff
@@ -93,7 +94,7 @@ func runList(cmd *cobra.Command, opts lsOpts) error {
 		return nil
 	}
 
-	if opts.json {
+	if format == formatter.JSON || format == formatter.TemplateJSON {
```
ulyssessouza, line 97: Looks like it doesn't work when passing `--json`. Could you please add a test for that?

gtardif: done

ulyssessouza (approved):
LGTM

PR #729, opened 2020-10-06
Add windows golden file for `context ls`

ulyssessouza:
**What I did**
Added the windows correspondent golden file for `context ls`

You can add a / mention to run tests executed by default only on main branch : 
/test-aci
/test-windows


gtardif (approved):

gtardif:
Some other red things dur to ACI issues today, but this fixes what it says it fixes. https://github.com/docker/compose-cli/pull/729/checks?check_run_id=1214094794#step:7:233

Issue #619, opened 2020-09-15
Include version info in user-agent when calling Azure / aws SDK


gtardif:
User agent is set to a hardcoded string ATM, we should include cli version info.

PR #734, opened 2020-10-07
Add version to UserAgent on  ACI

ulyssessouza:
**What I did**
Add version to UserAgent on  ACI
This also refactors the Version insertion to make it global

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/619

version

### DIFF
--- a/ecs/sdk.go
+++ b/ecs/sdk.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/docker/compose-cli/api/compose"
 	"github.com/docker/compose-cli/api/secrets"
+	"github.com/docker/compose-cli/internal"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/request"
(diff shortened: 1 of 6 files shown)


ecs/sdk.go
```diff
@@ -69,7 +72,7 @@ type sdk struct {
 
 func newSDK(sess *session.Session) sdk {
 	sess.Handlers.Build.PushBack(func(r *request.Request) {
-		request.AddToUserAgent(r, "Docker CLI")
+		request.AddToUserAgent(r, userAgent+"/"+internal.Version)
```
ulyssessouza, line 75: Does that break some metrics stuff ??? I've set to "docker-cli" just to be the same as it's ACI counterpart.

gtardif: On ECS, I don't know if it breaks things (@ndeloof ? ) 
Why change "Docker CLI" to "docker-cli" ? (I don't mind really, but trying to minimize changes, probably doesn't change anything to breaking some metrics)

gtardif (approved):
LGTM

ecs/sdk.go
gtardif: On ACI, we need to check with @macolso how this impacts Docker-ACI metrics, probably needs some adaptation.

ndeloof: We told AWS we use `Docker CLI` in user-agent string.
Introducing version is fine, could help check users updated to latest release, but this base string should not chnage

ulyssessouza: @ndeloof So maybe doing the other way around could be a good thing...
If we change to `Docker CLI` on both, maybe that would work. @gtardif Could you please check with @macolso if that's OK to change that to `Docker CLI` ?

ndeloof (approved):

Issue #678, opened 2020-07-18
add --password-stdin flag to secret create


fracturx:
This would make working with the plugin so much more easier, and secure. I don't want to type in my token each time i'd like to add a new secret when i could do this in docker login:
```powershell
aws ecr get-login-password | docker login -u MyUser --password-stdin
```
I really hope this is possible

PR #735, opened 2020-10-08
Add password file argument to secret create (handles '-' as stdin read)

aiordache:
Add password file argument to secret create
```
$ echo "secretpassword" | docker secret create test -
arn:aws:secretsmanager:eu-west-3:175142243308:secret:test-vDMzik
```
```
$ docker secret create test pass.txt
arn:aws:secretsmanager:eu-west-3:175142243308:secret:test-Ue0s6C

```
```
$ docker secret create test -
?  *************************************
arn:aws:secretsmanager:eu-west-3:175142243308:secret:test-Hru0Qu

```
Closes https://github.com/docker/compose-cli/issues/678

/test-ecs

### DIFF
--- a/api/secrets/api.go
+++ b/api/secrets/api.go
@@ -31,21 +31,17 @@ type Service interface {
 
 // Secret hold sensitive data
 type Secret struct {
-	ID          string            `json:"ID"`
-	Name        string            `json:"Name"`
-	Labels      map[string]string `json:"Labels"`
-	Description string            `json:"Description"`
-	username    string
-	password    string
+	ID      string            `json:"ID"`
+	Name    string            `json:"Name"`
+	Labels  map[string]string `json:"Tags"`
+	content []byte
 }
 
 // NewSecret builds a secret
-func NewSecret(name, username, password, description string) Secret {
+func NewSecret(name string, content []byte) Secret {
 	return Secret{
-		Name:        name,
-		username:    username,
-		password:    password,
-		Description: description,
+		Name:    name,
+		content: content,
 	}
 }
 
@@ -58,15 +54,7 @@ func (s Secret) ToJSON() (string, error) {
 	return string(b), nil
 }
 
-// GetCredString marshall a Secret's sensitive data into JSON string
-func (s Secret) GetCredString() (string, error) {
-	creds := map[string]string{
-		"username": s.username,
-		"password": s.password,
-	}
-	b, err := json.Marshal(&creds)
-	if err != nil {
-		return "", err
-	}
-	return string(b), nil
+// GetContent returns a Secret's sensitive data
+func (s Secret) GetContent() []byte {
+	return s.content
 }
--- a/cli/cmd/secrets.go
+++ b/cli/cmd/secrets.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 
 	"github.com/spf13/cobra"
@@ -27,13 +28,6 @@ import (
 	"github.com/docker/compose-cli/api/secrets"
 	"github.com/docker/compose-cli/formatter"
 )
-
-type createSecretOptions struct {
-	Label       string
-	Username    string
-	Password    string
-	Description string
-}
 
 // SecretCommand manage secrets
 func SecretCommand() *cobra.Command {
@@ -52,18 +46,39 @@ func SecretCommand() *cobra.Command {
 }
 
 func createSecret() *cobra.Command {
-	opts := createSecretOptions{}
 	cmd := &cobra.Command{
-		Use:   "create NAME",
+		Use:   "create [OPTIONS] SECRET [file|-]",
 		Short: "Creates a secret.",
-		Args:  cobra.ExactArgs(1),
+		Args:  cobra.RangeArgs(1, 2),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			c, err := client.New(cmd.Context())
 			if err != nil {
 				return err
 			}
+			file := "-"
+			if len(args) == 2 {
+				file = args[1]
+			}
+			if len(file) == 0 {
+				return fmt.Errorf("secret data source empty: %q", file)
+			}
+			var in io.ReadCloser
+			switch file {
+			case "-":
+				in = os.Stdin
+			default:
+				in, err = os.Open(file)
+				if err != nil {
+					return err
+				}
+				defer func() { _ = in.Close() }()
+			}
+			content, err := ioutil.ReadAll(in)
+			if err != nil {
+				return fmt.Errorf("failed to read content from %q: %v", file, err)
+			}
 			name := args[0]
-			secret := secrets.NewSecret(name, opts.Username, opts.Password, opts.Description)
+			secret := secrets.NewSecret(name, content)
 			id, err := c.SecretsService().CreateSecret(cmd.Context(), secret)
 			if err != nil {
 				return err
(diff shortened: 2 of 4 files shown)


ndeloof:
I wonder we should align with "classic" docker CLI `docker secret create [OPTIONS] SECRET [file|-]` UX and just drop support for passing plain text secret to this command

thaJeztah:
Yes, we should definitely deprecate/remove `--password`, passing the password as argument will lead it to be ending up in bash history, and other places.

Secrets (in the docker cli) are also not just passwords, so we should avoid the use of "password" as term

nebuk89:
> I wonder we should align with "classic" docker CLI `docker secret create [OPTIONS] SECRET [file|-]` UX and just drop support for passing plain text secret to this command

I agree, Sebastians point is a good one and this just makes everyones life messy having people think that --password is a good thing to use

api/secrets/api.go
```diff
@@ -35,16 +35,14 @@ type Secret struct {
...
 	Labels      map[string]string `json:"Labels"`
 	Description string            `json:"Description"`
-	username    string
-	password    string
+	content     string
 }
 
 // NewSecret builds a secret
-func NewSecret(name, username, password, description string) Secret {
+func NewSecret(name, password, description string) Secret {
```
thaJeztah, line 42:
```suggestion
func NewSecret(name, content, description string) Secret {
```

thaJeztah:
Also wondering where the `description` is coming from? For the existing CLI we don't have that, but we do provide labels as metadata (for users to use for their own use-case / conventions)

For the existing CLI we have

- `labels` as metadata (name/value)
- `name` (`[a-zA-Z0-9-_.]`, max 64 characters
- `id` (automatically generated, _no relation_ with the secret's content, so it's not a hash (e.g.))

Note that we also have a constraint on size (maximum: 500 kb)

Limiting the size that we read may be good to do, to prevent someone doing, e.g.

```bash
cat lord-of-the-rings-4k-HDR.mkv | docker secret create mymovie -
```


There's also an option for templating (allowing secrets and configs to substitute values at runtime), but that's not likely important for now

thaJeztah:
FWIW, I think the 64 (or 63?) character limit was to allow defining DNS schemes for resources, for which 63 characters is the maximum "label" length; https://www.ietf.org/rfc/rfc1035.txt

cli/cmd/secrets.go
```diff
@@ -56,13 +56,16 @@ func createSecret() *cobra.Command {
...
 			if err != nil {
 				return err
 			}
-			file := args[1]
+			file := "-"
+			if len(args) == 2 {
+				file = args[1]
+			}
 			if len(file) == 0 {
 				return fmt.Errorf("Secret data source empty: %q", file)
```
chris-crone, line 70: [Errors should start with lower case](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

```suggestion
				return fmt.Errorf("secret data source empty: %q", file)
```

cli/cmd/secrets.go
```diff
@@ -76,13 +79,11 @@ func createSecret() *cobra.Command {
 					return err
 				}
 				defer in.Close()
-
 			}
 			content, err := ioutil.ReadAll(in)
 			if err != nil {
 				return fmt.Errorf("Error reading content from %q: %v", file, err)
```
chris-crone, line 85: Same here
```suggestion
				return fmt.Errorf("error reading content from %q: %v", file, err)
```

cli/cmd/secrets.go
```diff
@@ -19,19 +19,19 @@ package cmd
...
 	"github.com/docker/compose-cli/api/client"
 	"github.com/docker/compose-cli/api/secrets"
 	"github.com/docker/compose-cli/formatter"
+	"github.com/docker/docker/pkg/system"
 )
 
 type createSecretOptions struct {
 	Label       string
-	Username    string
-	Password    string
```
chris-crone, line 34: This matches the existing CLI's secrets implementation which is flat (i.e.: secret name maps to a value) but I think it will break pulling private images where we need a dictionary with username and password.

ecs/sdk.go
```diff
@@ -524,14 +524,10 @@ func (s sdk) DeleteStack(ctx context.Context, name string) error {
 
 func (s sdk) CreateSecret(ctx context.Context, secret secrets.Secret) (string, error) {
 	logrus.Debug("Create secret " + secret.Name)
-	secretStr, err := secret.GetCredString()
-	if err != nil {
-		return "", err
-	}
-
 	response, err := s.SM.CreateSecret(&secretsmanager.CreateSecretInput{
+
```
thaJeztah, line 528: nit: stray empty line

cli/cmd/secrets.go
```diff
@@ -73,8 +95,6 @@ func createSecret() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVarP(&opts.Username, "username", "u", "", "username")
-	cmd.Flags().StringVarP(&opts.Password, "password", "p", "", "password")
 	cmd.Flags().StringVarP(&opts.Description, "description", "d", "", "Secret description")
```
thaJeztah, line 98: Still not sure about having a `--description` flag, unless there's an important use-case that we're addressing

cli/cmd/secrets.go
thaJeztah: Not a blocker, just a suggestion: I usually try to avoid "error" as part of the error message (it should be clear there's an error). Also could consider using `%w` to wrap the error, in case we want to do further processing (although not likely), for example;

```suggestion
				return fmt.Errorf("failed to read content from %q: %w", file, err)
```

thaJeztah (commented):
Overall looks good (but commits should be squashed). Left some minor suggestions, and a question around the `--description` flag

ndeloof (approved):

api/secrets/api.go
```diff
@@ -31,21 +31,17 @@ type Service interface {
...
 type Secret struct {
-	ID          string            `json:"ID"`
-	Name        string            `json:"Name"`
-	Labels      map[string]string `json:"Labels"`
-	Description string            `json:"Description"`
-	username    string
-	password    string
+	ID      string            `json:"ID"`
+	Name    string            `json:"Name"`
+	Tags    map[string]string `json:"Tags"`
```
thaJeztah, line 36: Why was this changed from Labels to Tags? Are these Azure or ECS-specific?

If these are the equivalent to labels, we could use `--label`, and translate them to `--tag` for the API call.
Alternatively (if they're separate from "user" labels), we could consider having a prefix (`--label com.azure.tag=foo`) 

(mostly curious) 😅

aiordache: It's ECS specific :sweat_smile: We tag most of AWS resources we create with compose and I thought it is easier to follow if I renamed Labels to Tags.

ndeloof: If the plan here is to use AWS tags to implement docker labels, then should better avoid to expose AWS terminology. 

If we go this way, I wonder how TaksDefinition's labels should be set, as it has both `Tags` and  `DockerLabels`. Maybe configured labels should be applied twice there.

aiordache: Ok, I will change back to Labels :sweat:

thaJeztah: Thanks! (and sorry for being a PITA!). We should definitely look if there's feature-disparity between the backends, and (where possible) re-use existing terminology to keep the UX consistent. At the same time, if we find things that will affect the usability too much, we should have a look if we can find some common (platform/backend-agnostic) way to address these

chris-crone (approved):
LGTM

ndeloof (approved):

Issue #733, opened 2020-10-07
ACI: Do not allow volume deletion if the volume is detected in use by runnning containers


gtardif:
`docker volume rm` currently deletes an azure file share, even if some containers are currently using it. This will cause runtime failures in the container when trying to reach files in the volume. 
When removing volumes, we should list containers (even outside the current resource group) and check if they are using the specified file share ; if so, return an error with the container id and resource group so the user know what to do in order to properly delete the container

PR #736, opened 2020-10-08
Look for all mounts before removing a volume

ulyssessouza:
**What I did**
This checks the volume mounts of all container groups from all
resource groups before removing a volume/fileshare

**Related issue**
https://github.com/docker/compose-cli/issues/733

**Tests**
/test-aci

### DIFF
--- a/aci/volumes.go
+++ b/aci/volumes.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/pkg/errors"
 
+	"github.com/Azure/azure-sdk-for-go/services/containerinstance/mgmt/2018-10-01/containerinstance"
 	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-06-01/storage"
 	"github.com/Azure/go-autorest/autorest/to"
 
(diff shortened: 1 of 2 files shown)


aci/volumes.go
```diff
@@ -164,7 +165,47 @@ func errorEvent(resource string) progress.Event {
...
+		if v.AzureFile != nil && v.AzureFile.StorageAccountName != nil && v.AzureFile.ShareName != nil &&
+			(*v.AzureFile.StorageAccountName+"/"+*v.AzureFile.ShareName) == id {
+			return true
+		}
+	}
+	return false
+}
+
 func (cs *aciVolumeService) Delete(ctx context.Context, id string, options interface{}) error {
+	err := checkVolumeUsage(ctx, cs.aciContext, id)
```
ulyssessouza, line 205: Maybe we should add a `--force` option to avoid the check...

aci/volumes.go
```diff
@@ -164,7 +165,47 @@ func errorEvent(resource string) progress.Event {
 	}
 }
 
+func checkVolumeUsage(ctx context.Context, aciContext store.AciContext, id string) error {
+	resourceGroups, err := NewACIResourceGroupHelper().ListGroups(ctx, aciContext.SubscriptionID)
+	if err != nil {
+		return err
+	}
+	for _, rg := range resourceGroups {
```
gtardif, line 173: You can probably avoid this loop, and use an ACI call that does not filter on resource group to get containers

ulyssessouza: I tried this already. The problem is that from the list of `container groups` I cannot know from which `resource group` it comes from.

aci/volumes.go
gtardif: maybe... not a blocker for this PR, we can add this later on if we want. cc @nebuk89

gtardif (approved):

gtardif:
rebased

PR #741, opened 2020-10-08
Simplify metrics gathering, separate running commands from getting help

gtardif:

**What I did**
* hardcode list or gathered metrics & simplify metrics code (generate command list walking through commands)
* separate metrics on `docker <COMMAND>` from  `docker <COMMAND> --help`

**Related issue**


metrics/generatecommands/main.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"github.com/docker/compose-cli/utils"
+)
+
+var managementCommands = []string{"ecs", "assemble", "registry", "template", "cluster", "scan"}
```
chris-crone, line 27: IIRC all of these except `ecs`, and `scan` were only ever in EE so we could remove them.

chris-crone (approved):

metrics/generatecommands/main.go
gtardif: Indeed, added here because I don't get them with Desktop, it's enterprise. I kept these because they were already listed as management commands , so we can at least know if some users are calling them (we won't get details with subcommands as I have not generated subcommands for `template` for example). 
@nebuk89, OK to just ignore assemble, registry, template, cluster ?

nebuk89: Yep, thats fine with me!

gtardif: Done

rumpl (approved):

Issue #743, opened 2020-10-08
ACI: improve resource limit / reservation passed to azure


gtardif:
When trying to set different values to memory limit and memory requests, we get the error :
containerinstance.ContainerGroupsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="ContainerLimitGreaterThanContainerGroupTotalRequest" Message="The 'MemoryInGB' limit '3' in container 'single--container--aci' is greater than the total request '2' of container group 'boring-stonebraker'."

from @macolso: 
> This is by design - ACI does not support resource limits and requests in the same way that k8s does. Resource limits allow users to control how much CPU and memory each container instance consumes from the container group's entire resource request. Therefore, the limit cannot be greater than the sum of resource requests within the container group.

> For example, imagine you have a container group with two container instances each requesting 1 CPU, but one of those containers runs a more CPU intensive application. You can set the resource limit for that container instance up to 2 CPUs, allowing it to consume the entire container groups CPU resources if need be.

We need to pass compose resource info (limits & reservations) to ACI when deploying an app

PR #748, opened 2020-10-08
Aci resource reservation & limits

gtardif:
**What I did**
* Add support for ACI resource requests and limits, knowing limits cannot be more than container group total requests (so this makes sense only for multi-container compose apps, otherwise requests == limits)
* updated inspect struct to group resource requests & limits, along with restart policy (align on existing inspect struct)
* ACI e2e test
* update gRPC inspect proto to reflect changes 
(slightly changing gRPC inspectResponse, should be OK in beta. modified fields are not used in VSCode extension currently using this API, cf https://github.com/microsoft/vscode-docker/blob/master/src/docker/DockerServeClient/DockerServeUtils.ts)

**Related issue**
https://github.com/docker/compose-cli/issues/743

/test-aci


aci/convert/convert.go
```diff
@@ -439,17 +470,26 @@ func fqdn(group containerinstance.ContainerGroup, region string) string {
...
 	cpuLimit := 0.
-	if cc.Resources != nil &&
-		cc.Resources.Limits != nil &&
-		cc.Resources.Limits.CPU != nil {
-		cpuLimit = *cc.Resources.Limits.CPU
+	cpuReservation := 0.
+	if cc.Resources != nil {
+		if cc.Resources.Limits != nil {
+			if cc.Resources.Limits.MemoryInGB != nil {
+				memLimits = *cc.Resources.Limits.MemoryInGB * 1024 * 1024 * 1024
```
ulyssessouza, line 479: nit: Use a convertion function for this and reuse for the other cases

aci/convert/convert_test.go
```diff
@@ -106,7 +108,13 @@ func TestContainerGroupToContainer(t *testing.T) {
 		Config: &containers.RuntimeConfig{
 			FQDN: "myapp.eastus.azurecontainer.io",
 		},
-		RestartPolicyCondition: "any",
+		HostConfig: &containers.HostConfig{
+			CPULimit:          3,
+			CPUReservation:    2,
+			MemoryLimit:       214748364,
```
ulyssessouza, line 114: nit: Use the same convertion function here to make it human readable/checkable

tests/composefiles/aci-demo/aci_demo_port_resources.yaml
```diff
@@ -0,0 +1,24 @@
...
+    ports:
+      - "80:80"
+    deploy:
+      resources:
+        limits:
+          cpus: '0.7'
+          memory: 1G
+        reservations:
+          cpus: '0.5'
+          memory: 0.5G
```
ulyssessouza, line 24: nit: \n

ulyssessouza (approved):
LGTM

PR #751, opened 2020-10-09
Fix e2e-aci tests on comparing with empty string

ulyssessouza:
**What I did**
Fix `e2e-aci` tests on comparing with empty string.
It's not that the old code is not "correct", but the comparison wasn't getting to the right assertion. When printing the value just before or comparing the empty string, that went back to work.

I'm not sure what happened here, maybe that's a GO specifics that I ignore. Anyways, IMHO comparing with empty string even looks more readable.

**Tests**
/test-aci


gtardif (approved):

gtardif:
rebased

Issue #663, opened 2020-09-25
Review & document ACI secret mapping


gtardif:
We have ACI secret support, but this was done before reviewing the volume implementation, there might be things to adjust / improve. 
- [x] Allow to map secrets to `run/secrets` (and not `run/secrets/subfolder`). We can do this by defining several secret values in ACI secret volumes (https://github.com/docker/compose-cli/blob/main/aci/convert/convert.go#L214), containers will see one secret file for each secret defined here, and we can map the entire thing to `/run/secret`. We need to be careful and not share the same secret volume with different containers, if they are not supposed to see all secrets, we'll need to define one volume per container.
- [x] remove `inline:` (https://github.com/docker/compose-cli/blob/main/aci/convert/convert.go#L52), secrets should not be define in compose files, these are not secrets anymore (use env properties or something else) 
- [x] add tests on ACI secret usage.

gtardif:
Currently secrets are mapped to subfolders of `/run/secrets`. 
- [ ] We could define several secret values inside a single ACI secret volume, and map this to `/run/secret`, the container then sees one file for each secret value. However we need to be careful to have different ACI secret volumes for containers that don't see the same secrets. 
- [ ] we should remove the `inline:` option to set secret value in the compose file directly. Secrets should not be written in compose files, these are not secrets anymore (use env variables instead)

ulyssessouza:
Implemented in https://github.com/docker/compose-cli/pull/753 and https://github.com/docker/compose-cli/pull/757

PR #757, opened 2020-10-12
Squash all secrets in a single one

ulyssessouza:
**What I did**

Squash all secrets in a single one
Also adds an e2e test

**Related issue**
https://github.com/docker/compose-cli/issues/663

**Tests**
/test-aci

### DIFF
--- a/aci/convert/convert.go
+++ b/aci/convert/convert.go
@@ -49,6 +49,8 @@ const (
 	volumeDriveroptsShareNameKey   = "share_name"
 	volumeDriveroptsAccountNameKey = "storage_account_name"
 	volumeReadOnly                 = "read_only"
+
+	serviceSecretPrefix = "aci-service-secret-"
 )
 
 // ToContainerGroup converts a compose project into a ACI container group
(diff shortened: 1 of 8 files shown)


aci/convert/convert.go
```diff
@@ -188,22 +190,30 @@ type projectAciHelper types.Project
...
+			data, err := ioutil.ReadFile(p.Secrets[scr.Source].File)
+			if err != nil {
+				return secretVolumes, err
+			}
+			if len(data) == 0 {
+				continue
+			}
+			dataStr := base64.StdEncoding.EncodeToString(data)
+			if scr.Target == "" {
+				scr.Target = scr.Source
```
gtardif, line 208: This will support onlt renaming the file, but target cannot be a path (i.e. include "/" or "\"). 
Fine for this PR, we should implement support path other than /run/secrets in a separate PR. Tested with @ulyssessouza, if we mount a secret for example with `target=/etc/nginx/conf.d/site.conf` then entire will be wiped by ACI to provide the secrets, secret files cannot be mixed with other files in the same folder

gtardif (commented):
You can also update https://github.com/docker/compose-cli/blob/main/docs/aci-compose-features.md an remove `Due to ACI secret volume mounting..`. since we are now mounting secrets the "compose way"

gtardif (approved):

PR #762, opened 2020-10-12
Revert "Temporarily disable some red ACI tests, to be put reverted on…

gtardif:
**What I did**
revert disabling some ACI tests since we are now back on the public azure platform

**Related issue**

/test-aci


ulyssessouza (approved):
LGTM

PR #770, opened 2020-10-13
Put back gRPC fields to avoid API breaking change with existing versi…

gtardif:
Breaking API change was introduced by https://github.com/docker/compose-cli/commit/cfc490fd92e834cb03290e034a93956b62a8777b

(Even though these fields are unused in VSCode Docker extension, removing them and re-using the fields number totally breaks APIs returning containers : ls, inspect)

**What I did**
* put back previous gRPC fields, and keep new field HostConfig at the end of the list. 
* Tested with node-sdk previous version

**Related issue**
previous version of node-sdk crashing when listing ACI containers


protos/containers/v1/containers.proto
```diff
@@ -45,12 +45,15 @@ message Container {
 	string command = 4;
 	uint64 cpu_time = 5;
 	uint64 memory_usage = 6;
-	uint64 pids_current = 7;
-	uint64 pids_limit = 8;
-	repeated string labels = 9;
-	repeated Port ports = 10;
-	string platform = 11;
-	HostConfig host_config = 12;
+	uint64 memory_limit = 7;
```
ulyssessouza, line 48: nit: Maybe we should add a comment to indicate that the outter fields are deprecated

ulyssessouza (approved):
LGTM

PR #771, opened 2020-10-13
Fix panic when invoking inspect with empty container id 

gtardif:
ACI SDK returns status 200 but nil container group properties

**What I did**
Check for empty container ID, and never dereference nil pointer

**Related issue**

/test-aci

### DIFF
--- a/aci/containers.go
+++ b/aci/containers.go
@@ -235,12 +235,15 @@ func (cs *aciContainerService) Delete(ctx context.Context, containerID string, r
 
 func (cs *aciContainerService) Inspect(ctx context.Context, containerID string) (containers.Container, error) {
 	groupName, containerName := getGroupAndContainerName(containerID)
+	if containerID == "" {
+		return containers.Container{}, errors.New("cannot inspect empty container ID")
+	}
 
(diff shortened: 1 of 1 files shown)


aci/containers.go
```diff
@@ -235,12 +235,15 @@ func (cs *aciContainerService) Delete(ctx context.Context, containerID string, r
...
+	if containerID == "" {
+		return containers.Container{}, errors.New("cannot inspect empty container ID")
+	}
 
 	cg, err := getACIContainerGroup(ctx, cs.ctx, groupName)
 	if err != nil {
 		return containers.Container{}, err
 	}
-	if cg.StatusCode == http.StatusNoContent {
+	if cg.StatusCode == http.StatusNoContent || cg.Containers == nil {
```
ulyssessouza, line 246: Here actually `*http.Response` can be `nil`. `autorest.Response` has a helper function for it. Also `*ContainerGroupProperties`:
```suggestion
	if cg.IsHTTPStatus(http.StatusNoContent) || cg.ContainerGroupProperties == nil || cg.Containers == nil {
```
Actually there is the same case in other parts of the file

ulyssessouza (changes requested):
Just an `if` check that can `panic`

Issue #669, opened 2020-08-12
Automate publication of docker/ecs-secrets-sidecar image


ndeloof:
We need to keep docker/ecs-secret-sidecar image in sync with github repo, so automate build and publication of this image.
We can introduce a dedicated Github Action workflow for this purpose, relying on docker/build-push-action

PR #772, opened 2020-10-13
Extract ACI_DNS_SIDECAR_IMAGE

ulyssessouza:
**What I did**
Propose the extraction of the ACI sidecar image specification to a "config" file then injected during the build. 
This would make it easier to bump versions afterwards.

**Tests**
/test-aci

### DIFF
--- a/aci/convert/convert.go
+++ b/aci/convert/convert.go
@@ -35,6 +35,7 @@ import (
 	"github.com/docker/compose-cli/api/compose"
 	"github.com/docker/compose-cli/api/containers"
 	"github.com/docker/compose-cli/context/store"
+	"github.com/docker/compose-cli/internal"
 	"github.com/docker/compose-cli/utils/formatter"
 )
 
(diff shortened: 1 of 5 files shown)


gtardif (changes requested):
I like isolating the images (and image version) so we have a central place for this, but having it in the code in a separate file seems fine, the rest seems overkill and more complex to understand the flow of data. 

I'd keep he ACIDNSSidecarImage in a separate go file as a const (in package internal or somewhere else near aci for now). 

Maybe later on we can regroup with ECS images (ecs-secrets-sidecar and ecs-searchdomain-sidecar) when we pin versions and do #669

chris-crone:
Related: https://github.com/docker/compose-cli/pull/799

gtardif:
I think now with pinned version on sidecar images and Dockerfile specific to each sidecar, this is obsolete and we can keep the current state, each sidecar has its own well scoped folder.

Issue #664, opened 2020-09-09
support ulimits


ledzeppelin:
Whenever I set something like
```
services:
  myapp:
    ulimits:
      nofile:
        soft: 9000
        hard: 10000
```

I get the warning `WARN[0000] services.ulimits: unsupported attribute`.  It seems like this is supported https://docs.amazonaws.cn/en_us/AmazonECS/latest/developerguide/AWS_Fargate.html
> Amazon ECS task definitions for Fargate support the `ulimits`

and that there's already some logic in https://github.com/docker/ecs-plugin/blob/master/pkg/amazon/backend/convert.go#L155 to parse it.

PR #779, opened 2020-10-14
add support for ulimits.nofile

ndeloof:
**What I did**
Claim support for `services.ulimits` - conversion was actually already implemented
According to ECS docs, fargate only allows use of `nofile` ulimit, so the custom CompatibilityChecker

**Related issue**
fix https://github.com/docker/compose-cli/issues/664

/test-ecs

image

### DIFF
--- a/ecs/compatibility.go
+++ b/ecs/compatibility.go
@@ -135,3 +135,12 @@ func (c *fargateCompatibilityChecker) CheckLoggingDriver(config *types.LoggingCo
 		c.Unsupported("services.logging.driver %s is not supported", config.Driver)
 	}
 }
+
+func (c *fargateCompatibilityChecker) CheckUlimits(service *types.ServiceConfig) {
+	for k := range service.Ulimits {
+		if k != "nofile" {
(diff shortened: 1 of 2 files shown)


ecs/compatibility.go
```diff
@@ -135,3 +135,12 @@ func (c *fargateCompatibilityChecker) CheckLoggingDriver(config *types.LoggingCo
 		c.Unsupported("services.logging.driver %s is not supported", config.Driver)
 	}
 }
+
+func (c *fargateCompatibilityChecker) CheckUlimits(service *types.ServiceConfig) {
```
gtardif, line 139: I don't get who calls these CheckXXX methods, is it AWS using reflection ? does CheckULimits implement an AWS interface ?

ndeloof: compose-go does: https://github.com/compose-spec/compose-go/blob/a70cd5945c2555b0d96d74cdecd67542331c9bdb/compatibility/checker.go#L364

by defining this (I'd say "override" if golang allowed it) we customize the default mechanism, which is just opt-in for supported attributes, so we can control allowed values

chris-crone (approved):

PR #782, opened 2020-10-14
Document cross build step reason & timing in CI

gtardif:

**What I did**
* Check CI exec timing and decided to keep full cross compile, rather than specifically compile for  darwin, since the  overall exec time will not be that different
* document why we have this step in CI

**Related issue**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,8 @@ jobs:
           path: ~/go/pkg/mod
           key: go-${{ hashFiles('**/go.sum') }}
 
+      # Ensure we don't discover cross platform build issues at release time.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -53,6 +53,8 @@ jobs:
           path: ~/go/pkg/mod
           key: go-${{ hashFiles('**/go.sum') }}
 
+      # Ensure we don't discover cross platform build issues at release time.
+      # time used to build linux here is gained back in the build for local E2E step
```
ulyssessouza, line 57:
```suggestion
      # Time used to build linux here is gained back in the build for local E2E step
```

PR #783, opened 2020-10-14
Let user pass EFS create option by driver_opts

### LINKED ISSUE #781: support EFS volume options
**Description**
As we create EFS filesytem as compose application volume, let user set EFS options
we can rely on `driver_opts` for this purpose.

ndeloof:
**What I did**

Allow user to pass EFS volume options by `driver_opts` in volume configuration

Move volume creation into cloudformation template with `Retain` policy so it survives `down`, reports on console as `DELETE_SKIPPED`

**Related issue**
fixes #781

/test-ecs

image

### DIFF
--- a/ecs/awsResources.go
+++ b/ecs/awsResources.go
@@ -19,15 +19,17 @@ package ecs
 import (
 	"context"
 	"fmt"
+	"strconv"
+	"strings"
 
 	"github.com/docker/compose-cli/api/compose"
-
 	"github.com/docker/compose-cli/errdefs"
 
 	"github.com/aws/aws-sdk-go/service/elbv2"
 	"github.com/awslabs/goformation/v4/cloudformation"
 	"github.com/awslabs/goformation/v4/cloudformation/ec2"
 	"github.com/awslabs/goformation/v4/cloudformation/ecs"
+	"github.com/awslabs/goformation/v4/cloudformation/efs"
 	"github.com/awslabs/goformation/v4/cloudformation/elasticloadbalancingv2"
 	"github.com/compose-spec/compose-go/types"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


gtardif (approved):

arauchberger:
hi @ndeloof ,
i'm currently stuck with my "docker compose goes AWS ECS" experiment.
to make it more precise: i have volumes declared, which get created, but no files can be written to the created AWS EFS later.

i'm not sure about the available driver_opts. can you point me to some documentation?

what i have declared in my docker-compose.yml is the following:

```yml
volumes:
  certs:
    #name: fs-025fdebaf395e8779
    driver_opts:
      uid: 8888
      gid: 8888
      permissions: 0777
```
i can see that the EFS and an AccessPoint is created (during docker compose up), the AP also has PosixUser, but no `CreationInfo` with ownerUID/GID. i think this is the root-cause for failing to write to the FS.

```yml
Resources:
  CertsAccessPoint:
    Properties:
      AccessPointTags:
        - Key: com.docker.compose.project
          Value: pitportal-patient
        - Key: com.docker.compose.volume
          Value: certs
        - Key: Name
          Value: pitportal-patient_certs
      FileSystemId:
        Ref: CertsFilesystem
      PosixUser:
        Gid: "8888"
        Uid: "8888"
    Type: AWS::EFS::AccessPoint
  CertsFilesystem:
    DeletionPolicy: Retain
    Properties:
      Encrypted: true
      FileSystemTags:
        - Key: com.docker.compose.project
          Value: pitportal-patient
        - Key: com.docker.compose.volume
          Value: certs
        - Key: Name
          Value: pitportal-patient_certs
    Type: AWS::EFS::FileSystem
```

a second question:
why is it not possible to reference an existing recreated EFS-FS/AP?
i tried but also failed. the `name` seems to be ignored and a new EFS is created instead.


any help highly appreciated 
thanks Andi

Issue #790, opened 2020-10-15
docker ps (with docker binar 0.1.22) throws an exception  


mreferre:
BUG REPORT INFORMATION
---------------------------------------------------

I am testing a compose file whose content is as follows ( all account IDs and ARNs have bene faked): 
```
version: "3.8"

x-aws-cluster: "docker2ecs-cluster" # my cluster 
x-aws-vpc: vpc-hdfgd67hjd # my VPC (default)
x-aws-loadbalancer: "arn:aws:elasticloadbalancing:us-west-2:xxxxxxxxx:loadbalancer/app/docker2ecs-LB/hdgshteykkl" # my ALB

services:
  docker2ecs:
    image: mreferre/docker2ecs:latest
    ports:
      - 80:80
    x-aws-pull_credentials: "arn:aws:secretsmanager:us-west-2:xxxxxxxxx:secret:dockerhubAccessToken-hdhKdk"
    x-aws-logs_retention: 30
    x-aws-policies: 
      - "arn:aws:iam::aws:policy/AdministratorAccess" # IAM role is created with an Admin policy and assigned to the container
    deploy:
      replicas: 1
      update_config:
        x-aws-min_percent: 50 
        x-aws-max_percent: 300         
    secrets: 
      - preferences # my secret injected into the container (@ /run/secrets/preferences)
    networks: 
      - dmz
    
secrets:
  preferences:
    name: "arn:aws:secretsmanager:us-west-2:xxxxxxxxx:secret:PREFERENCES-Jhdhgdf" # my secret
    external: true

networks:
  dmz:
    external: true
    name: "sg-0179xxxxx9f29" # my SG (allow port 80) 
```

I am using Desktop Edge (latest which is version `2.4.1.0`). I am also brute forcing updating my setup to use the latest `docker` CLI release available today (`0.1.22`) on this repo. 

This stack deploys fine but a subsequent `docker compose ps` does not work (see below)

**Steps to reproduce the issue:**
```
➜  Downloads ./dockercloud/docker/docker compose up
[+] Running 11/11
 ⠿ downloads                        CREATE_COMPLETE                                                                                                                                 147.0s
 ⠿ Docker2ecsTCP80TargetGroup       CREATE_COMPLETE                                                                                                                                   0.0s
 ⠿ Docker2ecsTaskExecutionRole      CREATE_COMPLETE                                                                                                                                  18.0s
 ⠿ CloudMap                         CREATE_COMPLETE                                                                                                                                  47.0s
[...]

PR #795, opened 2020-10-16
Store external cluster as metadata

ndeloof:
**What I did**
1. add a sanity check to sdk.DescribeService
2. store external cluster as template metadata so we can get this information back

DescribeService API, even we pass a service ARN, require cluster ID (?)
If cluster was set by user as `x-aws-cluster`, we can't get this ID from CloudFormation stack resources
Proposed change uses CloudFormation template Metadata to track the user-provided cluster so we can easily get this ID back on `compose ps`

**Related issue**
https://github.com/docker/compose-cli/issues/790

/test-ecs

image

### DIFF
--- a/ecs/cloudformation.go
+++ b/ecs/cloudformation.go
@@ -52,12 +52,12 @@ func (b *ecsAPIService) convert(ctx context.Context, project *types.Project) (*c
 		return nil, err
 	}
 
-	resources, err := b.parse(ctx, project)
+	template := cloudformation.NewTemplate()
+	resources, err := b.parse(ctx, project, template)
 	if err != nil {
 		return nil, err
 	}
 
-	template := cloudformation.NewTemplate()
 	err = b.ensureResources(&resources, project, template)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


ecs/cloudformation.go
```diff
@@ -43,6 +43,10 @@ func (b *ecsAPIService) Convert(ctx context.Context, project *types.Project) ([]
 		return nil, err
 	}
 
+	if c, ok := project.Extensions[extensionCluster]; ok {
```
gtardif, line 46: Any reason to add this here rather than in b.convert() ? 
And maybe that will integrate better with unit tests, add a unit test on this new metedata field

gtardif (commented):
LGTM, naive question about where the metadata field is added. 
Also, we could have caught this with a `compose ps` invocation in the ECS E2E tests, can we add it now ?

ecs/cloudformation.go
ndeloof: nope, will change this

gtardif (approved):

PR #800, opened 2020-10-16
Fix unit tests on windows: for secret target path, use path.Join etc. instead of filepath.Join

gtardif:
as these paths are container side, not cli side. (do not work with windows paths)

Signed-off-by: guillaume.tardif <guillaume.tardif@gmail.com>

**What I did**
Replace `import filepath` by `import path` ; `path` works with "/" as file separate and is not os sensitive

**Related issue**
https://github.com/docker/compose-cli/runs/1264272389

/test-windows

### DIFF
--- a/aci/convert/convert_test.go
+++ b/aci/convert/convert_test.go
@@ -21,7 +21,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path/filepath"
+	"path"
 	"testing"
 
 	"github.com/stretchr/testify/mock"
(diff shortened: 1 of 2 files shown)


aci/convert/convert_test.go
```diff
@@ -724,7 +724,7 @@ func TestConvertSecrets(t *testing.T) {
 	_, err = tmpFile.Write([]byte("test content"))
 	assert.NilError(t, err)
 	t.Cleanup(func() {
-		assert.NilError(t, os.Remove(tmpFile.Name()))
+		_ = os.Remove(tmpFile.Name())
```
gtardif, line 727: failure on windows. File should be closed after reading, this is handled OK by File.ReadAll(), but I suppose windows is doing this async (not the first time I see this) and the test cleanup is arriving too quickly and racing with the actual file closing...

chris-crone (approved):

aci/convert/convert_test.go
StefanScherer: Or maybe the Windows Defender is grabbing new files? 🤔
I could check if we have disabled it.

gtardif: ? maybe (on Github Action windows nodes ? ) I could reproduce this on my windows laptop

StefanScherer: ok, that was a reflex, I thought it is Jenkins one of the OVH machines 😂

Issue #739, opened 2020-10-08
Use EFS access points for volumes


ndeloof:
**Description**
Use EFS filesystems as compose volumes using `access points`

see https://aws.amazon.com/fr/blogs/containers/developers-guide-to-using-amazon-efs-with-amazon-ecs-and-aws-fargate-part-2/ and
https://aws.amazon.com/fr/blogs/containers/developers-guide-to-using-amazon-efs-with-amazon-ecs-and-aws-fargate-part-3/

ndeloof:
blocked by https://github.com/awslabs/goformation/issues/327

PR #812, opened 2020-10-19
Introduce use of EFS Access Point to mount filesystems as volumes

ndeloof:
**What I did**
Generate an [AccessPoint](https://aws.amazon.com/fr/s3/features/access-points/) for compose file's volumes so that binding can be customized:
- add support for `uid`,`gid` to let container access a volume without entering the file permission hell
- add support for `root_directory` so volume can actually point to a sub-path within the filesystem
- add support for creating such a sub-path for uid,gid + permissions

How to test:

Create a compose file to create filesystem as root: 
```yaml
services:
  server:
    image: jenkins/jenkins:lts
    user: root
    volumes:
      - config:/var/jenkins_home
volumes:
  config: {}
```

Then update compose file to remove the root user and double check filesystem still can be mounted read/writen by container without posix permission issues

```yaml
services:
  server:
    image: jenkins/jenkins:lts
    volumes:
      - config:/var/jenkins_home
volumes:
  config: 
    driver_opts:
      uid: 0
      gid: 0
```

This PR also introduce `awsResource` interface as an attempt to abstract references to AWS resources by ID or ARN, which depend on the resource type and API in use:
- some only use IDs (vpc)
- some can use both (load balancers)
- some require ARN (iam)
Also handle the heterogeneous behaviour CloudFormation resources regarding `!Ref`, which for some returns `ARN`, otherwise `ID` ¯\_(ツ)_/¯

**Related issue**
https://github.com/docker/compose-cli/issues/739

/test-ecs

image

### DIFF
--- a/ecs/backend.go
+++ b/ecs/backend.go
@@ -19,9 +19,6 @@ package ecs
 import (
 	"context"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/session"
-
 	"github.com/docker/compose-cli/api/compose"
 	"github.com/docker/compose-cli/api/containers"
 	"github.com/docker/compose-cli/api/resources"
@@ -32,6 +29,9 @@ import (
 	"github.com/docker/compose-cli/context/cloud"
 	"github.com/docker/compose-cli/context/store"
 	"github.com/docker/compose-cli/errdefs"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
 )
 
 const backendType = store.EcsContextType
--- a/ecs/volumes.go
+++ b/ecs/volumes.go
@@ -19,6 +19,8 @@ package ecs
 import (
 	"fmt"
 
+	"github.com/docker/compose-cli/api/compose"
+
 	"github.com/awslabs/goformation/v4/cloudformation"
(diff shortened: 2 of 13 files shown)


ecs/backend.go
```diff
@@ -32,6 +29,9 @@ import (
 	"github.com/docker/compose-cli/context/cloud"
 	"github.com/docker/compose-cli/context/store"
 	"github.com/docker/compose-cli/errdefs"
+
```
chris-crone, line 32: goimports, this should be above our local imports

ecs/volumes.go
```diff
@@ -19,6 +19,8 @@ package ecs
 import (
 	"fmt"
 
+	"github.com/docker/compose-cli/api/compose"
```
chris-crone, line 22: goimports, this should be below the other imports

ecs/backend.go
ndeloof: I like it so much spending energy with such stupid things :P

chris-crone: 😬

I'm very nitpicky about code style and my editor does this magically for me.

ndeloof: Goland also manage imports but from time to time new imports don't get introduce in the right place. I wish `goimports` had a mode to _ignore_ sections and just order things in canonical way os we don't have to worry about this.

chris-crone (approved):

Issue #742, opened 2020-10-08
Introduce volume command


ndeloof:
**Description**
With support for volume creation (https://github.com/docker/compose-cli/issues/737) we need a command to also manage EFS filesystems.

For consistency with the docker cli such commands should adopt `docker volume create|inspect|ls|rm` UX

PR #813, opened 2020-10-19
CLI command to manage ECS volumes

ndeloof:
**What I did**
implement VolumeService on ECS/EFS and removed ACI-only restriction on `volume` commands 

**Related issue**
https://github.com/docker/compose-cli/issues/742

image

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -182,12 +182,8 @@ func main() {
 	root.AddCommand(
 		run.Command(ctype),
 		compose.Command(ctype),
+		volume.Command(ctype),
 	)
-
-	if ctype == store.AciContextType {
-		// we can also pass ctype as a parameter to the volume command and customize subcommands, flags, etc. when we have other backend implementations
-		root.AddCommand(volume.ACICommand())
-	}
 
 	ctx = apicontext.WithCurrentContext(ctx, currentContext)
 	ctx = store.WithContextStore(ctx, s)
--- a/ecs/volumes.go
+++ b/ecs/volumes.go
@@ -17,13 +17,17 @@
 package ecs
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/docker/compose-cli/api/compose"
+	"github.com/docker/compose-cli/api/volumes"
+	"github.com/docker/compose-cli/errdefs"
 
 	"github.com/awslabs/goformation/v4/cloudformation"
 	"github.com/awslabs/goformation/v4/cloudformation/efs"
 	"github.com/compose-spec/compose-go/types"
+	"github.com/pkg/errors"
 )
 
 func (b *ecsAPIService) createNFSMountTarget(project *types.Project, resources awsResources, template *cloudformation.Template) {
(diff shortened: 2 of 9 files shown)


cli/cmd/volume/acivolume.go
```diff
@@ -20,6 +20,10 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/docker/compose-cli/ecs"
+
+	"github.com/docker/compose-cli/context/store"
```
chris-crone, line 25: goimports: These should be in the bottom section of the import list

cli/cmd/volume/acivolume.go
```diff
@@ -20,6 +20,10 @@ import (
 	"context"
```
chris-crone, line 20: I don't think the ECS code should be in the `acivolume.go` file. Either we need to make it generic (and name it something else) or we need to add a `volume_ecs.go` file.

cli/cmd/volume/acivolume.go
```diff
@@ -30,27 +34,27 @@ import (
...
 }
 
-func createVolume() *cobra.Command {
-	aciOpts := aci.VolumeCreateOptions{}
+func createVolume(ctype string) *cobra.Command {
+	var opts interface{}
 	cmd := &cobra.Command{
-		Use:   "create --storage-account ACCOUNT VOLUME",
-		Short: "Creates an Azure file share to use as ACI volume.",
+		Use:   "create [OPTIONS] VOLUME",
```
chris-crone, line 56: Could we have a different help message based on the context?

cli/main.go
```diff
@@ -182,13 +182,9 @@ func main() {
 	root.AddCommand(
 		run.Command(ctype),
 		compose.Command(ctype),
+		volume.Command(ctype),
```
chris-crone, line 185: What happens if you try to use the moby or example type contexts?

ecs/volumes.go
```diff
@@ -17,8 +17,14 @@
 package ecs
 
 import (
+	"context"
 	"fmt"
 
+	"github.com/docker/compose-cli/errdefs"
+	"github.com/pkg/errors"
+
+	"github.com/docker/compose-cli/api/volumes"
```
chris-crone, line 26: goimports: local in the bottom section, others globbed together

cli/cmd/volume/acivolume.go
ndeloof: right, I forgot about the file name here

cli/cmd/volume/acivolume.go
ndeloof: Wasn't the intent to have a uniform UX ?
Otherwise I don't see the benefit to have shared code and make each and every line context-conditional :P

cli/main.go
ndeloof: moby backend just pass parameter to classic CLI
example will get a VolumeService stub

cli/cmd/volume/acivolume.go
chris-crone: We want... both!

The UX should be as similar as possible but we want to help users do the right thing where we can. For ACI, we need them to specify the storage account on volume creation so hinting at that in the help would help users.

The failure case here isn't so bad; the user will not specify it and get an error saying please specify it.

chris-crone (approved):
LGTM

Issue #754, opened 2020-10-09
CLI issues with ZSH on OS X: "com.docker.cli: executable file not found"


bwateratmsft:
**Description**

Several users using ZSH on OS X have reported an error when running commands in VSCode (https://github.com/microsoft/vscode-docker/issues/2366). The commands run are typically done with `zsh -c 'docker build ...'` internally by VSCode.

Related:
https://github.com/docker/for-mac/issues/4956

**Steps to reproduce the issue:**
1. Install Docker Desktop for Mac 2.4.0.0.
2. Run `zsh -c 'docker some command'`

**Describe the results you received:**
`"exec: "com.docker.cli": executable file not found in $PATH".`

**Describe the results you expected:**
Successful execution of the command

Workaround:
Disabling the "Cloud Experience" in settings has worked for these users.

gtardif:
@sergeyk @HannesOberreiter maybe you will have some useful information to help us investigate further. 
If you can still reproduce the issue (after activating "Enable cloud experience" in Docker Desktop preferences), could you try to run these commands from a terminal: 
```
zsh -c 'which docker'
zsh -c 'which com.docker.cli'
# These 2 commands should be under /usr/local/bin/ with Docker Desktop setup

# Don't copy the entire PATH value in this ticket, just check if it contains /usr/local/bin/
zsh -c 'echo $PATH'

# to reproduce the issue
zsh -c 'docker ps'
```
Thanks for your feedback

HannesOberreiter:
@gtardif I get the same error after activating "Enable cloud experience" again.

```bash
> Executing task: docker run --rm -it  -p 8888:8888/tcp hannesoberreiter/tug-python362:latest <

exec: "com.docker.cli": executable file not found in $PATH
The terminal process "/bin/zsh '-c', 'docker run --rm -it  -p 8888:8888/tcp hannesoberreiter/tug-python362:latest'" failed to launch (exit code: 1).

Terminal will be reused by tasks, press any key to close it.
```

You test seems fine and the command `zsh -c 'docker ps'` also works in the integrated console.

```bash
> zsh -c 'which docker'
/usr/local/bin/docker
> zsh -c 'which com.docker.cli'
/usr/local/bin/com.docker.cli
> zsh -c 'echo $PATH' | grep "/usr/local/bin" -o
/usr/local/bin
> zsh -c 'docker ps'
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
```

[...]

PR #825, opened 2020-10-23
Fix PATH when classic CLI is cannot be found

ulyssessouza:
**What I did**

As the 2 binaries are meant to be in the same
folder, when it cannot be found in the PATH,
it adds the executable's folder in it so
the other binary can be found.

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/754

/test-windows

### DIFF
--- a/cli/mobycli/exec.go
+++ b/cli/mobycli/exec.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


gtardif:
Proposed a new commit to amend the PATH only when shelling out to moby cli.
Also added /test-windows in the PR description. We'll need a e2e test for this (next to `TestMissingExistingCLI` probably)

gtardif:
Actually already implemented [here](https://github.com/docker/compose-cli/blob/main/cli/main.go#L76), I couldn't make an e2e test and make sure it fails without this PR fix...

ulyssessouza:
@gtardif But [this](https://github.com/docker/compose-cli/blob/main/cli/main.go#L76) is not multiplatform. I think it should fail on `windows` because of the path separator.
And anyways, looks like this block of code is older than the bug report. So looks like the problem will persist.

PR #827, opened 2020-10-26
Removing API deprecated fields (unused in VSCode Docker extension)

gtardif:
Keeping field numbering to not break all gRPC calls.
VSCode Docker extension is not using these fields, API is still in beta, we should clean this now.

**What I did**
* Remove deprecated fields from proto container struct without changing field numbering
* test with current node-sdk client this does not break client. client will see zero values (& empty strings) for the removed fields, if trying to access them.

**Related issue**


gtardif:
cc @karolz-ms or @bwateratmsft just to double check, I don't think you are using these fields, and I tried the current node-sdk js client to list containers, get values, etc., everything runs smoothly. 
If you see any issue with this change (or if you want to try running a local `docker serve --address xxx` and connect VSCode extension to it on a custom socket)
Of course this will be propagated to node-sdk, but I mainly want to check we can clean this while the API is still in beta, and not break VSCode extension that is the main known user for this API

bwateratmsft:
> cc @karolz-ms or @bwateratmsft just to double check, I don't think you are using these fields, and I tried the current node-sdk js client to list containers, get values, etc., everything runs smoothly.
> If you see any issue with this change (or if you want to try running a local `docker serve --address xxx` and connect VSCode extension to it on a custom socket)
> Of course this will be propagated to node-sdk, but I mainly want to check we can clean this while the API is still in beta, and not break VSCode extension that is the main known user for this API

The change looks good to me, thanks for the heads up!

karolz-ms:
> 
> The change looks good to me, thanks for the heads up!

Ship it ☺️

chris-crone (approved):
LGTM

We should cut a v2 of the photo interface at some point that cleans out these fields.

PR #843, opened 2020-10-28
Pin version of ACI sidecar image, so that we don’t affect prod users when re-publishing :latest.

gtardif:

**What I did**
* pined ACI sidecar image
* updated Makefile: new target to publish new versions, ensuring we don’t override an existing image tag

**Related issue**
Will need something similar fo ECS sidecars ; I'd like to add more E2E tests on ECS to be sure not to break anything, will do in a separate PR

/test-aci


chris-crone (approved):
LGTM

Issue #759, opened 2020-10-12
ACI Healthcheck support


gtardif:
ACI [support health checks](https://docs.microsoft.com/en-us/azure/container-instances/container-instances-liveness-probe), we can see how to map this to [Docker healthChecks](https://docs.docker.com/compose/compose-file/#healthcheck).
Mapping might not be 1 to 1 for all attributes, but that would be useful, even with some limitations on what attributes are supported and how the docker health checks are converted to ACI attributes.

PR #845, opened 2020-10-29
Add Healthchecks to ACI

ulyssessouza:
**What I did**
Add Healthchecks to ACI

**Related issue**
https://github.com/docker/compose-cli/issues/759

/test-aci


karolz-ms:
@macolso @vturecek @bigmorty FYI

@ulyssessouza ❤️

aci/convert/convert.go
```diff
@@ -310,6 +350,22 @@ func ContainerGroupToContainer(containerID string, cg containerinstance.Containe
...
+	healthcheckDisabled := true
+	var healthcheckTest []string
+	var healthcheckInterval types.Duration
+	if cc.ReadinessProbe != nil &&
+		cc.ReadinessProbe.Exec != nil &&
+		cc.ReadinessProbe.Exec.Command != nil {
+		if len(*cc.ReadinessProbe.Exec.Command) > 0 {
+			healthcheckDisabled = false
+			healthcheckTest = *cc.ReadinessProbe.Exec.Command
+			if cc.ReadinessProbe.PeriodSeconds != nil {
```
gtardif, line 363: what happens if we set no interval ?

ulyssessouza: Then it will be an ACI default

BigMorty:
Great stuff! What does the syntax look like for using this?

ulyssessouza:
Hello @BigMorty ! The idea is to stick to the compose's healthcheck as in [this little doc for healthcheck](https://github.com/docker/compose-cli/pull/845/files#diff-ef301724409370ac0b48fc8cc6b9d148e949bc16eadec38cdaa62787295ba57bR213-R229). 
Another source is the [healthcheck section](https://github.com/compose-spec/compose-spec/blob/master/spec.md#healthcheck) on compose-spec.

This PR also adds `docker run` [options](https://github.com/docker/compose-cli/pull/845/files#diff-8f2ea717bac797052e1ab04f40face0b054ee557b6a79d3059b92dd4e5ad943bR12-R16) to specify the same fields for one-off containers.

aci/convert/convert.go
```diff
@@ -237,6 +275,35 @@ func (s serviceConfigAciHelper) getResourceRequestsLimits() (*containerinstance.
...
+		testArray := s.HealthCheck.Test
+		switch s.HealthCheck.Test[0] {
+		case "NONE", "CMD", "CMD-SHELL":
+			testArray = s.HealthCheck.Test[1:]
+		}
+		if len(testArray) == 0 {
+			return nil
+		}
+
+		var retries int32 = 10
```
gtardif, line 289: magic number 10, is this a docker default ? Maybe use a const with a clear name for this

aci/convert/convert.go
```diff
@@ -68,7 +69,7 @@ func ToContainerGroup(ctx context.Context, aciContext store.AciContext, p types.
 		return containerinstance.ContainerGroup{}, err
 	}
 
-	var containers []containerinstance.Container
+	var ctnrs []containerinstance.Container
```
gtardif, line 72: nit: keep name "containers", more readable

aci/convert/convert.go
```diff
@@ -181,6 +209,16 @@ func (s serviceConfigAciHelper) getAciContainer(volumesCache map[string]bool) (c
...
+			LivenessProbe: &containerinstance.ContainerProbe{
+				Exec: &containerinstance.ContainerExec{
+					Command: healthTest,
+				},
+				InitialDelaySeconds: healthStartPeriod,
+				PeriodSeconds:       healthInterval,
+				FailureThreshold:    healthRetries,
+				SuccessThreshold:    healthRetries,
+				TimeoutSeconds:      healthTimeout,
+			},
```
gtardif, line 221: This and the above variables should be replaced by `getLivenessProbe()`, or am I missing something ?

aci/convert/convert.go
```diff
@@ -110,7 +111,9 @@ func ToContainerGroup(ctx context.Context, aciContext store.AciContext, p types.
 			dnsLabelName = serviceDomainName
 		}
 
-		containers = append(containers, containerDefinition)
+		containerDefinition.LivenessProbe = service.getLivenessProbe()
```
gtardif, line 114: move the call of getLivenessProbe() into getAciContainer() (called above, L93)

aci/convert/convert.go
```diff
@@ -310,6 +377,39 @@ func ContainerGroupToContainer(containerID string, cg containerinstance.Containe
...
 		Env:  envVars,
 	}
+
+	var (
+		healthcheckDisabled bool = true
+		healthcheckTest     []string
+		healthcheckInterval types.Duration
+		healthRetries       int
+		healthTimeout       types.Duration
+		healthStartPeriod   types.Duration
```
gtardif, line 387: just define one var, type containers.HealthCheck ?

cli/cmd/run/run.go
```diff
@@ -57,8 +58,14 @@ func Command(contextType string) *cobra.Command {
 	cmd.Flags().VarP(&opts.Memory, "memory", "m", "Memory limit")
 	cmd.Flags().StringArrayVarP(&opts.Environment, "env", "e", []string{}, "Set environment variables")
 	cmd.Flags().StringArrayVar(&opts.EnvironmentFiles, "env-file", []string{}, "Path to environment files to be translated as environment variables")
-	cmd.Flags().StringVarP(&opts.RestartPolicyCondition, "restart", "", containers.RestartPolicyNone, "Restart policy to apply when a container exits")
+	cmd.Flags().StringVarP(&opts.RestartPolicyCondition, "restart", "", containers.RestartPolicyNone, "Restart policy to apply when a container exits (none|any|on-failure)")
```
gtardif, line 61: Maybe wait on this help change, we should do https://github.com/docker/compose-cli/issues/877

aci/convert/convert.go
ulyssessouza: This was to avoid the conflict with the package name `containers`. My IDE tells me all the time

aci/convert/convert.go
ulyssessouza: This is completely arbitrary. I was also thinking about just letting it as `nil` if not explicitly set. WDYT @gtardif ?

gtardif: I would let the ACI default apply and not specify anything if the user does not

aci/convert/convert.go
gtardif: ah, OK, fine

aci/convert/convert.go
ulyssessouza: Wow good catch! I totally missed this one! Thanx!

aci/convert/convert.go
ulyssessouza: Totally!

aci/convert/convert.go
ulyssessouza: Done

cli/cmd/run/run.go
ulyssessouza: Reverting this

gtardif:
Code and functionality LGTM, we can see healthchecks happening at the right frequency, container restart as expected. 

Unfortunately, we are experiencing many deployments failures when specifying healthchecks. The same command invocation will sometimes succeed, sometimes fail (example: `$ 
./bin/docker --context acicontext run -p 80:80 --restart any --health-cmd "curl http://localhost:80" --health-interval 3s  nginx`). In case of failure, the error messge is : 
```
Message="The async operation failed." AdditionalInfo=[{"error":{"message":""},"id":"/subscriptions/5dfb2c88-af2d-47e6-96da-22ee4be68587/resourceGroups/gtardif/providers/Microsoft.ContainerInstance/containerGroups/gifted-joliot","properties":{"events":[]},"startTime":"2020-11-05T09:04:20.046142Z","status":"Failed"}]
```

The azure console shows a failure event with `The resource operation completed with terminal provisioning state 'Failed'.`
We don't experience these failures when not specifying healthchecks.

@macolso @novinc if you can investigate a failed deployment, here is an ID that I'll keep available for a while: 
/resource/subscriptions/5dfb2c88-af2d-47e6-96da-22ee4be68587/resourceGroups/gtardif/providers/Microsoft.ContainerInstance/containerGroups/gifted-joliot/overview

The failure rate is quite high (a quick manual check on 10 deployments with the above command, 50% failures), we need to better understand what is happening in these cases.

gtardif:
just rebased

gtardif:
Last commit fixes the deployment issues, were dues to specifying zero values instead of leaving empty fields in LivenessProbe payload. Zero values were sometimes rejected.

gtardif (approved):

PR #850, opened 2020-10-29
Fix path appending

gtardif:
**What I did**
* windows PATH should use windows separator (;)
* do not add separator if initial PATH is empty (was creating PATH=:/path)

**Related issue**
Maybe fixing https://github.com/docker/compose-cli/issues/754, but not fixing the Konvoy scenario

/test-aci
/test-windows

### DIFF
--- a/tests/aci-e2e/e2e-aci_test.go
+++ b/tests/aci-e2e/e2e-aci_test.go
@@ -434,14 +434,7 @@ func TestContainerRunAttached(t *testing.T) {
 		endpoint = fmt.Sprintf("http://%s:%d", fqdn, port.HostPort)
 
 		assert.Assert(t, !strings.Contains(followLogsProcess.Stdout(), "/test"))
-		checkRequest := func(t poll.LogT) poll.Result {
-			r, _ := http.Get(endpoint + "/test")
-			if r != nil && r.StatusCode == http.StatusNotFound {
-				return poll.Success()
-			}
-			return poll.Continue("waiting for container to serve request")
-		}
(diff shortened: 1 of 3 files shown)


tests/aci-e2e/e2e-aci_test.go
```diff
@@ -434,14 +434,7 @@ func TestContainerRunAttached(t *testing.T) {
...
 		assert.Assert(t, !strings.Contains(followLogsProcess.Stdout(), "/test"))
-		checkRequest := func(t poll.LogT) poll.Result {
-			r, _ := http.Get(endpoint + "/test")
-			if r != nil && r.StatusCode == http.StatusNotFound {
-				return poll.Success()
-			}
-			return poll.Continue("waiting for container to serve request")
-		}
-		poll.WaitOn(t, checkRequest, poll.WithDelay(1*time.Second), poll.WithTimeout(60*time.Second))
+		HTTPGetWithRetry(t, endpoint+"/test", http.StatusNotFound, 2*time.Second, 60*time.Second)
```
ulyssessouza, line 437: nit:
```suggestion
		HTTPGetWithRetry(t, endpoint+"/test", http.StatusNotFound, 2*time.Second, time.Minute)
```

ulyssessouza (approved):
LGTM!

Just a little `nit`

tests/aci-e2e/e2e-aci_test.go
gtardif: Ah yes, we could use time.Minute ; but to revisit everwhere to be consistent, we have several duration set a 60 secs, or more (120, ...). I'll leave it as it is here for consistency with the rest

PR #854, opened 2020-10-30
Revert breaking change on `docker context ls —format “{{ json . }}` 

gtardif:
this is used by VSCode Docker extension
VSCode should move to using `docker context ls --format json` and parse a regular json array. But for existing deployment, and not introduce breaking changes, `docker context ls —format "{{ json . }}"` will return json lines as before

**What I did**
* Support old format with a specific case, returning json lines and not a full json array; only in the case of the legacy `docker context ls —format "{{ json . }}"` (space variations allowed in the format value)

**Related issue**
https://github.com/microsoft/vscode-docker/issues/2452

### DIFF
--- a/cli/cmd/context/ls.go
+++ b/cli/cmd/context/ls.go
@@ -70,7 +70,7 @@ func runList(cmd *cobra.Command, opts lsOpts) error {
 		return err
 	}
 	format := strings.ToLower(strings.ReplaceAll(opts.format, " ", ""))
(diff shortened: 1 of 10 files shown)


gtardif:
double checked the VSCode Docker extension : reproduced the issue and checked it is fixed with this PR

chris-crone (approved):
LGTM

gtardif:
Also had to rename the new field providing context type in the output of `docker context ls —format "{{ json . }}"`, was named "Type" and for some reason this broke the VSCode context switching. Renamed to "ContextType", as is is new anyway

PR #860, opened 2020-11-02
Nicer cleanup make target for ACI resources

gtardif:

**What I did**
* Display resource groups that are going to be cleaned up
* grep ignore case, sometimes we have case difference in test resource group names

**Related issue**


ulyssessouza (approved):
LGTM

PR #867, opened 2020-11-03
Local backend improvements

chris-crone:
**What I did**
* Set env vars on run
* Set restart policy on run
* Set memory/CPU limits on run
* Improved inspect
* Add rudimentary volume support
* Add command support

### DIFF
--- a/local/backend.go
+++ b/local/backend.go
@@ -145,6 +145,7 @@ func (ms *local) Run(ctx context.Context, r containers.ContainerConfig) error {
 	containerConfig := &container.Config{
 		Image:        r.Image,
 		Labels:       r.Labels,
+		Env:          r.Environment,
 		ExposedPorts: exposedPorts,
 	}
 	hostConfig := &container.HostConfig{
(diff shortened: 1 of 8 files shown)


local/backend.go
```diff
@@ -100,15 +102,78 @@ func (ms *local) Inspect(ctx context.Context, id string) (containers.Container,
...
+		return nil
+	}
+
+	var restartPolicy string
+	switch m.HostConfig.RestartPolicy.Name {
+	case "always":
+		restartPolicy = containers.RestartPolicyAny
+	case "on-failure":
+		restartPolicy = containers.RestartPolicyOnFailure
+	case "no", "":
```
gtardif, line 165: should we remove this case  and simply go to default ? ("" is default already, it's not really useful from a doc standpoint anyway)

gtardif (approved):
Minor nit, LGTM

local/backend.go
chris-crone: I thought about that but decided to be explicit instead and used the values from here: https://github.com/moby/moby/blob/v1.13.1/api/types/container/host_config.go#L198-L200

ndeloof (approved):

gtardif (approved):

PR #878, opened 2020-11-04
Remove obsolete flag `docker context ls --json` (that was marked hidden)

gtardif:
This had been kept for VSCode extension but is not used, `docker context ls --format “{{ json . }}”` is used in the extension.

**What I did**
* remove hiddent obsolete flag
* double check in VSCode extension nothing is broken

**Related issue**


ndeloof:
Why remove this?
For CLI homogeneity, having `--format json` for all command with output makes sense to me

chris-crone:
@ndeloof this is the `--json` flag which we don't have anywhere else, not `--format json` or `--format {{ json . }}

chris-crone (approved):
LGTM

ndeloof (approved):

PR #879, opened 2020-11-04
Flag to import access key & secret from file

ndeloof:
**What I did**
Introduce `--access-keys` flag on `context create ecs` command to import an AWS access key CSV file (which isn't actually CSV but key=value pairs file ¯\\_(ツ)_/¯


chris-crone (approved):
LGTM

Issue #863, opened 2020-11-02
Add "Host" to Context listing in new CLI / SDK


bwateratmsft:
@gtardif 

Currently in the Docker Node SDK, the Contexts API returns the name, current yes/no, and context type. Can the Docker host also be added? This would allow us to use the SDK to query contexts and still initialize Dockerode for local use.

PR #880, opened 2020-11-04
Add context details in API context ls

gtardif:

**What I did**
* Add endpoint info to API context list result, similar but not exactly the same as json output of `docker context ls --format json`. (replaced dictionary of endpoints to gRPC `oneof` (can be aci, docker, ecs)
* Tested js client in node-sdk

resulting js client code will look like: 
```
        context.hasDockerEndpoint()? 'host:' + context.getDockerEndpoint()?.getHost():'',
        context.hasEcsEndpoint()? context.getEcsEndpoint()?.getProfile():'',
        context.hasAciEndpoint()? context.getAciEndpoint()?.getResourceGroup():'',
        context.hasAciEndpoint()? context.getAciEndpoint()?.getRegion():''
```

**Related issue**
https://github.com/docker/compose-cli/issues/863


gtardif:
Actually, I'll extract the method converting the []context, and add some unit tests

protos/contexts/v1/contexts.proto
```diff
@@ -30,6 +30,27 @@ message Context {
     string name = 1;
 	string contextType = 2;
 	bool current = 3;
+	string description = 4;
+	oneof Endpoint {
+		DockerEndpoint docker_endpoint = 5;
```
chris-crone, line 35: Is this an existing Docker context? We need to make sure that we don't confuse users. What about the local/example contexts? (Not needed now but curious what we would call them)

server/proxy/contexts.go
```diff
@@ -47,15 +47,65 @@ func (cp *contextsProxy) List(ctx context.Context, request *contextsv1.ListReque
...
+}
+
+func convertContexts(contexts []*store.DockerContext, currentContext string) *contextsv1.ListResponse {
 	result := &contextsv1.ListResponse{}
 
 	for _, c := range contexts {
-		result.Contexts = append(result.Contexts, &contextsv1.Context{
+		endpointName := c.Type()
+		if c.Type() == store.DefaultContextType {
+			endpointName = "docker"
```
chris-crone, line 59: Why not default?

server/proxy/contexts.go
```diff
@@ -47,15 +47,65 @@ func (cp *contextsProxy) List(ctx context.Context, request *contextsv1.ListReque
...
+			Description: c.Metadata.Description,
+			Current:     c.Name == currentContext,
+		}
+		switch c.Type() {
+		case store.DefaultContextType:
+			context.Endpoint = getDockerEndpoint(endpoint)
+		case store.AciContextType:
+			context.Endpoint = getAciEndpoint(endpoint)
+		case store.EcsContextType:
+			context.Endpoint = getEcsEndpoint(endpoint)
```
chris-crone, line 75: We should have a default here so that we handle all cases. Specifically as we have the local and example types as well.

server/proxy/contexts.go
gtardif: to accommodate existing structure (mapped to `docker context ls` json output BTW): endpoints are in a dictionary, the key is `docker` for the host, and context type for other contexts.

protos/contexts/v1/contexts.proto
gtardif: Currently local contexts are not available at all in released CLI, this is the existing context. I could rename this MobyEndpoint if you prefer? this name is reflected in the client API of course, see example in the description

server/proxy/contexts.go
gtardif: I'll add a default with no endpoint info for now. (and we'll add specific endpoint as we add backends). 
Local backend is not yet available in released CLI, I think we can handle this in a separate PR

gtardif: done

protos/contexts/v1/contexts.proto
chris-crone: Yeah, I suppose that Docker is fine for now. We can worry about naming the local context later :)

chris-crone (approved):
LGTM

Issue #665, opened 2020-08-28
Large compose manifests hit Cloudformation 51200 byte limit


chrisfarms:
## Summary

Larger Compose manifests fail to `docker ecs compose up` as they hit the hard coded Cloudformation template size of `51200` bytes.

## Steps to reproduce

There will be various ways to hit this issue:

* Create a Compose manifest with large number of container definitions
* Create a Compose manifest with a huge number of environment variables
* Create a Compose manifest with a complicated set of networks
* etc

then perform a `docker ecs compose up`

#### Expected outcome

A happy provisioned set of ECS things

#### Observed outcome

```
'templateBody' failed to satisfy constraint: Member must have length less than or equal to 51200
```

## Potential improvements

Upload the cloudformation template to S3 in the background and reference that instead (S3-based templates do not suffer from the 51k limit)

flaviostutz:
> Thanks for the report @chrisfarms!
> 
> Could you please share the Compose file (with redacted parts if necessary) with us so that we can get an idea of what triggers this case. We're particularly interested in number of services, containers, env vars, etc. so that we can understand how common hitting this limit will be.
> 
> We will also investigate using S3 or splitting up the Cloud Formation stack so that we don't hit the limit.

I have a real example that is hitting the limits (I manually masked sensitive information keeping the same string sizes):

```
version: '3.0'

x-aws-loadbalancer: arn:aws:elasticloadbalancing:us-east-1:000000000:loadbalancer/app/whoami/000000
x-aws-vpc: vpc-000000000
x-aws-logs_retention: 14

services:

  professores-frontend:
    image: registry.gitlab.com/exampling/build/professores-frontend:0.20.0
    x-aws-pull_credentials: arn:aws:secretsmanager:us-east-1:00000000:secret:gitlab-registry-exampling
    ports:
      - target: 80
        protocol: tcp
        x-aws-protocol: http
    deploy:
      replicas: 1
      resources:
        limits:
          cpus: '0.5'
          memory: 256MB
        reservations:
          cpus: '0.05'
          memory: 64M
    environment:
      - DEBUG=mqtt
      - ENDPOINT_BACKEND=https://admin-backend.exampling.com
      - ESTUDANTES_ENDPOINT_BACKEND=https://alunos-backend.exampling.com
      - FAVICON_URL=https://res.cloudinary.com/e-exampling/image/upload/1111111111111111/logo-exampling.png
      - CRISP_WEBSITE_ID=00000-00000-00000-0000-000
[...]

PR #882, opened 2020-11-05
Marshall cloudformation template as yaml

ndeloof:
**What I did**
Allow CloudFormation template to be marshalled as yaml or json. yaml is made the default output format (minor breaking change). Introduced `--format` flag (already used in few other places) to select format. 

This makes template output more concise (8977 vs 13608 bytes for my basic test) when used as AWS API payload.

**Related issue**
will reduce the impact of https://github.com/docker/compose-cli/issues/665

/test-ecs

image


ndeloof:
Essential=false hack doesn't work with yaml, need to investigate further

gtardif (approved):

Issue #830, opened 2020-10-26
Make it clear to users that volumes are not destroyed on docker compose down


chris-crone:
**Description**
When the user runs `docker compose down`, we do not remove volumes that were created when they deployed their workload. This is because we assume that if they used a volume that they want to reuse it with subsequent deploys.

As volumes cost money, we should alert users that their volume continues to exist after `compose down`. We can also consider adding a flag to remove volumes on `compose down`. Note: If we add a flag it should be something like `--remove-volumes` and not have a short option like `docker-compose` does.

Issue #841, opened 2020-10-28
ACI: Allow volume creation from `compose up` deployment


gtardif:
Current ACI integration requires volumes to be created *before* (through `docker volume create`) they can be used in compose files. This is not the case with `docker-compose up` nor with the ECS integration. 

Compose files using ACI volumes include all the necessary information to create the volume while deploying to ACI with  `docker compose up`, if the Azure File Share does not exist yet. (of course if it exists the ACI deployment will simply connect to the existing volume)
the `compose down` command should display a warning to make sure users are aware there is some remaining resources not deleted when the compose file uses ACI volumes.

karolz-ms:
@bigmorty say you love this! ☺️

BigMorty:
I LOVE this! :-) @gtardif, can you explain in detail when/if the volume is deleted on compose down?

Issue #442, opened 2020-08-10
Provide a way in progress API to return warnings from backend


gtardif:
Typical usage : ACR auto login currently calls `fmt.Println` to display warning when we can't auto-login to ACR : 
```
 $ ./bin/docker --context acitest run -d xxxdockerregistrygta.azurecr.io/hello-aci
Could not automatically login to xxxdockerregistrygta.azurecr.io from your Azure login. Assuming you already logged in to the ACR registry
[+] Running 0/1
 ⠹ Group fervent-beaver  Waiting         2.1s
containerinstance.ContainerGroupsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InaccessibleImage" Message="The image 'xxxdockerregistrygta.azurecr.io/hello-aci' in container group 'fervent-beaver' is not accessible. Please check the image and registry credential."
```
The progress fwk should have some API to return this warning to the client, and the cli should display it the same way

PR #886, opened 2020-11-06
Add auto creation of fileshares

ulyssessouza:
**What I did**
Add auto creation of fileshares

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/830 https://github.com/docker/compose-cli/issues/841

**Tests**
/test-aci

### DIFF
--- a/aci/aci.go
+++ b/aci/aci.go
@@ -28,6 +28,7 @@ import (
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/to"
 	tm "github.com/buger/goterm"
+	"github.com/compose-spec/compose-go/types"
 	"github.com/gobwas/ws"
 	"github.com/gobwas/ws/wsutil"
 	"github.com/morikuni/aec"
@@ -35,6 +36,7 @@ import (
 
 	"github.com/docker/compose-cli/aci/convert"
 	"github.com/docker/compose-cli/aci/login"
+	"github.com/docker/compose-cli/api/client"
 	"github.com/docker/compose-cli/api/containers"
 	"github.com/docker/compose-cli/context/store"
 	"github.com/docker/compose-cli/errdefs"
@@ -64,12 +66,46 @@ func createACIContainers(ctx context.Context, aciContext store.AciContext, group
 	return createOrUpdateACIContainers(ctx, aciContext, groupDefinition)
 }
 
+func autocreateFileshares(ctx context.Context, project *types.Project) error {
+	clt, err := client.New(ctx)
+	if err != nil {
+		return err
+	}
+	for _, v := range project.Volumes {
+		if v.Driver != convert.AzureFileDriverName {
+			return fmt.Errorf("cannot use ACI volume, required driver is %q, found %q", convert.AzureFileDriverName, v.Driver)
+		}
+		shareName, ok := v.DriverOpts[convert.VolumeDriveroptsShareNameKey]
+		if !ok {
+			return fmt.Errorf("cannot retrieve fileshare name for Azure file share")
+		}
+		accountName, ok := v.DriverOpts[convert.VolumeDriveroptsAccountNameKey]
+		if !ok {
+			return fmt.Errorf("cannot retrieve account name for Azure file share")
+		}
+		_, err = clt.VolumeService().Inspect(ctx, fmt.Sprintf("%s/%s", accountName, shareName))
+		if err != nil { // Not found, autocreate fileshare
+			if !errdefs.IsNotFoundError(err) {
+				return err
+			}
+			aciVolumeOpts := &VolumeCreateOptions{
+				Account: accountName,
+			}
+			if _, err = clt.VolumeService().Create(ctx, shareName, aciVolumeOpts); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
 func createOrUpdateACIContainers(ctx context.Context, aciContext store.AciContext, groupDefinition containerinstance.ContainerGroup) error {
 	w := progress.ContextWriter(ctx)
 	containerGroupsClient, err := login.NewContainerGroupsClient(aciContext.SubscriptionID)
 	if err != nil {
 		return errors.Wrapf(err, "cannot get container group client")
 	}
+
 	groupDisplay := "Group " + *groupDefinition.Name
 	w.Event(progress.Event{
 		ID:         groupDisplay,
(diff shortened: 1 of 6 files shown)


devdreamsolution (approved):

aci/aci.go
```diff
@@ -64,12 +66,47 @@ func createACIContainers(ctx context.Context, aciContext store.AciContext, group
...
+	if len(project.Volumes) == 0 {
+		return nil
+	}
+	for _, v := range project.Volumes {
+		if v.Driver != convert.AzureFileDriverName {
+			return fmt.Errorf("cannot use ACI volume, required driver is %q, found %q", convert.AzureFileDriverName, v.Driver)
+		}
+		shareName, ok := v.DriverOpts[convert.VolumeDriveroptsShareNameKey]
+		if !ok {
+			return fmt.Errorf("cannot retrieve fileshare name for Azurefile")
```
chris-crone, line 79: Microsoft appears to use `file share`: https://docs.microsoft.com/en-us/azure/storage/files/storage-how-to-create-file-share?tabs=azure-portal

I'm also not sure if Azurefile should be Azure file share?
```suggestion
			return fmt.Errorf("cannot retrieve file share name for Azure file share")
```

aci/aci.go
```diff
@@ -64,12 +66,47 @@ func createACIContainers(ctx context.Context, aciContext store.AciContext, group
...
+		}
+		accountName, ok := v.DriverOpts[convert.VolumeDriveroptsAccountNameKey]
+		if !ok {
+			return fmt.Errorf("cannot retrieve account name for Azurefile")
```
chris-crone, line 83: Not sure if Azurefile is something in the SDK? If not, I would suggest:
```suggestion
			return fmt.Errorf("cannot retrieve account name for Azure file share")
```

aci/aci.go
```diff
@@ -64,12 +66,47 @@ func createACIContainers(ctx context.Context, aciContext store.AciContext, group
...
+		}
+		clt, err := client.New(ctx)
+		if err != nil {
+			return err
+		}
```
chris-crone, line 88: You could do this once outside the loop

aci/aci.go
```diff
@@ -64,12 +66,47 @@ func createACIContainers(ctx context.Context, aciContext store.AciContext, group
...
+		_, err = clt.VolumeService().Inspect(ctx, fmt.Sprintf("%s/%s", accountName, shareName))
+		if err != nil { // Not found, autocreate fileshare
```
chris-crone, line 90: Can we be more specific on the error? Like only on a not found error?

aci/compose.go
```diff
@@ -46,18 +46,51 @@ func newComposeService(ctx store.AciContext) aciComposeService {
 
 func (cs *aciComposeService) Up(ctx context.Context, project *types.Project, detach bool) error {
 	logrus.Debugf("Up on project with name %q", project.Name)
+
+	err := autocreateFileshares(ctx, project)
+	if err != nil {
```
chris-crone, line 51:
```suggestion
	if err := autocreateFileshares(ctx, project); err != nil {
```

aci/compose.go
```diff
@@ -46,18 +46,51 @@ func newComposeService(ctx store.AciContext) aciComposeService {
...
-	return createOrUpdateACIContainers(ctx, cs.ctx, groupDefinition)
+	cg, err := cgClient.Get(ctx, cs.ctx.ResourceGroup, projectName)
+	if err != nil {
+		return err
+	}
+	if cg.Volumes == nil {
+		return nil
+	}
+	for _, v := range *cg.Volumes {
+		if v.AzureFile == nil || v.AzureFile.StorageAccountName == nil || v.AzureFile.ShareName == nil {
```
chris-crone, line 77: What situation does this happen in? Would it be worth adding a log here?

aci/compose.go
```diff
@@ -46,18 +46,51 @@ func newComposeService(ctx store.AciContext) aciComposeService {
...
+			continue
+		}
+		fmt.Printf("WARNING: fileshare \"%s/%s\" will NOT be deleted. Use 'docker volume rm' if you want to delete this volume\n",
```
chris-crone, line 80: We ideally shouldn't be printing from the backend.

aci/compose.go
```diff
@@ -46,18 +46,51 @@ func newComposeService(ctx store.AciContext) aciComposeService {
...
+			*v.AzureFile.StorageAccountName, *v.AzureFile.ShareName)
+	}
+	return nil
 }
 
 func (cs *aciComposeService) Down(ctx context.Context, project string) error {
 	logrus.Debugf("Down on project with name %q", project)
 
+	err := cs.warnKeepVolumeOnDown(ctx, project)
+	if err != nil {
```
chris-crone, line 90:
```suggestion
	if err := cs.warnKeepVolumeOnDown(ctx, project); err != nil {
```

aci/aci.go
gtardif: I don't think so, you can have multiple volumes, you need to check for each if you need to create it or not

aci/compose.go
gtardif: Typically secrets are also part of the ACI volume list, but have different fields.

aci/compose.go
gtardif: Yes, ideally I agree... need quite some refactoring to allow this, I don't really want to include this in this PR. We have https://github.com/docker/compose-cli/issues/442 to track this

aci/aci.go
chris-crone: I meant creating the client. Since it's created from the context without any volume information, we can move it outside the loop.

aci/aci.go
```diff
@@ -64,12 +66,47 @@ func createACIContainers(ctx context.Context, aciContext store.AciContext, group
...
+			if !errdefs.IsNotFoundError(err) {
+				return err
+			}
+			aciVolumeOpts := &VolumeCreateOptions{
+				Account: accountName,
+			}
+			_, err = clt.VolumeService().Create(ctx, shareName, aciVolumeOpts)
+			if err != nil {
+				return err
+			}
```
chris-crone, line 97: nit:
```suggestion
			if _, err = clt.VolumeService().Create(ctx, shareName, aciVolumeOpts); err != nil {
				return err
			}
```

aci/aci.go
gtardif: done

gtardif (approved):

Issue #894, opened 2020-11-09
Debugging CLI error with valid compose file 


steviewanders:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Description**

When calling `docker compose up` in a valid ECS context run into `panic: runtime error: invalid memory address or nil pointer dereference` (full output below)

**Steps to reproduce the issue:**
1. Install all the Docker requirements and local environment like AWS config and ECS context
2. Run `docker compose up` in a directory with a `docker-compose.yaml` file that is valid to `docker-compose`

**Describe the results you received:**

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x21aaf82]

goroutine 39 [running]:
github.com/docker/compose-cli/ecs.(*ecsAPIService).createTaskRole(0xc00004e900, 0xc0006d0cc0, 0xc0000951fa, 0x2, 0x0, 0x0, 0x0, 0x3a40b00, 0x0, 0x0, ...)
        github.com/docker/compose-cli/ecs/cloudformation.go:395 +0x3a2
github.com/docker/compose-cli/ecs.(*ecsAPIService).createService(0xc00004e900, 0xc0006d0cc0, 0xc0000951fa, 0x2, 0x0, 0x0, 0x0, 0x3a40b00, 0x0, 0x0, ...)
        github.com/docker/compose-cli/ecs/cloudformation.go:104 +0x118
github.com/docker/compose-cli/ecs.(*ecsAPIService).convert(0xc00004e900, 0x2d76760, 0xc0006b65a0, 0xc0006d0cc0, 0xc00000e498, 0x2d94ec0, 0xc00000e4b0)
        github.com/docker/compose-cli/ecs/cloudformation.go:83 +0x638
github.com/docker/compose-cli/ecs.(*ecsAPIService).Convert(0xc00004e900, 0x2d76760, 0xc0006b65a0, 0xc0006d0cc0, 0xc, 0x0, 0x0, 0x1, 0x1)
        github.com/docker/compose-cli/ecs/cloudformation.go:41 +0x4d
github.com/docker/compose-cli/ecs.(*ecsAPIService).Up(0xc00004e900, 0x2d76760, 0xc0006b65a0, 0xc0006d0cc0, 0x0, 0x0, 0x0)
        github.com/docker/compose-cli/ecs/up.go:35 +0xc5
[...]

Issue #891, opened 2020-11-09
Not able to create two EFS volumes in the same compose file 


mreferre:
I am using the latest 1.0.2 build of the compose CLI.

**Description**

I am trying to deploy a container on Amazon ECS with two distinct directories as per the following sample:
```
version: "3.8"
services:
  ecsservice:
    image: <myimage> 
    volumes:
        - source:/sourcefolder
        - destination:/destinationfolder
volumes:
  source: 
    driver_opts:
      backup_policy: ENABLED
  destination:
    driver_opts:
      backup_policy: ENABLED
```

This errors out with complaining a duplicate in the IAM policy name. Converting the compose to CFN IAM in fact you notice that the tool creates two `PolicyDocument` with the same name (`PolicyName: TmpEcsserviceVolumeMountPolicy`)

**Steps to reproduce the issue:**
See above

**Describe the results you received:**
`Multiple IAM policies have the same policy name: [<......>VolumeMountPolicy]. Specify a unique name for each policy and run the stack operation again.`

**Describe the results you expected:**
Bing able to create two separate volumes. I could alternatively use a single EFS file system if I was able to create two separate EFS Access Points. Is it possible to specify the creation of separate APs so that I could map them in the container? 

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
docker version
Client: Docker Engine - Community
 Cloud integration: 1.0.2
 Version:           20.10.0-beta1
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        ac365d7
 Built:             Tue Oct 13 18:13:53 2020
 OS/Arch:           darwin/amd64
 Context:           myecscontext
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.0-beta1
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       9c15e82
  Built:            Tue Oct 13 18:17:18 2020
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          v1.4.1
  GitCommit:        c623d1b36f09f8ef6536a057bd658b3aa8632828
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

```

**Additional environment details (AWS ECS, Azure ACI, local, etc.):**
Amazon ECS

[...]

PR #897, opened 2020-11-10
Don't panic on bind mount

ndeloof:
**What I did**
Detect use of bind mounts which are not supported on Fargate

**Related issue**
https://github.com/docker/compose-cli/issues/894
https://github.com/docker/compose-cli/issues/891

/test-ecs


chris-crone (approved):
LGTM

gtardif (approved):

Issue #865, opened 2020-11-03
Azure Devops Integration fails on volume mount 


abinashkd:
**Description**

I am trying to integrate a docker compose file in Azure DevOps which will deploy to ACI. In the pipeline I am having problem with volume mounts. The images which doesn't require volume mount gets deployed successfully, but images with volume mount fails with could not find volume source "/proc/"

**Steps to reproduce the issue:**
1. Follow the documentation process of ACI for docker compose. 
2. In Azure DevOps create a release pipeline with below commands in CLI:

#To install docker compose cli
curl -L https://raw.githubusercontent.com/docker/compose-cli/main/scripts/install/install_linux.sh | sh
sudo docker login azure --client-id ** --client-secret ** --tenant-id **

#These to check if paths exists or not:
ls -ltr /var/run/docker.sock
ls -ltr /proc/
ls -ltr /sys/fs/cgroup

which docker 
docker version

sudo docker context create aci myacicontext --resource-group ** --location **

sudo docker context use myacicontext

sudo docker compose up -d

3.

**Describe the results you received:**

Downloading CLI...
Downloaded CLI!
Installing CLI...
Done!

login succeeded

total 8
drwxr-xr-x 3 vsts docker 4096 Nov  3 04:56 _projects-azure-CI
-rw-r--r-- 1 vsts docker  648 Nov  3 04:56 docker-compose.yaml

srw-rw---- 1 root docker 0 Nov  3 04:30 /var/run/docker.sock
total 0
lrwxrwxrwx  1 root            root                          0 Nov  3 04:30 thread-self -> 2575/task/2575
lrwxrwxrwx  1 root            root                          0 Nov  3 04:30 self -> 2575
....
total 0
dr-xr-xr-x 5 root root  0 Nov  3 04:30 systemd
dr-xr-xr-x 2 root root  0 Nov  3 04:30 rdma
...

/usr/local/bin/docker
Client:
 Cloud integration: 1.0.1
 Version:           19.03.13+azure
 API version:       1.40
 Go version:        go1.13.15
 Git commit:        cd8016b6bcc51a51f577bcffb37e6a870f7813f9
 Built:             Tue Sep 15 23:02:04 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          19.03.13+azure
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       bd33bbf0497b2327516dc799a5e541b720822a4c
[...]

PR #899, opened 2020-11-10
Specific error message when specifying path as a volume source

gtardif:

**What I did**
add specific error message when user tries to specify a host path in compose file

**Related issue**
https://github.com/docker/compose-cli/issues/865

/test-aci

### DIFF
--- a/aci/convert/volume.go
+++ b/aci/convert/volume.go
@@ -38,18 +38,17 @@ const (
 	volumeReadOnly                 = "read_only"
 )
 
-func (p projectAciHelper) getAciFileVolumes(ctx context.Context, helper login.StorageLogin) (map[string]bool, []containerinstance.Volume, error) {
-	azureFileVolumesMap := make(map[string]bool, len(p.Volumes))
+func (p projectAciHelper) getAciFileVolumes(ctx context.Context, helper login.StorageLogin) ([]containerinstance.Volume, error) {
 	var azureFileVolumesSlice []containerinstance.Volume
 	for name, v := range p.Volumes {
 		if v.Driver == azureFileDriverName {
 			shareName, ok := v.DriverOpts[volumeDriveroptsShareNameKey]
 			if !ok {
-				return nil, nil, fmt.Errorf("cannot retrieve fileshare name for Azurefile")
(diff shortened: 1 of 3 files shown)


aci/convert/volume.go
```diff
@@ -82,8 +82,11 @@ func (p projectAciHelper) getAciFileVolumes(ctx context.Context, helper login.St
 func (s serviceConfigAciHelper) getAciFileVolumeMounts(volumesCache map[string]bool) ([]containerinstance.VolumeMount, error) {
 	var aciServiceVolumes []containerinstance.VolumeMount
 	for _, sv := range s.Volumes {
+		if strings.Contains(sv.Source, "/") {
```
ndeloof, line 85: better rely on sv.Type as https://github.com/compose-spec/compose-go/blob/5afecaa4cb35039bebfc86ae0c3742dab7cae0f3/loader/volume.go#L121 do already handle filepath detection for source.

aci/convert/volume.go
```diff
@@ -82,8 +82,11 @@ func (p projectAciHelper) getAciFileVolumes(ctx context.Context, helper login.St
...
+			return []containerinstance.VolumeMount{}, fmt.Errorf("host path (%q) not allowed as volume source, you need to reference an Azure File Share defined in the 'volumes' section", sv.Source)
+		}
 		if !volumesCache[sv.Source] {
-			return []containerinstance.VolumeMount{}, fmt.Errorf("could not find volume source %q", sv.Source)
+			return []containerinstance.VolumeMount{}, fmt.Errorf("could not find volume source %q, service volumes must reference volumes defined in the 'volumes' section", sv.Source)
```
ndeloof, line 89: already checked by https://github.com/compose-spec/compose-go/blob/5afecaa4cb/loader/validate.go#L38

ndeloof (changes requested):
could better leverage compose-go to parse and validate volume specs

aci/convert/volume.go
gtardif: Thx! updated

aci/convert/volume.go
gtardif: fixed

ndeloof (approved):

Issue #839, opened 2020-10-28
Add tempfs support (feature request)


mattyweb:
**Description**

Was trying to deploy our app to ECS and it appeared that tmpfs isn't supported:

```
WARN[0000] services.tmpfs: unsupported attribute       
```

My service does data ingestion and has a few tasks that are reading and writing from a tmpfs directory. The final results are committed to a database so I'm happy to have the volume be ephemeral and disappear when the container does.

chris-crone:
Thanks for the feedback @mattyweb! It looks like `tmpfs` is [not supported by Fargate](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_linuxparameters):
> Note: If you are using tasks that use the Fargate launch type, the tmpfs parameter is not supported.

Can you just store your files in `/tmp` in the container as a workaround? IIRC containers running on Fargate have around 20 GB of free storage?

mattyweb:
Ah, OK, didn't know that. We use EC2 instances. I gather you're going for lowest common denominator features so it runs the same across EC2, Fargate and Azure?

BTW, not sure if this is the plan but it would be helpful if there were a Docker Compose CLI version of the reference doc here: https://docs.docker.com/compose/compose-file/

In case it's helpful, for starters I just tried running a current docker-compose.yml and got the following:

```
WARN[0000] services.restart: unsupported attribute      
WARN[0000] services.build: unsupported attribute        
ERRO[0000] service api doesn't define a Docker image to run: incompatible attribute 
ERRO[0000] published port can't be set to a distinct value than container port: incompatible attribute 
WARN[0000] services.restart: unsupported attribute      
WARN[0000] services.build: unsupported attribute        
ERRO[0000] service prefect doesn't define a Docker image to run: incompatible attribute 
WARN[0000] services.tmpfs: unsupported attribute        
WARN[0000] services.restart: unsupported attribute      
WARN[0000] non-external volumes are not supported: unsupported attribute 
```

I was able to fix everything over time, but took a bit of trial and error.

chris-crone:
> I gather you're going for lowest common denominator features so it runs the same across EC2, Fargate and Azure?

[...]

PR #903, opened 2020-11-12
Document ECS supported compose attributes

ndeloof:
**What I did**
Document ECS support for compose attributes as suggested by https://github.com/docker/compose-cli/issues/839

### DIFF
--- a/docs/ecs-compose-features.md
+++ b/docs/ecs-compose-features.md
@@ -1,0 +1,176 @@
+# Compose - Amazon ECS mapping
+
+This document outlines the conversion of an application defined in a Compose file to AWS resources.
+Each service is mapped to an ECS service in the project's cluster.
+
+## Compose fields mapping
+
+The table below lists supported Compose file fields and their AWS counterparts.
+
+__Legend:__
+
+- __✓:__ Implemented
+- __n:__ Not yet implemented
+- __x:__ Not applicable / no available conversion
+
+| Keys                           |Map|  Notes                                                       |
+|--------------------------------|---|--------------------------------------------------------------|
+| __Service__                    | ✓ |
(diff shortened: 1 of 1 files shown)


docs/ecs-compose-features.md
```diff
@@ -0,0 +1,165 @@
...
+| service.tmpfs                  | x |  Not supported on Fargate, see https://github.com/docker/compose-cli/issues/839
+| service.entrypoint             | ✓ | 
+| service.env_file               | ✓ |
+| service.environment            | ✓ |
+| service.expose                 | x |
+| service.extends                | ✓ |
+| service.external_links         | x |
+| service.extra_hosts            | x |
+| service.group_add              | x |
+| service.healthcheck            | ✓ |
```
gtardif, line 49: I thought we could not support healthchecks on ECS? (cf https://docker.slack.com/archives/C0111G2RCE4/p1600870550019100). Did I misundestand this ?

docs/ecs-compose-features.md
```diff
@@ -0,0 +1,165 @@
...
+
+secrets:
+  mysecret:
+    file: ./my_secret1.txt
+```
+
+
+## Container Resources
+
+CPU and memory reservations and limits can be set in compose.
```
gtardif, line 165: Even if it's standard compose file format, having an example here would be useful (or a link users can click to see compose file reference on CPU/memory limits and some example)

gtardif (approved):
1 minor point, 1 question, otherwise LGTM

docs/ecs-compose-features.md
ndeloof: we support container healthcheck (same a docker) but not Load Balancer healthcheck
and please don't ask me why ALB don't just rely on container healthchecks that are used to report service health on ECS console :P

PR #905, opened 2020-11-12
Use ACI .IsHttpStatus() rather than accessing .StatusCode directly

gtardif:
that might lead to nil pointer panic

**What I did**
is in the title

**Related issue**

/test-aci


gtardif:
@ndeloof just following what we discussed earlier on about nice golang features side-effects...

ndeloof (approved):

PR #906, opened 2020-11-13
Collect metrics --help flag

gtardif:

**What I did**
Add --help in specific flags we catch in metrics

**Related issue**
https://docker.slack.com/archives/C0111G2RCE4/p1605266835109100

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -17,6 +17,8 @@
 package metrics
 
 import (
+	"strings"
+
 	"github.com/docker/compose-cli/utils"
 )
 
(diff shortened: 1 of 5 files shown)


metrics/metrics.go
```diff
@@ -52,18 +54,15 @@ func GetCommand(args []string) string {
 	onlyFlags := false
 	for _, arg := range args {
 		if arg == "--help" {
```
chris-crone, line 56: nit: You could make this a switch so that it's easier to read

metrics/metrics.go
```diff
@@ -52,18 +54,15 @@ func GetCommand(args []string) string {
...
 			break
 		}
 		if isCommandFlag(arg) || (!onlyFlags && isCommand(arg)) {
-			if result == "" {
-				result = arg
-			} else {
-				result = result + " " + arg
-			}
-			if !isManagementCommand(arg) {
+			result = strings.TrimSpace(result + " " + arg)
```
chris-crone, line 64: I don't know this code well, why did you remove `if result == ""`?

gtardif: Trim spaces rather than have an if to add space only between words, saves 4 lines of code

chris-crone: I missed the addition of the TrimSpace :)

chris-crone (approved):
LGTM

metrics/metrics.go
gtardif: This is happening in the background, to record metrics, and is followed by an http call to local Desktop, so I don't think we care. (and the args we manipulate are already trimmed)

PR #913, opened 2020-11-16
Add support for gpu devices reservation

ndeloof:
**What I did**
In addition to swarm-style for generic_resources, support devices reservation as defined by Adopt https://github.com/compose-spec/compose-spec/pull/100

/test-ecs

### DIFF
--- a/ecs/convert.go
+++ b/ecs/convert.go
@@ -562,6 +562,11 @@ func gpuRequirements(s types.ServiceConfig) int64 {
 					return resource.DiscreteResourceSpec.Value
 				}
 			}
+			for _, device := range reservations.Devices {
+				if len(device.Capabilities) == 1 && device.Capabilities[0] == "gpu" {
+					return int64(device.Count)
+				}
+			}
 		}
 	}
 	return 0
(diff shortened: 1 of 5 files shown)


ecs/convert.go
```diff
@@ -562,6 +562,11 @@ func gpuRequirements(s types.ServiceConfig) int64 {
 					return resource.DiscreteResourceSpec.Value
 				}
 			}
+			for _, device := range reservations.Devices {
+				if len(device.Capabilities) == 1 && device.Capabilities[0] == "gpu" {
```
aiordache, line 566: Do we want to restrict the capabilities to "gpu" only? Could we be more flexible? I thought these flags are passed to a Docker Engine and the engine can return an error if we request a capability that is not supported.

ndeloof: we don't target docker engine here but AWS EC2 and need to select machine type based on number of requested GPUs

gtardif (approved):
Unsure if `CheckDeployResourcesDevicesCapabilities` is unused and should be removed, otherwise LGTM

aiordache (approved):

ndeloof:
CheckDeployResourcesDevicesCapabilities is invoked by compose-go https://github.com/compose-spec/compose-go/blob/d948014098818897aa3fe60cc1858c65788212d3/compatibility/deploy.go#L168

PR #912, opened 2020-11-16
Compose api

gtardif:
**What I did**
* Add Compose gRPC service, and proxy for `compose up`, `compose down`, `compose ls`, `compose ps`
* tested with node-sdk compose sample

**Related issue**
https://github.com/docker/compose-cli/issues/890

**(not mandatory) A picture of a cute animal, if possible in relation with what you did**

Issue #890, opened 2020-11-09
Add compose service to CLI API


gtardif:
compose-cli provides an API to run commands remotely (from js clients, pythong clients or other, generated from gRPC). 
This API currently does not expose compose commands (up, down, ps, ls), we should expose these commands to allow deployments of stacks through the API

PR #916, opened 2020-11-17
Fix proto message names for ComposeStacks() input / output ; same for ComposeServices()

gtardif:
Follow up after https://github.com/docker/compose-cli/pull/912

**What I did**
* realise I had renamed List() and Ps() to Stacks() and Services(), to avoid name collision and make things much more clear, bu the param names stayed ComposePsRequest, etc.

**Related issue**
https://github.com/docker/compose-cli/issues/890


chris-crone (approved):
LGTM

PR #919, opened 2020-11-17
Compose in go

ndeloof:
**What I did**
Implemented compose for local backend
this is not feature-complete and is missing e2e coverage, but we'd like to avoid long-running branches

**Related issue**
https://github.com/docker/roadmap/issues/15

image


local/containers.go
```diff
@@ -144,7 +145,10 @@ func (cs *containerService) Run(ctx context.Context, r containers.ContainerConfi
 }
 
 func (cs *containerService) create(ctx context.Context, containerConfig *container.Config, hostConfig *container.HostConfig, networkingConfig *network.NetworkingConfig, name string) (string, error) {
-	created, err := cs.apiClient.ContainerCreate(ctx, containerConfig, hostConfig, networkingConfig, name)
+	created, err := cs.apiClient.ContainerCreate(ctx, containerConfig, hostConfig, networkingConfig, &v1.Platform{
```
ndeloof, line 148: This cause error `1.41 is too new. Maximum supported API version is 1.40` on CI and on my Linux box :'(

local/compose.go
```diff
@@ -0,0 +1,565 @@
...
+			project.Volumes[k] = volume
+		}
+		err := s.ensureVolume(ctx, volume)
+		if err != nil {
+			return err
+		}
+	}
+
+	for _, service := range project.Services {
+		err := s.applyPullPolicy(ctx, service)
```
rumpl, line 72: Could we do this in parallel?

ndeloof: Sure we can, I wonder there's any reason parallel_pull is set to `False` on docker/compose (https://github.com/docker/compose/blob/854c003359bd07d0d3ca137d7a08509cfeab0436/compose/project.py#L724)
Engine will download layers in parallel anyway
This whole thing will need to be revisited anyway with addition of build support (by importing some code from `bake`)

local/compose.go
```diff
@@ -0,0 +1,597 @@
...
+			return nil
+		})
+	}
+	return eg.Wait()
+}
+
+func (s *local) Logs(ctx context.Context, projectName string, w io.Writer) error {
+	list, err := s.containerService.apiClient.ContainerList(ctx, moby.ContainerListOptions{
+		Filters: filters.NewArgs(
+			filters.Arg("label", "com.docker.compose.project="+projectName),
```
gtardif, line 215: We need const for all these compose labels...

local/compose.go
```diff
@@ -0,0 +1,597 @@
...
+		),
+	})
+	if err != nil {
+		return err
+	}
+	var wg sync.WaitGroup
+	consumer := formatter.NewLogConsumer(w)
+	for _, c := range list {
+		service := c.Labels["com.docker.compose.service"]
```
gtardif, line 224: const +1... and all others

gtardif (approved):
consts for compose labels could be a nice quick fix, I fixed a bug in logs... We could continue improving a lot of things but I'd rather merge this quickly and do follow up PRs, seems already plenty of features. (Will certainly need tests ta validate the many possible scenarios)

rumpl (approved):

PR #924, opened 2020-11-18
Run convergence in service dependency order

ndeloof:
**What I did**
sort services in respect to dependency order so as we recreate one we can flag it's dependent services as `force_recreate` before convergence is applied on those.

Also fix a panic with attempt to write to `results` channel while already closed. Replaced by a dedicated errors channel. 

Bonus: implement `service_healthy` depends_on condition for parity with docker-compose

can be tested using:
```yaml
services:
  jenkins:
    image: jenkins/jenkins:lts
    healthcheck: 
      test: "curl http://localhost:8080/login"
  redis:
    image: redis
    depends_on:
      jenkins:
        condition: service_healthy
```

`compose up` will demonstrate jenkins service being created, and redis one only created after a delay (healthy state)
then change jenkins service configuration, `compose up` will recreate jenkins service _then_ recreate redis one.

### DIFF
--- a/local/convergence.go
+++ b/local/convergence.go
@@ -40,7 +40,10 @@ const (
 )
 
 func (s *local) ensureService(ctx context.Context, project *types.Project, service types.ServiceConfig) error {
-	s.waitDependencies(ctx, project, service)
+	err := s.waitDependencies(ctx, project, service)
+	if err != nil {
+		return err
+	}
 
 	actual, err := s.containerService.apiClient.ContainerList(ctx, moby.ContainerListOptions{
 		Filters: filters.NewArgs(
@@ -117,7 +120,10 @@ func (s *local) waitDependencies(ctx context.Context, project *types.Project, se
 		switch config.Condition {
 		case "service_healthy":
 			eg.Go(func() error {
-				for range time.Tick(500 * time.Millisecond) {
+				ticker := time.NewTicker(500 * time.Millisecond)
+				defer ticker.Stop()
+				for {
+					<-ticker.C
 					healthy, err := s.isServiceHealthy(ctx, project, dep)
 					if err != nil {
 						return err
@@ -126,7 +132,6 @@ func (s *local) waitDependencies(ctx context.Context, project *types.Project, se
 						return nil
 					}
 				}
-				return nil
 			})
 		}
 	}
--- a/local/convergence.go
+++ b/local/convergence.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"time"
 
 	"github.com/compose-spec/compose-go/types"
 	moby "github.com/docker/docker/api/types"
@@ -39,6 +40,8 @@ const (
 )
 
 func (s *local) ensureService(ctx context.Context, project *types.Project, service types.ServiceConfig) error {
+	s.waitDependencies(ctx, project, service)
+
 	actual, err := s.containerService.apiClient.ContainerList(ctx, moby.ContainerListOptions{
 		Filters: filters.NewArgs(
 			filters.Arg("label", fmt.Sprintf("%s=%s", projectLabel, project.Name)),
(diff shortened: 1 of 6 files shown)


local/dependencies.go
```diff
@@ -25,37 +25,59 @@ import (
...
 			}
 			if containsAll(ready, service.GetDependencies()) {
 				service := service
 				scheduled = append(scheduled, service.Name)
 				eg.Go(func() error {
-					err := fn(service)
+					err := fn(ctx, service)
 					if err != nil {
-						close(results)
+						errors <- err
```
rumpl, line 49: Not sure I understand why we need the `errors` chan

ndeloof: other goroutine will try to close an already close channel, which results into panic

ndeloof: for docker/c2 I uses a chan Pair(result, error) for this same needs.
I'd welcome any canonical golang reference about running things in parallel, collect results (which triggers more things to run, this is not just a parallel "map reduce") and stop on first error

rumpl: But wait, things are not running in parallel here right? Your code is equivalent to

```go
func inDependencyOrder(ctx context.Context, project *types.Project, fn func(context.Context, types.ServiceConfig) error) error {
	var (
		scheduled []string
		ready     []string
	)
	services := sortByDependency(project.Services)

	results := []string{}
	for len(ready) < len(services) {
		eg, _ := errgroup.WithContext(ctx)
		for _, service := range services {
			if contains(scheduled, service.Name) {
				continue
			}
			if containsAll(ready, service.GetDependencies()) {
				service := service
				scheduled = append(scheduled, service.Name)
				eg.Go(func() error {
					err := fn(ctx, service)
					if err != nil {
						return err
					}
					results = append(results, service.Name)
					return nil
				})
			}
		}

		if err := eg.Wait(); err != nil {
			return err
		}
	}

	return nil
}
```

Basically, each time you run a goroutine you just block while it goes up no? What am I missing here?

rumpl: Ok no, you need the `results` chan but you can move the `eg` inside the first `for` loop I think

ndeloof: the idea is that first result _may_ allow to schedule another service to start.
example: with services A,B,C,D A depends on B, C depends on D
first iteration will create B and D in parallel
B starts first and sends results -> new iteration will trigger A to be created, while D might not (yet) be ready - especially considering support for `service_healthy`

note: this is mostly a direct transcription of docker-compose convergence logic here.

ndeloof: Would be nice this loop actually run in parallel with the pull_policy one, so that some service can start and get healthy while other still pull... maybe a bit over-engineering for something that only will take seconds ;P

rumpl: Oookay understood, it's hairy :D

ndeloof: Yes. The issue is the actual parrapel task list depends on the completed ones... This is where a dependency graph would help ;)

local/convergence.go
```diff
@@ -102,6 +111,28 @@ func (s *local) ensureService(ctx context.Context, project *types.Project, servi
 	return eg.Wait()
 }
 
+func (s *local) waitDependencies(ctx context.Context, project *types.Project, service types.ServiceConfig) error {
+	eg, ctx := errgroup.WithContext(ctx)
+	for dep, config := range service.DependsOn {
+		switch config.Condition {
+		case "service_healthy":
```
gtardif, line 118: Intuitively, I feel this waiting code could be integrated in `inDependencyOrder()`, and that would make things easier to understand (I hope). Here the complexity of the scheduling is split in 2 places and it makes it hard to understand the possible interactions between the 2. (Also note in terms of runtime, the function passed to `inDependencyOrder()` will precisely first call `waitDependencies`). 

Or possibly start services more brutally, and add a case here for dependencies, not only waiting on healthchecks, leaving each service the responsibility to wait for all it's dependencies in the proper state (requires refactoring to share status of all services)

gtardif: I would need to try things out, and discuss this more in detail...

gtardif (commented):
Open question on refactoring how services wait for their dependencies (see my comment)

In terms of UX, it would be nice to display a service "Waiting" status in the progress. With the jenkins example we see jenkins service "created, then nothing happens for some time, we don't know why. 
And possibly a "Healthy" status when jenkins gets healthy... displaying these statuses in progress might be linked to the technical refactoring of service scheduling :)

local/convergence.go
ndeloof: `inDependencyOrder` only sort services so we process services in the right order
service healthy handle the _condition_ a depdency is considered fullfiled and we can process the next step

> leaving each service the responsibility to wait for all it's dependencies in the proper state
this is what swarm decided to adopt, and cause for depends_on to be removed from compose file v3

gtardif (approved):
We can leave improving the UX (Waiting status in this case while redis is waiting for Jenkins) for a follow up...

PR #925, opened 2020-11-18
Local compose ls implementation

gtardif:

**What I did**
Add implementation for compose ls

**Related issue**


gtardif:
I'll add some tests on that.
Happy to discuss what should display in the status for the full project

ndeloof (approved):

PR #926, opened 2020-11-19
compose ps implementation

gtardif:

**What I did**
* swear about manipulating collections in go
* transform a collection of containers to a collection of service status with many for loops
* unit test

**Related issue**


gtardif:
haven't done the ports yet. 
Also, it would be better to read `desired` from the project file, we need to add labels on containers: `com.docker.compose.project.config_files` and `com.docker.compose.project.working_dir` like docker-compose, in order to reload the compose project for this. 
(these labels are also used from Desktop UI, so would be good to keep these anyway)

ndeloof (approved):

PR #945, opened 2020-11-21
Define graph.walk to run into dependencies in dependency order

ndeloof:
**What I did**
Implement "process services in dependency order" by introducing `graph.walk` and a _producer_ goroutine to schedule service processing as their dependencies have been managed

also introduce a `direction` so we can run the same in reverse order for `down` 


rumpl:
Shit, I was doing on the same thing...

rumpl:
I'll do mine anyway

PR #947, opened 2020-11-21
Use a dependency graph to start services

rumpl:
**What I did**

Use a dependency graph to start services

The algorithm is like so:
* get all the leaves of the graph, these are all the service that don't have any dependency
* once a service is started we take the list of its parents (dependents)
* if all the dependencies of each of those dependents are started then we can start it as well
* if not then we continue to the next dependent

Tested on 

```yaml
services:
  db:
    build: db
    image: gtardif/sentences-db

  words:
    build: words
    image: gtardif/sentences-api
    ports:
      - "8080:8080"
  web:
    build: web
    image: gtardif/sentences-web
    ports:
      - "80:80"

```

And 

```yaml
services:
  db:
    build: db
    image: gtardif/sentences-db

  words:
    build: words
    image: gtardif/sentences-api
    ports:
      - "8080:8080"
    depends_on:
      - db
  web:
    build: web
    image: gtardif/sentences-web
    ports:
      - "80:80"
    depends_on:
      - db
      - words

```

And 

```yaml
services:
  db:
    build: db
    image: gtardif/sentences-db

  words:
    build: words
    image: gtardif/sentences-api
    ports:
      - "8080:8080"
    depends_on:
      - db
  web:
    build: web
    image: gtardif/sentences-web
    ports:
      - "80:80"
    depends_on:
      - db
```

If a cycle is found it will return an error, for example for this compose file 

```yaml
services:
  db:
    build: db
    image: gtardif/sentences-db
    depends_on:
      - web

  words:
    build: words
    image: gtardif/sentences-api
    ports:
      - "8080:8080"
    depends_on:
      - db
  web:
    build: web
    image: gtardif/sentences-web
    ports:
      - "80:80"
    depends_on:
      - db
```

The error will be 

```
cycle found: db -> web -> db
```

The reverse (for `down`) is pretty much the same (not yet implemented):
* take all the root nodes
* instead of dependents we take its dependencies

Note: this is just a proposal, I might have missed some subtle things, happy to talk about it or just use your implementation @ndeloof

### DIFF
--- a/local/dependencies.go
+++ b/local/dependencies.go
@@ -21,6 +21,7 @@ package local
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/compose-spec/compose-go/types"
@@ -36,6 +37,10 @@ const (
 
 func inDependencyOrder(ctx context.Context, project *types.Project, fn func(context.Context, types.ServiceConfig) error) error {
 	g := NewGraph(project.Services)
+	if b, err := g.HasCycles(); b {
+		return err
+	}
+
 	leaves := g.Leaves()
 
 	eg, _ := errgroup.WithContext(ctx)
@@ -50,7 +55,7 @@ func inDependencyOrder(ctx context.Context, project *types.Project, fn func(cont
 func run(ctx context.Context, graph *Graph, eg *errgroup.Group, nodes []*Vertex, fn func(context.Context, types.ServiceConfig) error) error {
 	for _, node := range nodes {
 		n := node
-		// Don't start this service yet if all of their children have
+		// Don't start this service yet if all of its children have
 		// not been started yet.
 		if len(graph.FilterChildren(n.Service.Name, ServiceStopped)) != 0 {
 			continue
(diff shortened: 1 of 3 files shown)


rumpl:
Reversing the algorithm would be easy too, basically instead of leaves we take the roots and then once a service is started we take the dependencies instead of dependents.

local/dependencies.go
```diff
@@ -20,93 +20,225 @@ package local
```
rumpl, line 20: It's easier to look at the [final file](https://github.com/docker/compose-cli/blob/697964ca7cc0216e67ab4162bdaf4d849168b910/local/dependencies.go) rather than the diff

rumpl:
I'm also thinking about adding debug info that will just output graphviz graphs each time the graph changes

ndeloof (approved):

ndeloof:
it seems you enjoyed doing this, so let's use yours

rumpl:
> it seems you enjoyed doing this, so let's use yours

I did 🤓

PR #948, opened 2020-11-22
Use json progress message

rumpl:
It contains the same output that `docker pull` shows


ndeloof:
ProgressMessage is marked as deprecated
https://github.com/moby/moby/blob/master/pkg/jsonmessage/jsonmessage.go#L142
is it safe to use it ?

rumpl:
Indeed, fixed it

ndeloof (approved):

PR #952, opened 2020-11-23
Change resource reservation in e2e tests, fighting ACI flakyness

gtardif:

**What I did**
change resource reservation in compose resource limit ACI e2e test

**Related issue**

/test-windows
/test-aci

### DIFF
--- a/aci/aci.go
+++ b/aci/aci.go
@@ -18,9 +18,11 @@ package aci
 
 import (
 	"context"
+	"encoding/json"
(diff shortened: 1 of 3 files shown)


gtardif:
Adding debug info, we can see docker credentials are sent to ACI on windows nodes, GHAction automatically `docker login --user githubactions` before running tests. ACI Has some flakyness currently on payloads freezing when credentials are provided, and tests simply timeout.

PR #956, opened 2020-11-23
Avoid port collision with compose up test

ulyssessouza:
**What I did**
Avoid port collision with compose up test (`TestLocalBackendComposeUp`), since the 2 tests run in parallel.

### DIFF
--- a/local/e2e/backend_test.go
+++ b/local/e2e/backend_test.go
@@ -78,7 +78,7 @@ func TestLocalBackendRun(t *testing.T) {
 	})
 
 	t.Run("run with ports", func(t *testing.T) {
-		res := c.RunDockerCmd("run", "-d", "-p", "8080:80", "nginx")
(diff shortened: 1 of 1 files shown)


local/e2e/backend_test.go
```diff
@@ -78,15 +78,15 @@ func TestLocalBackendRun(t *testing.T) {
 	})
 
 	t.Run("run with ports", func(t *testing.T) {
-		res := c.RunDockerCmd("run", "-d", "-p", "8080:80", "nginx")
+		res := c.RunDockerCmd("run", "-d", "-p", "8888:80", "nginx")
```
ndeloof, line 81: could we maybe not assign a fixed exposed port and just get the actual random one by inspect response just belond ?

ndeloof (approved):

ndeloof:
unrelated: why are e2e local tests under `local` and not `tests/local-e2e` like we did for other backends ?

PR #961, opened 2020-11-25
Add container labels to improve compatibility with docker-compose

gtardif:

**What I did**
* Add container labels version, oneoff, workingDir, configfiles

**Related issue**

### DIFF
--- a/local/compose.go
+++ b/local/compose.go
@@ -363,10 +363,15 @@ func getContainerCreateOptions(p *types.Project, s types.ServiceConfig, number i
 	if err != nil {
 		return nil, nil, nil, err
 	}
+	//TODO: change oneoffLabel value for containers started with `docker compose run`
 	labels := map[string]string{
 		projectLabel:         p.Name,
 		serviceLabel:         s.Name,
+		versionLabel:         ComposeVersion,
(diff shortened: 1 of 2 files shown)


gtardif:
@ulyssessouza this will probably help you to be able to find the service order in the `down` command, without specifying the compose file in the command line

local/compose.go
```diff
@@ -366,7 +366,11 @@ func getContainerCreateOptions(p *types.Project, s types.ServiceConfig, number i
 	labels := map[string]string{
 		projectLabel:         p.Name,
 		serviceLabel:         s.Name,
+		versionLabel:         ComposeVersion,
+		oneoffLabel:          "False",
```
rumpl, line 370: Should we add a "TODO" here? I feel like this shouldn't be a constant here

gtardif: Yes, that will be true for "docker compose run" containers I think... I'll add that

ndeloof: that's also needed so that contianer created by `docker compose up` can be excluded by docker-compose one_off filters, so still relevant here

ndeloof (approved):

Issue #938, opened 2020-11-20
Local compose: down container order


gtardif:
Using compose command in local context, when runnning `compose up` containers are started in order based on dependencies specified in `depends_on`. 
Similarly, `compose down` should shut down containers in the reverse order.

Issue #965, opened 2020-11-26
Local compose: Add container labels


gtardif:
Add missing container labels used in docker-compose at runtime (to be more runtime-compatible with docker-compose, and these labels make sense to keep compose meta-data at runtime)

- [x] 'com.docker.compose.container-number'
- [x] 'com.docker.compose.oneoff'
- [x] 'com.docker.compose.project'
- [x] 'com.docker.compose.project.working_dir'
- [x] 'com.docker.compose.project.config_files'
- [ ] 'com.docker.compose.project.environment_file'
- [x] 'com.docker.compose.service'
- [x] 'com.docker.compose.version'
- [x] 'com.docker.compose.slug'
- [x] 'com.docker.compose.config-hash'

Oneoff : label is set but hardocded, needs to be changed for container started with `compose run`

PR #968, opened 2020-11-26
Add order to down command

ulyssessouza:
**What I did**
Add order to down command

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/938

/test-aci
/test-windows

**Dependencies**
https://github.com/docker/compose-cli/issues/965 on the 'com.docker.compose.project.config_files' label implementation

### DIFF
--- a/local/compose.go
+++ b/local/compose.go
@@ -204,11 +204,6 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, detach 
 		}
 	}
 
-	err = s.ensureImagesExists(ctx, project)
-	if err != nil {
-		return err
-	}
-
 	err = InDependencyOrder(ctx, project, func(c context.Context, service types.ServiceConfig) error {
 		return s.ensureService(c, project, service)
 	})
--- a/local/compose.go
+++ b/local/compose.go
@@ -29,6 +29,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/compose-spec/compose-go/cli"
 	"github.com/compose-spec/compose-go/types"
 	"github.com/docker/buildx/build"
 	"github.com/docker/cli/cli/config"
@@ -203,7 +204,12 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, detach 
 		}
 	}
 
-	err = inDependencyOrder(ctx, project, func(c context.Context, service types.ServiceConfig) error {
+	err = s.ensureImagesExists(ctx, project)
+	if err != nil {
+		return err
+	}
+
+	err = InDependencyOrder(ctx, project, func(c context.Context, service types.ServiceConfig) error {
 		return s.ensureService(c, project, service)
 	})
 	return err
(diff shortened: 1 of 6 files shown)


local/compose.go
```diff
@@ -163,47 +164,95 @@ func toProgressEvent(jm jsonmessage.JSONMessage, w progress.Writer) {
...
-				Status: progress.Done,
-			})
-			return nil
-		})
+	return project, nil
+}
+
+func loadProjectOptionsFromLabels(c moby.Container) (*cli.ProjectOptions, error) {
+	var configFiles []string
+	if c.Labels[configFilesLabel] == "" {
```
ulyssessouza, line 245: To remove when https://github.com/docker/compose-cli/issues/965 is ready

local/dependencies.go
```diff
@@ -168,6 +210,10 @@ func (g *Graph) AddEdge(source string, destination string) error {
 	return nil
 }
 
+func leaves(g *Graph) []*Vertex {
```
rumpl, line 213: This doesn't seem to be used

local/dependencies.go
```diff
@@ -207,6 +275,27 @@ func (g *Graph) FilterChildren(key string, status ServiceStatus) []*Vertex {
 	return res
 }
 
+func filterParents(g *Graph, k string, s ServiceStatus) []*Vertex {
+	return g.FilterParents(k, s)
+}
+
+// FilterParents returns children of a certain vertex that are in a certain status
```
rumpl, line 282:
```suggestion
// FilterParents returns the parents of a certain vertex that are in a certain status
```

gtardif (changes requested):
I ran into a bug when trying: 
```
$ ./bin/docker --context localctx compose up -f tests/composefiles/demo_multi_port.yaml
$ ./bin/docker --context localctx compose down -f tests/composefiles/demo_multi_port.yaml
stat /Users/gtardif/docker/src/github.com/docker/compose-cli/tests/composefiles/tests/composefiles/demo_multi_port.yaml: no such file or directory
```

labels (being compatible with docker-compose) include config files as provided by users, and working dir if not specified is derived from where the compose file lives: 
"com.docker.compose.project.config_files": "tests/composefiles/demo_multi_port.yaml",
"com.docker.compose.project.working_dir": "/Users/gtardif/docker/src/github.com/docker/compose-cli/tests/composefiles",

We would need to remove the path from composefile, before appending to workingDir. 
There will probably be some crappy corner cases as well there, we'll need a fallback on non ordered down on containers, if we can't read files, same as for stdin.

gtardif:
if we rely on additional labels specifying depends_on metadata rather than re-reading config files, this will not work with stacks started by docker-compose (could also fallback to non ordered down in this case...)

ndeloof:
I suggest (this can come as a follow-up PR) we introduce label `com.docker.compose.dependencies` to store service dependencies metadata. We for sure can retrieve the compose file with those labels, but this file might have been edited in the meantime. Make sense anyway for compatibility with docker-compose.

IIRC `inDependencyOrder` relies only on `service.GetDependencies` so we could have this one support both (labels metadata + read from config file)
We could also make a tiny change in docker-compose so it sets this same label, even the `down` logic would not be updated to limit the effort.

local/dependencies.go
```diff
@@ -37,29 +37,54 @@ const (
...
+	downDirectionTraversalConfig = graphTraversalConfig{
+		extremityNodesFn:            roots,
+		adjacentNodesFn:             getChildren,
+		filterAdjacentByStatusFn:    filterParents,
+		adjacentServiceStatusToSkip: ServiceStarted,
+		targetServiceStatus:         ServiceStopped,
+	}
+)
+
+func inDependencyOrder(ctx context.Context, project *types.Project, traversalConfig graphTraversalConfig, fn func(context.Context, types.ServiceConfig) error) error {
```
rumpl, line 65: I wish we had two functions `inDependencyOrder` and `inReverseDependencyOrder`, the `traversalConfig` is confusing

ndeloof: Was thinking the same. Policy to define travel within the graph is nice, but this should be hidden with higher level vocabulary.
`inDependecyOrder` vs `inReverseDependencyOrder` would be more readable for the price of an extra nested  func call.

local/compose.go
```diff
@@ -115,56 +121,127 @@ func getContainerName(c moby.Container) string {
...
+			w.Event(progress.Event{
+				ID:     cName,
+				Text:   "Removing",
+				Status: progress.Working,
+			})
+			if err = s.apiClient.ContainerRemove(ctx, c.ID, moby.ContainerRemoveOptions{}); err != nil {
 				return err
 			}
-			w.Event(progress.RemovedEvent(getContainerName(container)))
+			w.Event(progress.Event{
```
ndeloof, line 189: why not use more concise `RemovedEvent` and comparable?

local/compose.go
```diff
@@ -115,56 +121,127 @@ func getContainerName(c moby.Container) string {
...
+	eg, _ := errgroup.WithContext(ctx)
+	w := progress.ContextWriter(ctx)
+
+	project, err := s.projectFromContainerLabels(ctx, projectName)
+	if err != nil || project == nil {
+		return err
+	}
+
+	err = InReverseDependencyOrder(ctx, project, func(c context.Context, service types.ServiceConfig) error {
+		filter := filters.NewArgs(projectFilter(project.Name), serviceFilter(service.Name))
```
ndeloof, line 133: tl;dr: I'll look into this "detail" later, just a note for myself

to avoid many calls to `ContainerList` API with service filter we could list all project containers and filter the returned slice. 

I guess this would be a natural design if one could write something like
```go
containers.filter (c) -> c.Labels[serviceLabel] == service.Name
```
but thanks to Golang it is simpler to invoke API multiple times to avoid 10 extra line of boiler plate slice-filterting code

ndeloof (changes requested):
LGTM
please revert removal for `progress.RemovedEvent` and others - probably caused by rebase conflict

ndeloof (approved):

local/compose.go
```diff
@@ -98,7 +99,12 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, detach
 		}
 	}
 
-	err = inDependencyOrder(ctx, project, func(c context.Context, service types.ServiceConfig) error {
+	err = s.ensureImagesExists(ctx, project)
```
gtardif, line 102: Nit: this is already at the start of the method ; probably a rebase issue, remove this `ensureImagesExists()`

gtardif: will fix this and rebase anyway...

gtardif (approved):

PR #980, opened 2020-11-27
filter compose project to remove all services not required by command

ndeloof:
**What I did**
`up` can get a subset of compose services to run as arg.
for this purpose, the compose app model is "filtered" to remove all unrelated services, so that the implementation code doesn't have to handle this.

*(not mandatory) A picture of a cute animal, if possible in relation with what you did**
image


gtardif (approved):
LGTM, just linter to fix

PR #988, opened 2020-11-30
Add platform when running containers

rumpl:
**What I did**

* added a `--platform` flag to `docker run`, only shown for local context
* added the platform to grpc


gtardif (approved):
Open question: should we parse platform in compose-cli or just pass it to the engine and let the error message bubble up if any ? 
it's just more error handling and duplication at https://github.com/docker/compose-cli/pull/988/files#diff-ee0973c422f8a2edb935ba75fef5bc9a2d874c04cde4ce1166e1f871ddff708cR94 and https://github.com/docker/compose-cli/pull/988/files#diff-dece6d607fbdaa34e820e5b7b2fa6d7796f3e4e318262968d15d65e630a8f351R173

Otherwise, LGTM

gtardif:
Need to keep the platform parsing in here to transform spec.Platform => string.

Issue #990, opened 2020-11-30
Local compose: bug in using external networks


gtardif:
from compose acceptance tests: 

`Error response from daemon: network project1_networks_foo not found`
It seems we have a prefix `project_` added to network name, regardless of specified network name or external network

Issue #975, opened 2020-11-27
Compose local: Error when referenced network not found


gtardif:
From docker-compose acceptance tests: 
* non-exitsting external network ` assert 'declared as external, but could not be found' in result.stderr`
* service network undefined: `assert 'Service "web" uses an undefined network "foo"' in result.stderr`

PR #991, opened 2020-11-30
Compose network error

gtardif:
**What I did**
* Detect when external network are not available when doing `docker compose up`
* Avid duplicate prefixing of network name with `myproject_`

**Related issue**
https://github.com/docker/compose-cli/issues/975
https://github.com/docker/compose-cli/issues/990

### DIFF
--- a/local/compose.go
+++ b/local/compose.go
@@ -558,6 +558,10 @@ func (s *composeService) ensureNetwork(ctx context.Context, n types.NetworkConfi
 	_, err := s.apiClient.NetworkInspect(ctx, n.Name, moby.NetworkInspectOptions{})
 	if err != nil {
 		if errdefs.IsNotFound(err) {
+			if n.External.External {
+				return fmt.Errorf("Network %s declared as external, but could not be found", n.Name)
+			}
+
 			createOpts := moby.NetworkCreate{
 				// TODO NameSpace Labels
 				Labels:     n.Labels,
--- a/local/compose.go
+++ b/local/compose.go
@@ -59,7 +59,7 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, detach 
 		return err
 	}
 	for k, network := range project.Networks {
-		if !network.External.External && network.Name != "" {
+		if !network.External.External && network.Name == k {
 			network.Name = fmt.Sprintf("%s_%s", project.Name, k)
 			project.Networks[k] = network
 		}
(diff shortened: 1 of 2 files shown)


local/compose.go
```diff
@@ -558,6 +558,10 @@ func (s *composeService) ensureNetwork(ctx context.Context, n types.NetworkConfi
 	_, err := s.apiClient.NetworkInspect(ctx, n.Name, moby.NetworkInspectOptions{})
 	if err != nil {
 		if errdefs.IsNotFound(err) {
+			if n.External.External {
+				return fmt.Errorf("Network %s declared as external, but could not be found", n.Name)
```
rumpl, line 562: Shouldn't the linter fail on an error that starts with uppercase?

gtardif: fixed. Indeed I thought linter did detect these things, it seems not (just double checked the linter is active in /local).

local/convergence.go
```diff
@@ -240,8 +240,12 @@ func (s *composeService) runContainer(ctx context.Context, project *types.Projec
 		return err
 	}
 	id := created.ID
-	for net := range service.Networks {
-		name := fmt.Sprintf("%s_%s", project.Name, net)
+	for netName := range service.Networks {
+		network := project.Networks[netName]
+		name := fmt.Sprintf("%s_%s", project.Name, netName)
+		if network.External.External {
+			name = network.Name
```
ndeloof, line 247: IIRC, network name should be the one set in compose file (when set) even if this is a managed one (external=false). Would need to double check with docker-compose

gtardif: Indeed, tested with docker-compose, and if network name is specified in yaml, it's used as is, no prefix

gtardif: This network name setting needed to be fixed earlier, when creating/checking networks, in compose.go. Here we can just use the network.name that is set in the project.

local/compose.go
```diff
@@ -59,7 +59,7 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, detach
 		return err
 	}
 	for k, network := range project.Networks {
-		if !network.External.External && network.Name != "" {
+		if !network.External.External && network.Name == k {
```
gtardif, line 62: ATM we can't tell if the network name was set by the user or set by default as the key value. Assuming if name = key that the name was not set by user. 
Will do a separate PR to update compose-go and let network name empty if not set by user, so we can differentiate the 2 cases

ndeloof (approved):

Issue #931, opened 2020-11-20
Local compose: push command


gtardif:
Using compose command in local context, I should be able to run `docker compose push` and push images in my compose file, like in `docker-compose`.

Other backends (cloud backends) will return not implemented on this command.

ndeloof:
Considering the "build, test local, push, run on the cloud" scenario, this build process _should_ be designed so we can retrieve the image sha256 and use those for the "cloud" part.
This would help fix https://github.com/docker/compose-cli/issues/612
Maybe we can store image shas into a `.docker-compose` file at project root, or something comparable, to avoid makeing changes to user's `compose.yaml` file

PR #995, opened 2020-12-01
introduce `compose push` command

ndeloof:
**What I did**
Introduce `compose push` command and local implementation

**Related issue**
https://github.com/docker/compose-cli/issues/931

image


local/compose.go
```diff
@@ -65,6 +70,100 @@ func (s *composeService) Build(ctx context.Context, project *types.Project) erro
...
+			})
+			if err != nil {
+				return err
+			}
+			dec := json.NewDecoder(stream)
+			for {
+				var jm jsonmessage.JSONMessage
+				if err := dec.Decode(&jm); err != nil {
+					if err == io.EOF {
+						break
```
rumpl, line 131: Wouldn't this return the error? Don't we want to `return nil` here?

ndeloof: nope, this means end of the json events stream

gtardif (changes requested):
* Need file and workdir flags 
* when pushing a compose file with 3 images, push stopped after the first one
* on second push, it did push 3 images in parallel, but the progress seems overlapped between 3 images (see below) 

changing image names in test/composefiles/demo-multi-ports.yaml from `gtardif/sentences-db` to `gtardif/sentences-db-testpush`, same suffix for other images

First push: 
```
 $ ../../bin/docker --context localctx compose push 
[+] Running 0/11
 ⠏                                                                                                                                                                                                             ⠼ eb80542d2af1 Pushed                                                                                                                                                                                         ⠼ 3a02a7e71bd2 Mounted from gtardif/sentences-db                                                                                                                                                              ⠼ a11407e16748 Mounted from gtardif/sentences-db                                                                                                                                                              ⠼ ede3ce660009 Mounted from gtardif/sentences-db                                                                                                                                                              ⠼ ef143b46e2fb Mounted from gtardif/sentences-db                                                                                                                                                              ⠼ 7fb8888f8580 Mounted from gtardif/sentences-db                                                                                                                                                             ⠼ f3b7999da396 Mounted from gtardif/sentences-db                                                                                                                                                              ⠼ f39dbbe69a84 Mounted from gtardif/sentences-db                                                                                                                                                              ⠼ 0ec25d65c4dc Mounted from gtardif/sentences-db                                                                                                                                                              ⠼ 2aebd096e0e2 Mounted from gtardif/sentences-db                                                                                                                                                             ```
and then only the db images has been pushed
```
$ docker pull gtardif/sentences-web-testpush
Using default tag: latest
Error response from daemon: manifest for gtardif/sentences-web-testpush:latest not found: manifest unknown: manifest unknown
$ docker pull gtardif/sentences-db-testpush
Using default tag: latest
latest: Pulling from gtardif/sentences-db-testpush
Digest: sha256:b849665d557e4ff83445c37d794556a50b495910f5dd8dfdd71dd7e86dd24685
Status: Image is up to date for gtardif/sentences-db-testpush:latest
docker.io/gtardif/sentences-db-testpush:latest
```

2nd push (stopped with Ctrl+C cause I wanted to check image pull) 
Note `[+] Running 0/10Waiting    `

```
$ ../../bin/docker --context localctx compose push 
[+] Running 0/10
 ⠏  The push refers to repository [docker.io/gtardif/sentences-web-testpush]                                                                                                                                   ⠴ d4852371a647 Pushing                                                      [==================================================>]  430.6kB                                                                     ⠴ 8c0845160e01 Pushing                                                      [==================================================>]  7.192MB                                                                     ⠴ 11066148d212 Pushed                                                                                                                                                                                          ⠴ 984032f5df2f Pushing                                                      [=========>                                         ]  43.22MB/217.8MB                                                            9.5s
 ⠴ 5e92f64ba125 Pushing                                                      [==>                                                ]  10.93MB/256.4MB                                                            9.5s
 ⠴ d74c72ec04b8 Waiting                                                                                                                                                                                         ⠴ f6253634dc78 Waiting                                                                                                                                                                                         ⠴ 9069f84dbbe9 Waiting                                                                                                                                                                                        [+] Running 0/10Waiting                                                                                                                                                                                         ⠙  The push refers to repository [docker.io/gtardif/sentences-web-testpush]                                                                                                                                  10.1s6 ⠧ d4852371a647 Pushing                                                      [==================================================>]  430.6kB                                                                     ⠧ 8c0845160e01 Pushing                                                      [==================================================>]  7.192MB                                                                     ⠧ 11066148d212 Pushed                                                                                                                                                                                          ⠧ 984032f5df2f Pushing                                                      [==========>                                        ]  44.28MB/217.8MB                                                            9.6s
 ⠧ 5e92f64ba125 Pushing                                                      [==>                                                ]  10.93MB/256.4MB                                                            9.6s
 ⠧ d74c72ec04b8 Waiting                                                                                                                                                                                         ⠧ f6253634dc78 Waiting                                                            
```

File: local/compose.go (hunk 1 above)
Comment (rumpl) [reply local/compose.go:131]: Yeah but the end of the stream is not an error right?

File: local/compose.go [hunk 2]
```diff
@@ -65,6 +70,100 @@ func (s *composeService) Build(ctx context.Context, project *types.Project) erro
 	return s.build(ctx, project, opts)
 }
 
+func (s *composeService) Push(ctx context.Context, project *types.Project) error {
+	configFile, err := config.Load(config.Dir())
+	if err != nil {
+		return err
+	}
+	eg, ctx := errgroup.WithContext(ctx)
```
rumpl, line 78:
```suggestion
	eg, _ := errgroup.WithContext(ctx)
```

The `ctx` returned by the `WithContext` doesn't end on ctrl+c

ndeloof: ack. But then I wonder: an error in one of the goroutines will have errgroup to invoke it's `cancel` to stop others. If we don't use this context we will have to wait for all of them to complete. How to get both?

ndeloof: Looking at `WithCancel` source it seems it has been designed to propagate cancel events from parent, so this _should_ work. But maybe I missunderstand how to use this

ndeloof: just confirmed [ctrl+c] works with context returned by `errgroup.WithContext`

local/compose.go
```diff
@@ -65,6 +70,103 @@ func (s *composeService) Build(ctx context.Context, project *types.Project) erro
...
+	}
+	if jm.Error != nil {
+		status = progress.Error
+		text = jm.Error.Message
+	}
+	if jm.Progress != nil {
+		text = jm.Progress.String()
+	}
+	w.Event(progress.Event{
+		ID:         fmt.Sprintf("%s: %s", prefix, jm.ID),
```
gtardif, line 163: Nit: might be nicer adding "Pushing" as `fmt.Sprintf("Pushing %s: %s", prefix, jm.ID)` ,
```
[+] Running 0/24
 ⠴ web: d4852371a647 Layer already exists
 ⠴ web: 8c0845160e01 Layer already exists
 ⠴ words: 80f0fd999f52 Mounted from gtardif/sentences-api-testpush    
```

gtardif (changes requested):
Sorry, still weird case, when one image is not available locally, it's just silently ignored (tried with 1 image not build locally and 2 images available). We should check images availability before pushing

local/compose.go
ndeloof: Right, moby json.progress events hardly match the progresswriter model :P
By the way, I also noticed we don't stop when such an error happens (docker-compose has an option to ignore pull failures, we could do the same)

gtardif (approved):

Issue #970, opened 2020-11-26
Local compose: attach to logs when 'compose up' by default


gtardif:
if the `--detach` flag is not specified, `docker compose up` should attach to compose logs by default.

ndeloof:
I wonder about interactions with the progressWriter ...
Or shall we just chain `logs` after `up` completed? Which is not the same as `docker-compose up`

flaviostutz:
Isn't this issue solved by the mentioned PR #998?

PR #998, opened 2020-12-02
Attach to container on `up` and pipe logs to stdout

ndeloof:
**What I did**
`up` do (re)create containers, then attach (or collect logs for running ones) and eventually start.
Doing so we reproduce docker-compose behaviour
also handle SIGTERM to run `down`

**Related issue**
https://github.com/docker/compose-cli/issues/970

image

### DIFF
--- a/cli/cmd/compose/up.go
+++ b/cli/cmd/compose/up.go
@@ -18,13 +18,14 @@ package compose
 
 import (
 	"context"
+	"github.com/docker/compose-cli/progress"
+	"os"
 
 	"github.com/compose-spec/compose-go/cli"
 	"github.com/spf13/cobra"
 
 	"github.com/docker/compose-cli/api/client"
 	"github.com/docker/compose-cli/context/store"
-	"github.com/docker/compose-cli/progress"
 )
 
 func upCommand(contextType string) *cobra.Command {
@@ -54,25 +55,26 @@ func runUp(ctx context.Context, opts composeOptions, services []string) error {
 		return err
 	}
 
-	_, err = progress.Run(ctx, func(ctx context.Context) (string, error) {
-		options, err := opts.toProjectOptions()
-		if err != nil {
-			return "", err
-		}
-		project, err := cli.ProjectFromOptions(options)
-		if err != nil {
-			return "", err
-		}
-		if opts.DomainName != "" {
-			// arbitrarily set the domain name on the first service ; ACI backend will expose the entire project
-			project.Services[0].DomainName = opts.DomainName
-		}
+	options, err := opts.toProjectOptions()
+	if err != nil {
+		return err
+	}
+	project, err := cli.ProjectFromOptions(options)
+	if err != nil {
+		return err
+	}
+	if opts.DomainName != "" {
+		// arbitrarily set the domain name on the first service ; ACI backend will expose the entire project
+		project.Services[0].DomainName = opts.DomainName
+	}
 
-		err = filter(project, services)
-		if err != nil {
-			return "", err
-		}
-		return "", c.ComposeService().Up(ctx, project, opts.Detach)
+	err = filter(project, services)
+	if err != nil {
+		return err
+	}
+
+	_, err = progress.Run(ctx, func(ctx context.Context) (string, error) {
+		return "", c.ComposeService().Up(ctx, project, opts.Detach, os.Stdout)
 	})
 	return err
 }
--- a/cli/cmd/compose/up.go
+++ b/cli/cmd/compose/up.go
@@ -18,14 +18,18 @@ package compose
 
 import (
 	"context"
-	"github.com/docker/compose-cli/progress"
+	"errors"
+	"fmt"
+	"io"
 	"os"
 
 	"github.com/compose-spec/compose-go/cli"
+	"github.com/compose-spec/compose-go/types"
 	"github.com/spf13/cobra"
 
 	"github.com/docker/compose-cli/api/client"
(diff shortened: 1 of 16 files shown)


local/compose.go
```diff
@@ -175,40 +176,151 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, detach
...
-			project.Networks[k] = network
-		}
-		network.Labels = network.Labels.Add(networkLabel, k)
-		network.Labels = network.Labels.Add(projectLabel, project.Name)
-		network.Labels = network.Labels.Add(versionLabel, ComposeVersion)
-		err := s.ensureNetwork(ctx, network)
-		if err != nil {
-			return err
+
+	_, err = progress.Run(ctx, func(ctx context.Context) (string, error) {
```
rumpl, line 180: Having `progress.Run` in the backend means that the gRPC side has no way to know what is going on when doing `compose up`, ideally we would pass an `io.Writer` if we want to attach

local/compose.go
```diff
@@ -175,40 +176,151 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, detach
...
+	})
+	if detach || err != nil {
+		return err
+	}
+
+	return s.attach(ctx, project)
+}
+
+func (s *composeService) attach(ctx context.Context, project *types.Project) error {
+	consumer := formatter.NewLogConsumer(ctx, os.Stdout)
```
rumpl, line 233: See comment above, the `os.Stdout` is problematic if we call compose up over gRPC

cli/cmd/compose/up.go
```diff
@@ -54,25 +53,22 @@ func runUp(ctx context.Context, opts composeOptions, services []string) error {
 		return err
 	}
 
-	_, err = progress.Run(ctx, func(ctx context.Context) (string, error) {
```
gtardif, line 57: If we remove this here, we need to add the progres.Run in all backends. Currently this hides all the display in ACI/ECS compose up, as they will fallback to noop progress.

local/compose.go
ndeloof: ack.
Unfortunately this means we can't implement `Up attached` as a single func, as we need progressWriter to complete before we can pipe logs to the console

cli/cmd/compose/up.go
ndeloof: so the "WiP" status of this PR :P
I'm looking for a way to concile progress with attached logs

ndeloof: seems using progressWriter.Stop() just works
Last issue I need to fix is running `down` on ctrl+c, which never complete

local/compose.go
```diff
@@ -204,12 +210,119 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, detach
...
+			ShowStdout: true,
+			ShowStderr: true,
+			Follow:     true,
+		})
+		if err != nil {
+			return nil, err
+		}
+		reader = logs
+	} else {
+		cnx, err := s.apiClient.ContainerAttach(ctx, container.ID, moby.ContainerAttachOptions{
```
rumpl, line 307: Question: not sure I understand this?

ndeloof: This is implemented the same way docker-compose does (same as `docker run -i`).
Note: ContainerLogs might return empty if container is configured with an inadequate log-driver.

local/compose.go
```diff
@@ -204,12 +210,119 @@ func (s *composeService) Up(ctx context.Context, project *types.Project, detach
...
+			Stream: true,
+			Stdin:  true,
```
rumpl, line 309: Do we need stdin here?

local/container.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (l containerStdout) Close() error {
+	l.HijackedResponse.Close()
+	return nil
+}
+
+var _ io.WriteCloser = containerStdin{}
+
+type containerStdin struct {
```
rumpl, line 54: I can't seem to find where this is used

local/compose.go
ndeloof: not for `up` but we will for `run`, and having this set by default avoid having multiple execution paths (as we won't send anything on stdin stream this has no impact)

local/container.go
ndeloof: not needed for `up` but we will for `run`. I imported this from docker/c2

local/compose.go
rumpl: Aaaa, all this code makes sense now!

rumpl (approved):

gtardif (commented):
Trying this with the ELK awesome-compose example, I had: 
```elasticsearch-logstash-kibana [attach|…47⚑ 4] 13:22  $ 
../../../bin/docker --context localctx compose up
[+] Running 4/4
 ⠿ Network "elasticsearch-logstash-kibana_elastic"  Created                                                                      ⠿ elasticsearch-logstash-kibana_elasticsearch_1    Created                                                                     
 ⠿ elasticsearch-logstash-kibana_logstash_1         Created                                                                     
 ⠿ elasticsearch-logstash-kibana_kibana_1           Created                                                                     Attaching to elasticsearch-logstash-kibana_logstash_1, elasticsearch-logstash-kibana_kibana_1, elasticsearch-logstash-kibana_elasticsearch_1
Error response from daemon: Ports are not available: listen tcp 0.0.0.0:5601: bind: address already in use
```
Need to investigate more where this comes from...

gtardif (approved):
ELK Error was a local setup issue, managed to run this without any pb

PR #1003, opened 2020-12-03
Add local `compose run` command

aiordache:
Implement  'docker compose run <service name> command>

* execute a command on a one-off service container.
* starts only the necessary dependencies (volumes and networks includes).

```yaml
version: '3.8'
services:
  back:
    image: nginx
    command: echo "Hello there!!"
    depends_on: 
      - db
    networks: 
      - backnet

  db:
    image: nginx
    networks: 
      - backnet
    volumes:
      - data:/test

  front:
    image: nginx
    networks: 
      - frontnet
  
networks: 
  frontnet: {}
  backnet: {}

volumes: 
  data: {}

```
```
$ docker compose run back
[+] Running 2/2
 ⠸ test_db_1                      Running                                                               -9223372036.9s
 ⠿ test_back_run_671a29f3964b  Created                                                                         0.0s
Hello there!!

```
```
$ docker compose run back sh -c "echo Hi!"
[+] Running 3/3
 ⠿ Network "test_backnet"         Created                                                                         0.0s
 ⠿ test_db_1                      Started                                                                         0.3s
 ⠿ test_back_run_ddc38dc7670f  Created                                                                         0.0s
Hi!

```
Second run:
```
$ docker compose run back sh -c "echo Hi!"
[+] Running 2/2
 ⠸ test_db_1                      Running                                                               -9223372036.9s
 ⠿ test_back_run_279ea977cfb2  Created                                                                         0.0s
Hi!

```

```
$ docker ps -a
CONTAINER ID   IMAGE     COMMAND                  CREATED         STATUS                     PORTS     NAMES
e7325ddb266b   nginx     "/docker-entrypoint.…"   2 minutes ago   Exited (0) 2 minutes ago             test_back_run_279ea977cfb2
e03f2eac1706   nginx     "/docker-entrypoint.…"   5 minutes ago   Exited (0) 5 minutes ago             test_back_run_ddc38dc7670f
147938a36404   nginx     "/docker-entrypoint.…"   5 minutes ago   Up 5 minutes               80/tcp    test_db_1

```

### DIFF
--- a/cli/cmd/compose/compose.go
+++ b/cli/cmd/compose/compose.go
@@ -90,6 +90,7 @@ func Command(contextType string) *cobra.Command {
 		listCommand(),
 		logsCommand(),
 		convertCommand(),
+		runCommand(),
 	)
 
 	if contextType == store.LocalContextType || contextType == store.DefaultContextType {
@@ -99,7 +100,7 @@ func Command(contextType string) *cobra.Command {
 			pullCommand(),
 		)
 	}
-
+	command.Flags().SetInterspersed(false)
 	return command
 }
 
--- a/cli/cmd/compose/run.go
+++ b/cli/cmd/compose/run.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/compose-spec/compose-go/types"
 	"github.com/spf13/cobra"
 
 	"github.com/docker/compose-cli/api/compose"
@@ -72,17 +73,30 @@ func runRun(ctx context.Context, opts runOptions) error {
 		return err
 	}
 
+	dependencies := []types.ServiceConfig{}
+	originalServices := project.Services
 	containerID, err := progress.Run(ctx, func(ctx context.Context) (string, error) {
-		return c.ComposeService().CreateOneOffContainer(ctx, project, compose.RunOptions{
-			Name:    opts.Name,
-			Command: opts.Command,
-		})
+		for _, service := range originalServices {
+			if service.Name != opts.Name {
+				dependencies = append(dependencies, service)
+			}
+		}
+		project.Services = types.Services(dependencies)
+		if err := c.ComposeService().Create(ctx, project); err != nil {
+			return "", err
+		}
+		if err := c.ComposeService().Start(ctx, project, nil); err != nil {
+			return "", err
+		}
+		return "", nil
 	})
 	if err != nil {
 		return err
 	}
+
+	project.Services = originalServices
 	// start container and attach to container streams
-	err = c.ComposeService().Run(ctx, containerID, opts.Detach)
+	containerID, err = c.ComposeService().RunOneOffContainer(ctx, project, compose.RunOptions{Name: opts.Name, Command: opts.Command, Detach: opts.Detach})
 	if err != nil {
 		return err
 	}
--- a/cli/cmd/compose/run.go
+++ b/cli/cmd/compose/run.go
@@ -18,11 +18,12 @@ package compose
 
 import (
 	"context"
-	"fmt"
+	"os"
 
 	"github.com/compose-spec/compose-go/types"
 	"github.com/spf13/cobra"
 
+	"github.com/docker/compose-cli/api/client"
 	"github.com/docker/compose-cli/api/compose"
 	"github.com/docker/compose-cli/progress"
 )
@@ -74,20 +75,7 @@ func runRun(ctx context.Context, opts runOptions) error {
 
 	originalServices := project.Services
 	_, err = progress.Run(ctx, func(ctx context.Context) (string, error) {
-		dependencies := types.Services{}
-		for _, service := range originalServices {
-			if service.Name != opts.Name {
-				dependencies = append(dependencies, service)
-			}
-		}
-		project.Services = dependencies
-		if err := c.ComposeService().Create(ctx, project); err != nil {
-			return "", err
-		}
-		if err := c.ComposeService().Start(ctx, project, nil); err != nil {
-			return "", err
-		}
-		return "", nil
+		return "", startDependencies(ctx, c, project, opts.Name)
 	})
 	if err != nil {
 		return err
@@ -95,13 +83,32 @@ func runRun(ctx context.Context, opts runOptions) error {
 
 	project.Services = originalServices
 	// start container and attach to container streams
-	runOpts := compose.RunOptions{Name: opts.Name, Command: opts.Command, Detach: opts.Detach, AutoRemove: opts.Remove}
-	containerID, err := c.ComposeService().RunOneOffContainer(ctx, project, runOpts)
-	if err != nil {
+	runOpts := compose.RunOptions{
+		Name:       opts.Name,
+		Command:    opts.Command,
+		Detach:     opts.Detach,
+		AutoRemove: opts.Remove,
+		Writer:     os.Stdout,
+		Reader:     os.Stdin,
+	}
+	return c.ComposeService().RunOneOffContainer(ctx, project, runOpts)
+}
+
+func startDependencies(ctx context.Context, c *client.Client, project *types.Project, requestedService string) error {
+	originalServices := project.Services
+	dependencies := types.Services{}
+	for _, service := range originalServices {
+		if service.Name != requestedService {
+			dependencies = append(dependencies, service)
+		}
+	}
+	project.Services = dependencies
+	if err := c.ComposeService().Create(ctx, project); err != nil {
 		return err
 	}
-	if opts.Detach {
-		fmt.Printf("%s", containerID)
+	if err := c.ComposeService().Start(ctx, project, nil); err != nil {
+		return err
 	}
 	return nil
+
 }
--- a/cli/cmd/compose/run.go
+++ b/cli/cmd/compose/run.go
@@ -36,10 +36,6 @@ type runOptions struct {
 	ConfigPaths []string
 	Environment []string
 	Detach      bool
-	Publish     []string
-	Labels      []string
-	Volumes     []string
-	NoDeps      bool
 	Remove      bool
 }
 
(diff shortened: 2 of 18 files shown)


local/run.go
```diff
@@ -0,0 +1,170 @@
...
+	"sort"
+
+	"github.com/compose-spec/compose-go/types"
+	"github.com/docker/compose-cli/api/compose"
+)
+
+func (s *composeService) Run(ctx context.Context, project *types.Project, opts compose.RunOptions) (string, error) {
+	name := opts.Name
+
+	fmt.Printf("Running container for service %q\n", name)
```
rumpl, line 33: This shouldn't be in the backend code.

cli/cmd/compose/run.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func runRun(ctx context.Context, opts runOptions) error {
+
+	c, err := client.New(ctx)
+	if err != nil {
+		return err
+	}
+
+	_, err = progress.Run(ctx, func(ctx context.Context) (string, error) {
```
ndeloof, line 94: `compose run` has to attach stdin/stdout to the service container, not compatible with running inside a progress context.
see UpCommand's `CreateAndStart`

ndeloof: see https://github.com/docker/compose-cli/pull/1015/files#diff-338f90b88c22fe2a1befe8f111f6fc425f7e7d4f0ee375d23d925d5246a98d7bR405
will need to be exposed on compose/api so that you can implement `run` as `Create+Attach+Start`

gtardif: Should we expose the attach method, or have something like to the current up flow, but add an API `compose.Start(ctx, project, servicename, consumer)`, similar to compose.Up(ctx, project, consumer)` ?

ndeloof: Imho it would make sense to have a `compose.Attach()` method in the API, as here we don't just want to collect stdout, but also want to bind stdin.

ndeloof: `docker-compose run` is interactive by default

local/compose/run.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	//apply options to service config
+	updateOneOffServiceConfig(&service, project.Name, opts)
+
+	err = s.createContainer(ctx, project, service, service.ContainerName, 1)
+	if err != nil {
+		return "", err
+	}
+	err = s.apiClient.ContainerStart(ctx, service.ContainerName, apitypes.ContainerStartOptions{})
```
ndeloof, line 61: attach must take place before start
see https://github.com/docker/cli/blob/b350e14b1f83e1870a234ff2c554450707218204/cli/command/container/run.go#L150-L167

cli/cmd/compose/compose.go
```diff
@@ -99,7 +100,7 @@ func Command(contextType string) *cobra.Command {
 			pullCommand(),
 		)
 	}
-
+	command.Flags().SetInterspersed(false)
```
gtardif, line 103: what was this needed for ?

cli/cmd/compose/run.go
```diff
@@ -0,0 +1,136 @@
...
+	runCmd.Flags().StringArrayVarP(&opts.Publish, "publish", "p", []string{}, "Publish a container's port(s). [HOST_PORT:]CONTAINER_PORT")
+	runCmd.Flags().StringVar(&opts.Name, "name", "", "Assign a name to the container")
+	runCmd.Flags().BoolVar(&opts.NoDeps, "no-deps", false, "Don't start linked services.")
+	runCmd.Flags().StringArrayVarP(&opts.Labels, "label", "l", []string{}, "Set meta data on a container")
+	runCmd.Flags().StringArrayVarP(&opts.Volumes, "volume", "v", []string{}, "Volume. Ex: storageaccount/my_share[:/absolute/path/to/target][:ro]")
+	runCmd.Flags().BoolVarP(&opts.Detach, "detach", "d", false, "Run container in background and print container ID")
+	runCmd.Flags().StringArrayVarP(&opts.Environment, "env", "e", []string{}, "Set environment variables")
+	runCmd.Flags().BoolVar(&opts.Remove, "rm", false, "Automatically remove the container when it exits")
+
+	//addComposeCommonFlags(runCmd.Flags(), &opts.ComposeOpts)
```
gtardif, line 83: remove this line

cli/cmd/compose/run.go
```diff
@@ -0,0 +1,136 @@
...
+
+	runCmd.Flags().SetInterspersed(false)
+	return runCmd
+}
+
+func runRun(ctx context.Context, opts runOptions) error {
+	// target service
+	services := []string{opts.Name}
```
gtardif, line 91: prefer setting targetService as variable name, and not require comment

cli/cmd/compose/run.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+
+			if len(args) > 1 {
+				opts.Command = args[1:]
+			}
+			opts.Name = args[0]
+			return runRun(cmd.Context(), opts)
+		},
+	}
+	runCmd.Flags().StringVar(&opts.WorkingDir, "workdir", "", "Work dir")
```
gtardif, line 72: Also propose `-f` option like for up, next to workdir

cli/cmd/compose/run.go
gtardif: or even inline with the call to filter()

cli/main.go
```diff
@@ -150,7 +150,8 @@ func main() {
 	opts.AddConfigFlags(root.PersistentFlags())
 	opts.AddContextFlags(root.PersistentFlags())
 	root.Flags().BoolVarP(&opts.Version, "version", "v", false, "Print version information and quit")
-
+	root.PersistentFlags().SetInterspersed(false)
```
gtardif, line 153: What was this needed for ?

local/compose/run.go
```diff
@@ -0,0 +1,250 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return service.ContainerName, err
+}
+
+func (s *composeService) Run(ctx context.Context, container string, detach bool) error {
+	if detach {
+		// start container
```
gtardif, line 67: comment not needed

local/compose/run.go
```diff
@@ -0,0 +1,250 @@
...
+			project.Networks[k] = network
+		}
+		network.Labels = network.Labels.Add(networkLabel, k)
+		network.Labels = network.Labels.Add(projectLabel, project.Name)
+		network.Labels = network.Labels.Add(versionLabel, ComposeVersion)
+
+		err := s.ensureNetwork(ctx, network)
+		if err != nil {
+			return err
+		}
```
gtardif, line 182: Need to find a better way to avoid duplicating all this. (Happy to pair program on this a bit if you like)

local/compose/run.go
```diff
@@ -0,0 +1,250 @@
...
+			project.Volumes[k] = volume
+		}
+		volume.Labels = volume.Labels.Add(volumeLabel, k)
+		volume.Labels = volume.Labels.Add(projectLabel, project.Name)
+		volume.Labels = volume.Labels.Add(versionLabel, ComposeVersion)
+		err := s.ensureVolume(ctx, volume)
+		if err != nil {
+			return err
+		}
+	}
```
gtardif, line 205: Need to find a better way to avoid duplicating all this. (Happy to pair program on this a bit if you like)

cli/cmd/compose/compose.go
gtardif: OK, this is for passing commands after the service name

local/compose/create.go
```diff
@@ -44,6 +44,20 @@ func (s *composeService) Create(ctx context.Context, project *types.Project) err
 		return err
 	}
 
+	if err := s.ensureProjectNetworks(ctx, project); err != nil {
+		return err
+	}
+
+	if err := s.ensureProjectNetworks(ctx, project); err != nil {
+		return err
+	}
```
ndeloof, line 53: Ran twice, we can be sure networks will be set :P

local/compose/run.go
```diff
@@ -0,0 +1,141 @@
...
+		return "", err
+	}
+	if err := s.ensureProjectVolumes(ctx, project); err != nil {
+		return "", err
+	}
+	if err = s.ensureRequiredServices(ctx, project, service); err != nil {
+		return "", err
+	}
+
+	updateOneOffServiceConfig(&service, project.Name, opts)
```
ndeloof, line 50: Why not move this on the command line side, so that we can apply any possible flag to override the service definition, and not pass `compose.RunOptions` to the backend. even hosted in the `api/compose` package those really look like a CLI flag mapping.

local/compose/run.go
```diff
@@ -0,0 +1,141 @@
...
+		slugLabel:   slug,
+		oneoffLabel: "True",
+	}
+	service.Tty = true
+	service.StdinOpen = true
+}
+
+func (s *composeService) ensureRequiredServices(ctx context.Context, project *types.Project, service types.ServiceConfig) error {
+	err := InDependencyOrder(ctx, project, func(c context.Context, svc types.ServiceConfig) error {
+		if svc.Name != service.Name { // only start dependencies, not service to run one-off
```
ndeloof, line 132: This will not just crate dependencies, but also services that have no relation with the one we want to run.
Will also try to create a service that depends on the target one and we skipped here

local/compose/run.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	if err := s.ensureProjectNetworks(ctx, project); err != nil {
+		return "", err
+	}
+	if err := s.ensureProjectVolumes(ctx, project); err != nil {
+		return "", err
+	}
+	if err = s.ensureRequiredServices(ctx, project, service); err != nil {
+		return "", err
+	}
```
ndeloof, line 48: AFAICT this is equivalent to `s.Create(filteredProject)` Where `filteredProject` is a subset of the original project to only contains run target (transitive) dependencies

ndeloof (changes requested):
Reading the code it seems to me we will create too much services for a run target.
Would be saner to filter the project model starting from run target to only keep dependencies (transitiviely) as a list of "actually required services" then invoke `Create`. 

update: my bad, `setup` on top of the command line Command do filter the project model for service and dependencies

local/compose/run.go
gtardif: yes, makes sense

local/compose/run.go
gtardif: I'll look into this, to reuse `s.Create(filteredProject)` I guess we only need to remove the required service to we only start dependencies at this stage

local/compose/run.go
gtardif: dependencies are filtered on the cli side, same as when calling `compose up SERVICE SERVICE`.

cli/cmd/compose/run.go
```diff
@@ -0,0 +1,99 @@
...
+	err = c.ComposeService().Run(ctx, containerID, opts.Detach)
+	if err != nil {
+		return err
+	}
+	if opts.Detach {
+		fmt.Printf("%s", containerID)
+		return nil
+	}
+	if opts.Remove {
+		return c.ContainerService().Delete(ctx, containerID, containers.DeleteRequest{
```
ndeloof, line 94: Can't we just set AutoRemove ?

local/compose/create.go
gtardif: Fixed, the call to ensure volumes had been replaced by duplicated call to network at some point...

cli/cmd/compose/run.go
```diff
@@ -72,17 +73,30 @@ func runRun(ctx context.Context, opts runOptions) error {
...
-		return c.ComposeService().CreateOneOffContainer(ctx, project, compose.RunOptions{
-			Name:    opts.Name,
-			Command: opts.Command,
-		})
+		for _, service := range originalServices {
+			if service.Name != opts.Name {
+				dependencies = append(dependencies, service)
+			}
+		}
+		project.Services = types.Services(dependencies)
```
ndeloof, line 84: `append` magic allows to declare `dependencies` variable with type `types.Services`

ndeloof (approved):

PR #1005, opened 2020-12-03
Implement `docker compose pull`

### LINKED ISSUE #930: Local compose: pull command
Using compose command in local context, I should be able to run `docker compose pull` and pull images in my compose file, like in `docker-compose`.

Other backends (cloud backends) will return not implemented on this command.

rumpl:
**What I did**

Implemented `docker compose pull`

pull

**Related issue**
Fixes #930 

image

### DIFF
--- a/api/client/compose.go
+++ b/api/client/compose.go
@@ -37,32 +37,30 @@ func (c *composeService) Push(ctx context.Context, project *types.Project) error
 	return errdefs.ErrNotImplemented
 }
 
-// Up executes the equivalent to a `compose up`
+func (c *composeService) Pull(ctx context.Context, project *types.Project) error {
+	return errdefs.ErrNotImplemented
+}
+
 func (c *composeService) Up(context.Context, *types.Project, bool) error {
 	return errdefs.ErrNotImplemented
 }
 
-// Down executes the equivalent to a `compose down`
 func (c *composeService) Down(context.Context, string) error {
 	return errdefs.ErrNotImplemented
 }
 
-// Logs executes the equivalent to a `compose logs`
 func (c *composeService) Logs(context.Context, string, io.Writer) error {
 	return errdefs.ErrNotImplemented
 }
 
-// Ps executes the equivalent to a `compose ps`
 func (c *composeService) Ps(context.Context, string) ([]compose.ServiceStatus, error) {
 	return nil, errdefs.ErrNotImplemented
 }
 
-// List executes the equivalent to a `docker stack ls`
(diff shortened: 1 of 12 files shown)


api/client/compose.go
```diff
@@ -37,32 +37,30 @@ func (c *composeService) Push(ctx context.Context, project *types.Project) error
...
-// Up executes the equivalent to a `compose up`
+func (c *composeService) Pull(ctx context.Context, project *types.Project) error {
+	return errdefs.ErrNotImplemented
+}
+
 func (c *composeService) Up(context.Context, *types.Project, bool) error {
 	return errdefs.ErrNotImplemented
 }
 
-// Down executes the equivalent to a `compose down`
```
ulyssessouza, line 45: Linter will not like it

cli/cmd/compose/compose.go
```diff
@@ -93,7 +93,10 @@ func Command(contextType string) *cobra.Command {
 	if contextType == store.LocalContextType {
 		command.AddCommand(
 			buildCommand(),
-			pushCommand())
+			pushCommand(),
+			pullCommand(),
+		)
+
```
ulyssessouza, line 99: nit: empty line

local/compose.go
```diff
@@ -136,14 +137,135 @@ func (s *composeService) Push(ctx context.Context, project *types.Project) error
...
+
+	if info.IndexServerAddress == "" {
+		info.IndexServerAddress = registry.IndexServer
+	}
+
+	w := progress.ContextWriter(ctx)
+	eg, ctx := errgroup.WithContext(ctx)
+
+	for _, service := range project.Services {
+		service := service
```
ulyssessouza, line 166:
```suggestion
```

rumpl: It's needed because the body of the for loop is a goroutine, this line creates a new binding to the service variable so that each goroutine gets it's own `service`, without this all the goroutines would get the same `service`

api/client/compose.go
rumpl: He will because the doc comes from the interface and not its implementation

local/compose.go
gtardif: nit: in these cases I try to set a different variable name, that makes it a bit more obvious why we need a separate variable, rather than shadow the variable name.

gtardif (approved):
Very nice display !

local/compose.go
rumpl: Legit, will change

Issue #1007, opened 2020-12-03
Local compose mount error when not running compose up from compose folder directly


gtardif:
Trying to run awesome-compose ELK stack, it starts volumes properly if I simply run `docker compose up` from the ELK folder. If I run trom a parent folder, with `-f somefolder/docker-compose.yml` or `--workdir` flag, I get: 
```
$ ./bin/docker --context localctx compose up --workdir tests/composefiles/elasticsearch-logstash-kibana/
[+] Running 2/3
 ⠿ Service "elasticsearch"  Created                                                                                                                   0.6s
 ⠿ Service "kibana"         Created                                                                                                                   0.7s
 ⠇ Service "logstash"       Creating                                                                                                                  0.7s
Error response from daemon: invalid mount config for type "bind": invalid mount path: 'tests/composefiles/elasticsearch-logstash-kibana/tests/composefiles/elasticsearch-logstash-kibana/logstash/pipeline/logstash-nginx.config' mount path must be absolute
```
`./...` in compose volumes is not resolved the same way depending on how we start `compose up`. 
`docker-compose` resolves the relative path properly when using `-f somefolder/docker-compose.yml`

gtardif:
Actually using the `-f` option works fine with both `docker-compose up` and `docker compose up`, it's only with `docker compose up --workdir ...` that the volume is not mounted on the correct path. 
docker-compose `--project-directory` does not seem to be the same option (it requires -f or a compose file in the current folder, not the folder specified with the option), so this does not seem to be a regression compared to `docker-compose`

PR #1009, opened 2020-12-04
Compose fix relative volume

gtardif:
**What I did**
* avoid double appending of working dir when setting up volumes, instead resolve absolute path if required
* Added unit test & first compose volume e2e test

**Related issue**
Fixes https://github.com/docker/compose-cli/issues/1007


ndeloof (approved):

gtardif:
rebased and fixed e2e tests now requiring `compose up -d` :)

PR #1010, opened 2020-12-04
Give progress its own context

### LINKED ISSUE #992: Local compose: allow ctrl+C during image build
when running `docker compose up`, we can't ctrl+C the command during image build. (build freezes, never returns)

rumpl:
**What I did**

Fixed ctrl+c on build

The build and progress can't share the same context, the progress context
needs to live longer for the progress to be able to read all the messages
from build.

**Related issue**

Fixes #992 

image

### DIFF
--- a/local/build.go
+++ b/local/build.go
@@ -103,8 +103,15 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opts
 			Driver: d,
 		},
 	}
+
+	// Progress needs its own context that lives longer than the
+	// build one otherwise it won't read all the messages from
+	// build and will lock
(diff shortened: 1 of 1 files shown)


gtardif (approved):

local/build.go
```diff
@@ -103,8 +103,11 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opts
 			Driver: d,
 		},
 	}
+	ctx2, cancel := context.WithCancel(context.TODO())
```
gtardif, line 106: nit: maybe a more explicit context name or a full comment to explain why we can't share the context here

rumpl: Done

Issue #940, opened 2020-11-20
Local compose: run command


gtardif:
Using compose command in local context, I should be able to run `docker compose run` and run a specific command in a new container started from one service spec, like in `docker-compose`

gtardif:
Will need to update oneoff label for "compose run" containers, at https://github.com/docker/compose-cli/blob/main/local/compose.go#L371

ndeloof:
Implementation note: new `compose run` command could tweak the compose model to force target service with: 

* scale=1 
* command, env, etc as set by command
* create random slug 
* set label com.docker.compose.slug for target service
* for container name = project_service_run_slug

then 
* use the existing `Create` API to create containers for the stack 
* introduce a new `Attach` API to attach to container in/ou streams
* Start the stack without attaching to other services output

PR #1015, opened 2020-12-07
Revisit logs/up API to pass a LogConsumer vs io.Writer

ndeloof:
**What I did**
Refactored the compose API to use a `LogConsumer` callback vs a raw `io.Writer`. Log message then get typed by service/container and can be processed by consumer for output (raw, colored, prefixed, filtered ...)

This refactoring started with the intent to introduce `attachContainerStreams` so we can define `compose.Attach` API method that will be required to implement `compose run`, see https://github.com/docker/compose-cli/issues/940#issuecomment-739770823

/test-ecs

image

### DIFF
--- a/local/compose.go
+++ b/local/compose.go
@@ -19,6 +19,7 @@
 package local
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/json"
@@ -54,7 +55,6 @@ import (
 	"github.com/docker/compose-cli/api/compose"
 	"github.com/docker/compose-cli/config"
 	errdefs2 "github.com/docker/compose-cli/errdefs"
-	"github.com/docker/compose-cli/formatter"
 	"github.com/docker/compose-cli/progress"
 )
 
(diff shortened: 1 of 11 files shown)


local/compose.go
```diff
@@ -391,44 +390,59 @@ func (s *composeService) attach(ctx context.Context, project *types.Project, w i
...
 	if err != nil {
 		return err
 	}
 
-	reader, err := s.getContainerStdout(ctx, container)
+	return s.attachContainerStreams(ctx, container, service.Tty, nil, w)
+}
+
+func (s *composeService) attachContainerStreams(ctx context.Context, container moby.Container, tty bool, r io.Reader, w io.Writer) error {
+	stdin, stdout, err := s.getContainerStdout(ctx, container)
```
gtardif, line 406: Nit: rename this getContainerStdinStdout

local/compose.go
```diff
@@ -804,7 +838,7 @@ func getContainerCreateOptions(p *types.Project, s types.ServiceConfig, number i
 		StopTimeout: toSeconds(s.StopGracePeriod),
 	}
 
-	mountOptions, err := buildContainerMountOptions(p, s, inherit)
+	mountOptions, err := buildContainerMountOptions(s, inherit)
```
gtardif, line 841: Thx! (not sure why the CI linter didn't see this unused code)

gtardif (approved):
LGTM, small nit

PR #1018, opened 2020-12-07
`compose up` and other compose commands running on “Moby” context type.

gtardif:

**What I did**
* in compose command, do not warn command is not available on default context type, rather use "local" backend when on "moby" context type.
* change e2e tests to not require a local context.
* local backend is not behind a feature flag anymore

**Related issue**
https://github.com/docker/dev-tooling-team/issues/244

### DIFF
--- a/cli/cmd/compose/build.go
+++ b/cli/cmd/compose/build.go
@@ -45,7 +45,7 @@ func buildCommand() *cobra.Command {
 }
 
 func runBuild(ctx context.Context, opts buildOptions, services []string) error {
-	c, err := client.New(ctx)
+	c, err := client.NewWithDefaultLocalBackend(ctx)
 	if err != nil {
 		return err
 	}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,12 +60,12 @@ jobs:
 
       - name: Test
         env:
-          BUILD_TAGS: example,local
+          BUILD_TAGS: example
         run: make -f builder.Makefile test
 
       - name: Build for local E2E
         env:
-          BUILD_TAGS: example,local,e2e
+          BUILD_TAGS: example,e2e
         run: make -f builder.Makefile cli
 
       - name: E2E Test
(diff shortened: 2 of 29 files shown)


gtardif:
I wish I could do something more centralised in the compose command, in pre-run. But ctx being immutable, I can't change the context to set the client or something similar here, I can't either change the cmd to replace the context.

gtardif:
We can look at working on contexts with remote engine in a separate PR

gtardif:
Also removed build directive: local backend is not behind a feature flag anymore

cli/cmd/compose/build.go
```diff
@@ -45,7 +45,7 @@ func buildCommand() *cobra.Command {
 }
 
 func runBuild(ctx context.Context, opts buildOptions, services []string) error {
-	c, err := client.New(ctx)
+	c, err := client.NewWithDefaultLocalBackend(ctx)
```
rumpl, line 48: I would prefer if we had something like

```go
c, err := client.New(ctx, client.WithContextType(store.LocalContextType)
```

Or something in that vein

gtardif: I didn't want to change the existing API ` client.New(ctx)` all over the place, so I need a function with a different name. We can do `c, err := client.NewWithDefaultBackend(ctx, store.LocalContextType)` in all compose commands. 

(I had this initially, then thought we are not likely to use this with any other backend type, so I published only `client.NewWithDefaultLocalBackend(ctx)`, but passing the `store.LocalContextType` makes it a bit more explicit)

gtardif: Maybe I missed your point, let me quickly check what I can do in the direction of `client.New(ctx, client.WithContextType(store.LocalContextType)`...

gtardif: OK, I missed your point with optional param `...opts ClientOpts`. 
For now, we have only this default backend option, I prefer the explicit call with a known list of typed params. 
If we see we need more options to the client creation, we can change for `client.New(ctx, ... opts ClientOpt)`

ndeloof (approved):

PR #1024, opened 2020-12-07
Fix volume ensure when creating local volume before running services

gtardif:
**What I did**
* when creating volumes, don't return previous error of the check to see if volume exists ...
* Add volume creation in compose e2e volume test

**Related issue**
Found in several awesome compose examples

### DIFF
--- a/local/compose.go
+++ b/local/compose.go
@@ -1086,24 +1086,24 @@ func (s *composeService) ensureVolume(ctx context.Context, volume types.VolumeCo
 	// TODO could identify volume by label vs name
 	_, err := s.apiClient.VolumeInspect(ctx, volume.Name)
 	if err != nil {
-		if errdefs.IsNotFound(err) {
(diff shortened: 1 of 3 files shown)


local/compose.go
```diff
@@ -1086,24 +1086,24 @@ func (s *composeService) ensureVolume(ctx context.Context, volume types.VolumeCo
...
-				Name:       volume.Name,
-				Driver:     volume.Driver,
-				DriverOpts: volume.DriverOpts,
-			})
-			if err != nil {
-				w.Event(progress.ErrorEvent(eventName))
-				return err
-			}
-			w.Event(progress.CreatedEvent(eventName))
+		if !errdefs.IsNotFound(err) {
```
rumpl, line 1089: I'm not sure I understand what changed in the code. I only see that this condition changed but we are returning the same `err` no?

ndeloof (approved):

rumpl (approved):

PR #1030, opened 2020-12-08
compose: Add service labels

### LINKED ISSUE #1029: Local compose: support service labels in compose files
Used by https://github.com/docker/awesome-compose/tree/master/traefik-golang

rumpl:
**What I did**

Added labels defined in the compose file to the created containers

Fixes #1029


local/compose/create.go
```diff
@@ -94,6 +94,10 @@ func getContainerCreateOptions(p *types.Project, s types.ServiceConfig, number i
 		containerNumberLabel: strconv.Itoa(number),
 	}
 
+	for k, v := range s.Labels {
```
ndeloof, line 97: doing so, compose file can override `com.docker.compose.xx` labels. Sounds crazy, but docker-compose prevent this. I can't imagine any use-case to require this, but users are full of creativity.

ndeloof (approved):

local/compose/create.go
rumpl: Oh ok, so `docker-compose` silently rewrites the label, ok will fix

ndeloof (approved):

PR #1036, opened 2020-12-09
Don't share the compose file with ACI

rumpl:
**What I did**
Fixed the compose ACI test, ACI doesn't support service labels, we need to split

/test-aci


tests/compose-e2e/compose_test.go
```diff
@@ -49,15 +49,8 @@ func TestLocalComposeUp(t *testing.T) {
 
 	const projectName = "compose-e2e-demo"
 
-	t.Run("build", func(t *testing.T) {
```
rumpl, line 52: Note to reviewer: removed for now, we will add tests for build later

gtardif: OK, if it's really in the way ; could we not keep this until we replace it by more proper test on build ? 
(just I prefer to not reduce e2e coverage, as we don't have too much there ATM)

gtardif: could we just point to the /sentences compose files and make this pass ?

rumpl: The build fails if I point it to the sentences compose file because it can't find anything to build...

I'll create a new compose file with build, later

gtardif (approved):

Issue #974, opened 2020-11-27
ACI : Fix flaky test


gtardif:
This happens from time to time (seen at least twice). 
https://github.com/docker/compose-cli/runs/1460186738

=== CONT  TestContainerRunAttached
Error:     e2e.go:174: assertion failed: 
        Command:  /tmp/123959829/docker prune --force
        ExitCode: 1
        Error:    exit status 1
        Stdout:   
        Stderr:   1 error occurred:
        	* pollingTrackerBase#pollForStatus: failed to send HTTP request: StatusCode=0 -- Original Error: Get "": unsupported protocol scheme ""

gtardif:
While looking at this, I coud not reproduce this easily to check what's wrong. 
Looking at the code, it seems to be failing in https://github.com/docker/compose-cli/blob/main/aci/resources.go#L57, when the SDK waits for the result of the delete call and polls result on the Future. 
When trying to reproduce this, I had another issue in prune when deleting contains up for a few days: 
```
cannot delete container group: containerinstance.ContainerGroupsClient#Delete: Failure sending request: StatusCode=200 -- Original Error: Code="Failed" Message="The async operation failed." AdditionalInfo=[{"id":"/subscriptions/5dfb2c88-af2d-47e6-96da-22ee4be68587/resourceGroups/gtardifeastus2/providers/Microsoft.ContainerInstance/containerGroups/focused-jepsen","location":"eastus2","name":"focused-jepsen","properties":{"containers":[{"name":"focused-jepsen","properties":{"environmentVariables":[],"image":"nginx","instanceView":{"currentState":{"detailStatus":"Waiting to run","state":"Waiting"},"restartCount":0},"ports":[{"port":80}],"resources":{"limits":{"cpu":1,"memoryInGB":1},"requests":{"cpu":1,"memoryInGB":1}}}}],"imageRegistryCredentials":[{"server":"index.docker.io","username":"gtardif"}],"initContainers":[],"instanceView":{"events":[],"state":"Running"},"ipAddress":{"ip":"52.252.64.3","ports":[{"port":80,"protocol":"TCP"}],"type":"Public"},"osType":"Linux","provisioningState":"Failed","restartPolicy":"Never","sku":"Standard"},"tags":{"docker-single-container":"docker-single-container"},"type":"Microsoft.ContainerInstance/containerGroups"}]
```

This happened on many containers (maybe 1/2 or 1/3)
[...]

PR #1037, opened 2020-12-09
Trying to fix ACI flakyness on docker prune --force

gtardif:

**What I did**
Wait 1 sec and retry on ACI `docker prune --force` when we hit this specific SDK error

**Related issue**
Fixes https://github.com/docker/compose-cli/issues/974

/test-aci


gtardif:
Ah, ACI test failing because this is from my fork :/ 
Anyway, the change is not likely to be hit, as the flaky test did not fail that often... We'll only have feedback after a while running this on main

ndeloof (approved):

PR #1047, opened 2020-12-11
Check if stdin is nil before closing

rumpl:
**What I did**

Check if stdin is nil before closing, `getContainerStreams` returns a `nil` `stdin` if the container is already running


gtardif (approved):

glours (approved):
LTGM

PR #1052, opened 2020-12-11
Add test for compose pull

rumpl:
**What I did**

Added test for compose pull


gtardif (commented):
why all the other yaml files, in addition to docker-compose.yaml ? I guess fo more tests, coming in other PRs ? (or missing tests in this one ?)

rumpl:
I copied the whole directory from `docker/compose`, there will be other tests with those files, I removed them from here, it's better you are right

gtardif (approved):

PR #1057, opened 2020-12-14
Compose unit test networks + image names

gtardif:
**What I did**
* added unit tests on image names & removed duplicate name compute
* added unit tests on network labels and names
* found bug in networks with given names
* added e2e test on network

There are still cases not fixed here, due to compose-go setting network names to map keys by default. 
We need to do all the name setting in one place, compose-go or compose-cli. Still, fixing what I can here and adding e2e test will ensure we don't regress.

**Related issue**
https://github.com/docker/dev-tooling-team/issues/257

### DIFF
--- a/ecs/awsResources.go
+++ b/ecs/awsResources.go
@@ -388,7 +388,7 @@ func (b *ecsAPIService) ensureVolumes(r *awsResources, project *types.Project, t
 				},
 				{
 					Key:   "Name",
-					Value: fmt.Sprintf("%s_%s", project.Name, name),
+					Value: volume.Name,
 				},
 			},
 			KmsKeyId:                        kmsKeyID,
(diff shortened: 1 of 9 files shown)


gtardif:
> We need to do all the name setting in one place, compose-go or compose-cli.
Same question happens for volumes

Setting these names in compose-go will help various implementations to have the same defaults and behaviour for names, and remove code from compose-cli. 
On the other side, if we want to keep all information about what users typed in compose files, and only process compose files info down the road, compose-go should then let empty values as defined in compose files.

ndeloof (approved):

gtardif:
see with @ndeloof , will move all the name settings to compose-spec/compose-co, that way we'll ensure the default names etc. do not depend on implementation. Will bump compose-go in this PR when ready

gtardif:
Closing in favor of #1136

PR #1062, opened 2020-12-15
Make Moby command detection (based on help output) more robust.

gtardif:
It seems some tabs have been replaced by spaces in CLI 20.10.

We need to install docker CLI ourselves in Github actions nodes, local e2e was failing locally, I assume this is since CLI 20.10.

symptom: `docker --context aci images` was returning 
`unknown command "images" for "docker"` 
instead of 
`Command "images" not available in current context (test-example), you can use the "default" context to run this command`

**What I did**
* replace exact tab matching by regex matching any space

**Related issue**

### DIFF
--- a/cli/mobycli/exec.go
+++ b/cli/mobycli/exec.go
@@ -22,7 +22,7 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
-	"strings"
+	"regexp"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


gtardif:
See CI ticket https://github.com/docker/compose-cli/issues/1063

gtardif:
already covered with e2e test TestUnsupportedCommand ; needs #1063 to update GHA nodes

ndeloof (approved):

Issue #921, opened 2020-11-17
A load balancer cannot be attached to multiple subnets in the same AvailabilityZone


shawnhind:
**Description**
Our VPC has multiple subnets in the same AZs.  When running docker compose up using this VPC specified it gives this error:
A load balancer cannot be attached to multiple subnets in the same Availability Zone (Service: AmazonElasticLoadBalancing; Status Code: 400; Error Code: InvalidConfigurationRequest

**Steps to reproduce the issue:**
1. Create a VPC with multiple subnets in the same VPC
2. Set x-aws-vpc to that VPC in your compose file
3. Run docker compose up

**Describe the results you received:**
Error message (above)

**Describe the results you expected:**
Should have successfully created the load balancer.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
Client:
 Cloud integration: 1.0.2
 Version:           19.03.12
 API version:       1.40
 Go version:        go1.13.15
 Git commit:        48a66213fe17
 Built:             Mon Aug  3 00:00:00 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          19.03.12
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       48a66213fe17
  Built:            Mon Aug  3 00:00:00 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.1.5_catatonit
  GitCommit:        

```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```
[
    {
        "Name": "aws",
        "Metadata": {
            "Description": "(us-east-1)",
            "Type": "ecs"
        },
        "Endpoints": {
            "docker": {
                "SkipTLSVerify": false
            },
            "ecs": {
                "Profile": "default"
            }
        },
        "TLSMaterial": {},
        "Storage": {
            "MetadataPath": 
            "TLSPath": 
        }
    }
]

```

**Output of `docker info`:**

```
Client:
 Debug Mode: false

Server:
 Containers: 13
  Running: 0
  Paused: 0
  Stopped: 13
[...]

PR #1064, opened 2020-12-15
Only consider public subnets

ndeloof:
**What I did**
Add ability to detect "public subnets" and only consider those when building the stack's network configuration.

Note: private subnets (i.e. not associated with an Internet Gateway) don't make sense for Docker ECS integration as service would not be able to pull sidecar images from DockerHub.

**Related issue**
https://github.com/docker/compose-cli/issues/921

/test-ecs

image


aiordache (approved):
LGTM!

dalegaspi:
i suppose this is not addressed as of v20.10.5 because it's still referencing the private subnets in the resulting CF template. :-(

as of right now i'm just doing a `docker compose convert` then comment out the private subnets.

komatom:
in which version of docker cli is this?

Issue #928, opened 2020-11-20
Use `local` backend code to implement ecs-local


ndeloof:
**Description**
ecs-local ("simulation mode") context type require docker-compose on PATH and just generate compose file with additional aws-endpoints container/network.
With the introduction of compose support in `local` backend, we can rely on this to run the compose stack in simulation mode without using an external command.

ndeloof:
Blocked by https://github.com/docker/compose-cli/issues/1070

Issue #1070, opened 2020-12-16
Detect and manage orphaned containers


ndeloof:
docker-compose can detect container labelled for service but not defined in compose file and manage them to cleanup the environment.
This results in a warning like:
```
WARNING: Found orphan containers (tutu_ecs-local-endpoints_1) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up.
```
For ECS-local context we inject a simulation container and network to the compose model. Without orphaned containers management `compose down` fails as we try to remove network while container is still running.

PR #1071, opened 2020-12-16
use compose-in-go code to implement local ECS simulation context

ndeloof:
**What I did**
Replaced shell-out to `docker-compose` with delegation to local backend compose implementation

This also introduces `removeOrphaned` option on compose.Down api, not explosed to the CLI yet

**Related issue**
close https://github.com/docker/compose-cli/issues/928
https://github.com/docker/compose-cli/issues/1070

image

### DIFF
--- a/api/compose/api.go
+++ b/api/compose/api.go
@@ -37,7 +37,7 @@ type Service interface {
 	// Up executes the equivalent to a `compose up`
 	Up(ctx context.Context, project *types.Project, detach bool) error
 	// Down executes the equivalent to a `compose down`
-	Down(ctx context.Context, projectName string) error
+	Down(ctx context.Context, projectName string, removeOrphans bool) error
 	// Logs executes the equivalent to a `compose logs`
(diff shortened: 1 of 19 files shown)


api/compose/api.go
```diff
@@ -37,7 +37,7 @@ type Service interface {
 	// Up executes the equivalent to a `compose up`
 	Up(ctx context.Context, project *types.Project, detach bool) error
 	// Down executes the equivalent to a `compose down`
-	Down(ctx context.Context, projectName string) error
+	Down(ctx context.Context, projectName string, removeOprhans bool) error
```
rumpl, line 40: Could this be an object rather than a boolean?

ndeloof: I don't get your point here. Would you prefer of a `DownOptions` struct to hold this bool?

rumpl: Yes exactly

gtardif (approved):

Issue #1074, opened 2020-12-16
The submitted information didn't contain changes. Submit different information to create a change set.


SEFI2:
**Description**

I am getting this message, even though I have changes and want to deploy them:
`The submitted information didn't contain changes. Submit different information to create a change set.`

**Steps to reproduce the issue:**
1. docker-compose up --context myecscontext

**Describe the results you received:**
As I understood, compose-cli cannot create a changeset after creating a template, but couldn't get where exactly it fails.

**Describe the results you expected:**
I had a lot of things changed: `code`, `Dockerfiles`, and `docker-compose`, but still it shows this information.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
docker-compose version 1.27.4, build 40524192
```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```
[
    {
        "Name": "myecscontext",
        "Metadata": {
            "Type": "ecs"
        },
        "Endpoints": {
            "docker": {
                "SkipTLSVerify": false
            },
            "ecs": {
                "Profile": "default"
            }
        },
        "TLSMaterial": {},
        "Storage": {
            "MetadataPath": "/Users/kadyrbeknarmamatov/.docker/contexts/meta/8bb20fb47c4248774ab660063891d2332facb5997d914f61b3bdfeb471eaddba",
            "TLSPath": "/Users/kadyrbeknarmamatov/.docker/contexts/tls/8bb20fb47c4248774ab660063891d2332facb5997d914f61b3bdfeb471eaddba"
        }
    }
]
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.4.2-docker)
  scan: Docker Scan (Docker Inc., v0.5.0)

Server:
 Containers: 8
  Running: 0
  Paused: 0
  Stopped: 8
 Images: 74
 Server Version: 20.10.0
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
[...]

PR #1075, opened 2020-12-17
create Services with image digests when known

ndeloof:
**What I did**
Update image in service definition with RepoDigests when known by local engine and non ambiguous.
This "quick fix" prevent `compose up` to fail with "_no change detected_" for the expected scenario a user rebuilds a service image then push and update his app on ECS.

**Related issue**
https://github.com/docker/compose-cli/issues/1074

/test-ecs

### DIFF
--- a/ecs/backend.go
+++ b/ecs/backend.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/docker/docker/client"
 )
 
 const backendType = store.EcsContextType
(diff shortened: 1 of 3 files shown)


thaJeztah:
Should this be used only when `--force-recreate` is set? (perhaps we need a `--resolve-image` flag at some point, similar to that on stack deploy)

ndeloof:
we can't really implement `--force-recreate` on ECS using CloudFormation, as the service state is managed by CloudFormation and we can't force an artificial change to get service updated.

docker-compose up do detect the running containers don't run the configured image (see https://github.com/docker/compose/blob/5e3708e6051dc13c4902d028044c6ab3f1482406/compose/service.py#L377, the config_hash contains the image ID), this is what I'm trying to reproduce here. I don't really care about the swarm way :P

chris-crone (commented):
I don't think this'll make things worse but if I understand correctly, it will only work if you have a local copy of a pushed image?

ndeloof:
@chris-crone yes indeed, this is really only to address the usage scenario "rebuild, push, up"
updates to image pushed by third party (typically: CI) won't be detected.
We _could_ introduce the same logic used by docker-app to make this a stronger solution in a future PR.

gtardif:
Agreed, it does not cover all cases but does significantly improve the local push scenario, and I don't see any regression otherwise. We can probably do the same on ACI, failing to update stacks for the same reason (to be done in a separate PR).

thaJeztah:
So it depends a bit on how we position compose; For `docker-compose` running on a local engine, it can be expected that it's a single developer, "inner loop" situation.

When doing the same on a cloud provider, chances are that that developer is not the only person pushing these images; taking the local image cache as source of truth may not be a good choice for that (another person on the team could've pushed an image, and now you're downgrading the image to an older version). For sure, this would only occur if no explicit tag was specified / rolling tag (using `:latest` should not be "best practice here"). Still, using the registry as source of truth would likely be better.

That said; compose could be used to _deploy_ a stack. In that case, you likely want to have more control over "when" a new version of the image should be deployed (even when using `:latest`), which is the `--resolve-image` flag was added on `docker stack deploy` (and a `--force` flag on `docker service update`).

ndeloof:
closing to focus on https://github.com/docker/compose-cli/pull/1120

PR #1076, opened 2020-12-17
Fix metrics flakyness

### LINKED ISSUE #1033: Flaky test on metrics gathering
Probably due to some timing, I've seen this a couple of times already: 
```
http server started on /tmp/195461985/docker-cli.sock
	[TestContextMetrics] docker --help
	[TestContextMetrics] docker run --help
Error:     e2e_test.go:155: assertion failed: 
        --- ←
        +++ usage
          []string{
        - 	`{"command":"help run","context":"moby","source":"cli","status":"success"}`,
          	`{"command":"--help","context":"moby","source":"cli","status":"su`...,
          	`{"command":"--help run","context":"moby","source":"cli","status"`...,
          }
```
https://github.com/docker/compose-cli/pull/1032/checks?check_run_id=1518710270

gtardif:
**What I did**
* cleanup previous fix (#1038) that didn't correct the root cause
* realised the metrics mock server was started totally async, need to wait a bit
* could not find an easy API to check the server is ready, wait for 500 ms, other solutions would loop and wait anyway... 

**Related issue**
Fixes #1033 

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -151,7 +151,7 @@ func TestContextMetrics(t *testing.T) {
 		c.RunDockerCmd("--help")
 		c.RunDockerCmd("run", "--help")
 
-		usage := s.GetUsage(3)
+		usage := s.GetUsage()
 		assert.DeepEqual(t, []string{
 			`{"command":"help run","context":"moby","source":"cli","status":"success"}`,
 			`{"command":"--help","context":"moby","source":"cli","status":"success"}`,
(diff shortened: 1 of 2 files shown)


gtardif:
Actually it seems it does not fully work (also debugging this in a loop: https://github.com/gtardif/compose-cli/runs/1571246788?check_suite_focus=true). I'll add a proper wait loop checking we receive a first metric

gtardif:
And running this in loop, green (https://github.com/gtardif/compose-cli/runs/1571545817?check_suite_focus=true ; needed to run all E2E tests to reproduce the issue, not only the metrics one)

tests/e2e/e2e_test.go
```diff
@@ -144,14 +144,26 @@ func TestContextMetrics(t *testing.T) {
 	s.Start()
 	defer s.Stop()
 
+	started := false
+	for i := 0; i < 30; i++ {
```
ndeloof, line 148: Let me play the Gopher pursist for once :P

`NewMetricsServer.Start` could return a `chan bool` and send `true` when ready, then test can wait on `<- start`

LGTM otherwise

ndeloof (approved):

tests/e2e/e2e_test.go
gtardif: yes, but then we need "c" to run commands in Start... (Start could also just wait before returning)

Issue #1046, opened 2020-12-10
docker compose (experimental): container names 


EdSF:
Running the experimental `docker compose` does not use the defined `container_name` in docker-compose.yml (while running `docker-compose up` does.

Note: I actually "stumbled" on this by forgetting the dash `-` (while not in Azure CLI context). 

Having said that, when using the Azure CLI context, which only supports `docker compose` (no dash), this occurs there too. If you have network resources that depend on `container_name` for connections (re: between containers), these will fail (names defined in compose file aren't used).

**Steps to reproduce the issue:**
1. Define `container_name` for your services in `docker-compose.yml`
2. run the new/experimental `docker compose` (no dash)
3. Everything succeeds, but the container names _looks like_ some combination of `services` names + numeric suffix

**Describe the results you received:**

Container names look "generated" from `services` names defined in docker-compose.yml with a numeric suffix

**Describe the results you expected:**

Container names as defined in `container-name` key in compose file - in other words, just like what you'd expect running `docker-compose up`

gtardif:
Thanks for the feedback, indeed we know we don't have full support for all compose features yet, we'll prioritize this in our next improvements in local compose. 
... And also try to match this in ACI as you mentioned, this is not available in ACI either

EdSF:
> Thanks for the feedback, indeed we know we don't have full support for all compose features yet, we'll prioritize this in our next improvements in local compose.
> ... And also try to match this in ACI as you mentioned, this is not available in ACI either

Thank you! 

I was in an Azure webinar a couple of days ago and I had this question for them (ACI/ACR) because I couldn't figure out the container naming (and thought it was an `Azure CLI` issue) while trying things out based on [Tutorial: Deploy a multi-container group using Docker Compose](https://docs.microsoft.com/en-us/azure/container-instances/tutorial-docker-compose)(and experienced the network connection issue..)

ndeloof:
AFAIK @aiordache fixed this as part of her implementation of `compose run`

PR #1079, opened 2020-12-23
Use `container_name` property on service

ulyssessouza:
**What I did**
This applies `container_name` when available in the service definition.

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/1046

### DIFF
--- a/local/compose/convergence.go
+++ b/local/compose/convergence.go
@@ -26,6 +26,7 @@ import (
 	moby "github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/api/types/network"
+	"github.com/docker/docker/client"
 	"golang.org/x/sync/errgroup"
 
 	status "github.com/docker/compose-cli/local/moby"
@@ -35,11 +36,25 @@ import (
 const (
 	extLifecycle  = "x-lifecycle"
 	forceRecreate = "force_recreate"
+
+	doubledContainerNameWarning = "WARNING: The %q service is using the custom container name %q. " +
+		"Docker requires each container to have a unique name. " +
+		"Remove the custom name to scale the service.\n"
 )
+
+func containerExists(ctx context.Context, c *client.Client, name string) bool {
+	container, err := c.ContainerInspect(ctx, name)
+	return err == nil && container.ContainerJSONBase != nil && container.Name == "/"+name
+}
 
 func (s *composeService) ensureService(ctx context.Context, observedState Containers, project *types.Project, service types.ServiceConfig) error {
 	scale := getScale(service)
 	actual := observedState.filter(isService(service.Name))
+	if scale > 1 && service.ContainerName != "" {
+		return fmt.Errorf(doubledContainerNameWarning,
+			service.Name,
+			service.ContainerName)
+	}
 
 	eg, _ := errgroup.WithContext(ctx)
 	if len(actual) < scale {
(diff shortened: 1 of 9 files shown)


aci/convert/convert.go
```diff
@@ -181,9 +181,12 @@ func (s serviceConfigAciHelper) getAciContainer() (containerinstance.Container,
 	if err != nil {
 		return containerinstance.Container{}, err
 	}
-
+	containerName := s.Name
```
gtardif, line 184: I don't think we can easily support container_name in ACI, as we use the _ naming convention to retrieve containers from a compose stack. This would break other scenarios like `docker ps` => `docker logs` etc.

gtardif: in local docker-compose, container_name means container has no `projectname_` prefix. I don't think we should support it in ACI with a different meaning, that would be confusing more than anything else. We could just ignore it with a warning

local/compose/convergence.go
```diff
@@ -60,7 +60,7 @@ func (s *composeService) ensureService(ctx context.Context, project *types.Proje
 		missing := scale - len(actual)
 		for i := 0; i < missing; i++ {
 			number := next + i
-			name := fmt.Sprintf("%s_%s_%d", project.Name, service.Name, number)
+			name := getContainerNamePrintLabel(project.Name, service, number)
```
gtardif, line 63: Need to validate container_name is not used with more than one container for a service

local/compose/convergence.go
```diff
@@ -114,6 +114,14 @@ func (s *composeService) ensureService(ctx context.Context, project *types.Proje
 	return eg.Wait()
 }
 
+func getContainerNamePrintLabel(projectName string, service types.ServiceConfig, number int) string {
```
gtardif, line 117: nit : what's the meaning of this function name ? not clear to me

gtardif (changes requested):
Mainly remove support on ACI, seems more confusing than bringing value. 
container_name also has the side effect of adding a network alias for the service, we should implement this in local compose.

ndeloof (approved):

gtardif (commented):
container name is also used as a network alias, we need to add this when setting the container network config as a new network alias, in addition to the service name. 
(awesome compose ELK sample uses this network alias)
This could be done in a separate PR if it helps, or add a separate commit to this PR

tests/e2e/e2e_test.go
```diff
@@ -587,3 +587,21 @@ func TestFailOnEcsUsageAsPlugin(t *testing.T) {
...
+				"Docker requires each container to have a unique name. " +
+				"Remove the custom name to scale the service.",
+		})
+	})
+
+	t.Run("compose up container_name with scale == 1", func(t *testing.T) {
+		c.RunDockerCmd("-f", "../composefiles/container_name/docker-compose-good.yml", "compose", "up", "-d")
+		c.RunDockerCmd("-f", "../composefiles/container_name/docker-compose-good.yml", "compose", "down")
+	})
+}
```
gtardif, line 607: Maybe rather add a unit test on ensureService() in convergence.go for the failure case, E2E tests will become slow at some point...

gtardif (approved):
LGTM, with unit test (just for the failure message), rather than an E2E test

tests/e2e/e2e_test.go
ulyssessouza: PTAL

gtardif:
Fixed linter

local/compose/convergence.go
```diff
@@ -143,17 +182,30 @@ func nextContainerNumber(containers []moby.Container) (int, error) {
...
+		err = fmt.Errorf(doubledContainerNameWarning,
+			config.Name,
+			config.ContainerName)
 	}
-	return 1
+	return scale, err
 }
 
 func (s *composeService) createContainer(ctx context.Context, project *types.Project, service types.ServiceConfig, name string, number int, autoRemove bool) error {
+	if containerExists(ctx, s.apiClient, name) {
```
gtardif, line 204: This will add an inspect api call for every container we create... we should be able to do better

PR #1080, opened 2020-12-24
Build tidy

chris-crone:
**What I did**
* Used loops and logic for cross compiles rather than hard coding each platform
* Added the ability to cross compile a single platform
* Added builds for linux/arm

This should make it really easy to add darwin/arm64 when Go 1.16 is GA

Manually tested that the tarballs generated by this contain what we're expecting.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -86,6 +86,18 @@ RUN --mount=target=. \
     GIT_TAG=${GIT_TAG} \
     make BINARY=/out/docker  -f builder.Makefile cross
 
+FROM base AS make-cross-single
+ARG BUILD_TAGS
+ARG GIT_TAG
+ARG TARGETOS
+ARG TARGETARCH
+RUN --mount=target=. \
+    --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    BUILD_TAGS=${BUILD_TAGS} \
+    GIT_TAG=${GIT_TAG} \
+    make BINARY=/out/docker -f builder.Makefile cross-${TARGETOS}-${TARGETARCH}
+
 FROM scratch AS protos
 COPY --from=make-protos /compose-cli/protos .
 
@@ -94,6 +106,9 @@ COPY --from=make-cli /out/* .
 
 FROM scratch AS cross
 COPY --from=make-cross /out/* .
+
+FROM scratch AS cross-single
+COPY --from=make-cross-single /out/* .
 
 FROM base AS test
 ENV CGO_ENABLED=0
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,13 @@ e2e-aci: ## Run End to end ACI tests. Set E2E_TEST=TestName to run a single test
 e2e-ecs: ## Run End to end ECS tests. Set E2E_TEST=TestName to run a single test
 	go test -timeout 20m -count=1 -v $(TEST_FLAGS) ./tests/ecs-e2e
 
+cross-%:
+	@docker build . --target cross-single \
+	--platform $(subst -,/,$*) \
+	--build-arg BUILD_TAGS \
+	--build-arg GIT_TAG=$(GIT_TAG) \
+	--output ./bin \
+
 cross: ## Compile the CLI for linux, darwin and windows
 	@docker build . --target cross \
 	--build-arg BUILD_TAGS \
--- a/builder.Makefile
+++ b/builder.Makefile
@@ -61,7 +61,7 @@ cli:
 
 cross-%:
 	GOOS=$(word 1,$(subst -, ,$(*))) GOARCH=$(word 2,$(subst -, ,$(*))) \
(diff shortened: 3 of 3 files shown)


builder.Makefile
```diff
@@ -57,11 +59,12 @@ protos:
 cli:
 	GOOS=${GOOS} GOARCH=${GOARCH} $(GO_BUILD) $(TAGS) -o $(BINARY_WITH_EXTENSION) ./cli
 
+cross-%:
+	GOOS=$(word 1,$(subst -, ,$(*))) GOARCH=$(word 2,$(subst -, ,$(*))) \
```
gtardif, line 63: I think for 5 values in a makefile, I'd prefer listing the 5 lines explicitely and avoid all the string manipulation that is hard to test, validate and is a nightmare to read :). 
It would also allow for target auto-completion and make it obvious what targets are supported, I'm happy to live with small duplication for this

builder.Makefile
```diff
@@ -57,11 +59,12 @@ protos:
...
+	$(GO_BUILD) $(TAGS) -o $(BINARY)-$(*)$(if $(filter windows,$(word 1,$(subst -, ,$(*)))),.exe,) ./cli
+
 .PHONY: cross
-cross:
-	GOOS=linux   GOARCH=amd64 $(GO_BUILD) $(TAGS) -o $(BINARY)-linux-amd64 ./cli
-	GOOS=darwin  GOARCH=amd64 $(GO_BUILD) $(TAGS) -o $(BINARY)-darwin-amd64 ./cli
-	GOOS=windows GOARCH=amd64 $(GO_BUILD) $(TAGS) -o $(BINARY)-windows-amd64.exe ./cli
+cross: $(foreach plat,$(CROSS_PLATFORMS),cross-$(subst /,-,$(plat)))
```
gtardif, line 67: same here

builder.Makefile
```diff
@@ -86,8 +89,8 @@ check-go-mod:
 .PHONY: package
 package: cross
 	mkdir -p dist
-	tar -czf dist/docker-linux-amd64.tar.gz $(TAR_TRANSFORM) packaging/LICENSE $(BINARY)-linux-amd64
-	tar -czf dist/docker-darwin-amd64.tar.gz $(TAR_TRANSFORM) packaging/LICENSE $(BINARY)-darwin-amd64
+	$(foreach plat,$(filter-out windows%,$(CROSS_PLATFORMS)), \
```
gtardif, line 92: and here

Dockerfile
```diff
@@ -86,6 +86,18 @@ RUN --mount=target=. \
...
+ARG BUILD_TAGS
+ARG GIT_TAG
+ARG TARGETOS
+ARG TARGETARCH
+RUN --mount=target=. \
+    --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    BUILD_TAGS=${BUILD_TAGS} \
+    GIT_TAG=${GIT_TAG} \
+    make BINARY=/out/docker -f builder.Makefile cross-${TARGETOS}-${TARGETARCH}
```
gtardif, line 99: could we avoid duplicating all this, and reuse the make-cross ? (adding optional ARGS for TARGETOS and TARGETARCH maybe ? )

Makefile
```diff
@@ -55,6 +55,13 @@ e2e-aci: ## Run End to end ACI tests. Set E2E_TEST=TestName to run a single test
 e2e-ecs: ## Run End to end ECS tests. Set E2E_TEST=TestName to run a single test
 	go test -timeout 20m -count=1 -v $(TEST_FLAGS) ./tests/ecs-e2e
 
+cross-%:
```
gtardif, line 58: nit: maybe define some targets with explicit os/arch delegating to this one, so we know which ones are actually supported, and we have auto-completion on targets

gtardif: of course we would need different names for these explicit targets, so we can delegate to the generic one, so maybe only `cross-mac`, `cross-win`, `cross-arm` ? 
If this is not an easy fix, forget about it, just a small UX improvement

Dockerfile
chris-crone: `make cross` currently builds all the cross platforms so I'm not sure this'd be easy. By default, TARGETOS and TARGETARCH are set to the builder's value.

builder.Makefile
chris-crone: Yeah, I agree that it's hairy to read. Your suggestion would mean that to add a platform, we'd need to:
1. Add `cross-new-platform`
2. Add a call from `cross` to `cross-new-platform`
3. Add the new platform to `package`

If we'd like like to make it possible to build the platform from the top-level Makefile, we'd also need to add a new target there.

I guess it comes down to how often we're going to touch this code. We should likely add builds for `linux/arm` and `darwin/arm64` in the near future.

Makefile
chris-crone: We could add something like that as a wrapper on top. `cross-mac` should likely do `darwin/amd64` and `darwin/arm64` (i.e.: `cross-mac: cross-darwin-amd64 cross-darwin-arm64`) for example

Not sure what `cross-arm` would do?

gtardif:
I guess this is obsolete now...

PR #1086, opened 2020-12-28
Make --file and --project-name options global

### LINKED ISSUE #1069: Make `--file` and `--project-name` global compose options
To allow compatibility with dokcer-compose in terms of "smooth transition" we should move `--file` and `--project-name` option to the `compose` command and not repeat them on each relevant subcommand.

dominikbraun:
**What I did**

This PR adds global `--file` and `--project-name` to the `compose` command and removes them from all sub-commands. Some command-specific options like `runOptions` have been extended in order to embed global compose options.

I'm available on Slack for discussions around this.

**Related issue**

Fixes #1069.

### DIFF
--- a/cli/cmd/compose/compose.go
+++ b/cli/cmd/compose/compose.go
@@ -85,20 +85,20 @@ func Command(contextType string) *cobra.Command {
 	}
 
 	command.AddCommand(
-		upCommand(composeOpts, contextType),
-		downCommand(composeOpts),
-		psCommand(composeOpts),
-		listCommand(composeOpts),
-		logsCommand(composeOpts),
-		convertCommand(composeOpts),
-		runCommand(composeOpts),
+		upCommand(&composeOpts, contextType),
+		downCommand(&composeOpts),
+		psCommand(&composeOpts),
+		listCommand(&composeOpts),
+		logsCommand(&composeOpts),
+		convertCommand(&composeOpts),
+		runCommand(&composeOpts),
 	)
 
 	if contextType == store.LocalContextType || contextType == store.DefaultContextType {
 		command.AddCommand(
-			buildCommand(composeOpts),
-			pushCommand(composeOpts),
-			pullCommand(composeOpts),
+			buildCommand(&composeOpts),
+			pushCommand(&composeOpts),
+			pullCommand(&composeOpts),
 		)
 	}
 
--- a/cli/cmd/compose/compose.go
+++ b/cli/cmd/compose/compose.go
@@ -86,11 +86,11 @@ func Command(contextType string) *cobra.Command {
 
 	command.AddCommand(
 		upCommand(contextType),
-		downCommand(),
-		psCommand(),
-		listCommand(),
-		logsCommand(),
-		convertCommand(),
+		downCommand(globalOpts),
+		psCommand(globalOpts),
+		listCommand(globalOpts),
+		logsCommand(globalOpts),
+		convertCommand(globalOpts),
 		runCommand(),
 	)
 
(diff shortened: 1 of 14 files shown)


dominikbraun:
The E2E test fails:

```
Stderr:   can't find a suitable configuration file in this directory or any parent: not found
```

Is this even my fault? 🤔

cli/cmd/compose/compose.go
```diff
@@ -85,22 +85,26 @@ func Command(contextType string) *cobra.Command {
 
 	command.AddCommand(
 		upCommand(contextType),
-		downCommand(),
-		psCommand(),
-		listCommand(),
-		logsCommand(),
-		convertCommand(),
-		runCommand(),
+		downCommand(composeOpts),
```
chris-crone, line 88: You'll need to pass the `composeOpts` struct by reference here.

It's confusing but this is where we register commands and flags are only parsed later on. Passing the struct by copy here means that the persistent flags will set the value of the struct declared in this function but the passed struct will be copied before this and never have its fields set.

cli/cmd/compose/pull.go
```diff
@@ -30,8 +30,11 @@ type pullOptions struct {
 	composeOptions
 }
 
-func pullCommand() *cobra.Command {
-	opts := pullOptions{}
+func pullCommand(composeOpts composeOptions) *cobra.Command {
+	opts := pullOptions{
+		composeOptions: composeOpts,
```
chris-crone, line 35: This also needs to be a reference.

We haven't yet parsed the flags so we're making a copy of an empty struct here. What we want to do is pass a reference so that when the flags are parsed, our struct will have its fields set.

cli/cmd/compose/run.go
```diff
@@ -29,17 +29,18 @@ import (
...
-	Name        string
-	Command     []string
-	WorkingDir  string
-	ConfigPaths []string
-	Environment []string
-	Detach      bool
-	Remove      bool
+	Name    string
+	Command []string
+	Detach  bool
```
chris-crone, line 34: Note that we have `Name` and `Detach` in the `composeOptions` struct so we're shadowing them here. We should remove these fields as we did `ConfigPaths`, etc.

chris-crone (commented):
Thanks @dominikbraun! Great to see an external contribution :) 

Cobra argument parsing is confusing so if you don't understand my comments, don't hesitate to ask questions!

chris-crone:
> Is this even my fault?

We have some tests that can be flaky but in this case the test is failing because the `ConfigPaths` aren't being set because of the argument parsing copy issue I commented on.

cli/cmd/compose/compose.go
dominikbraun: Absolutely - I have no idea why I didn't have the execution order in mind here.

dominikbraun:
> > Is this even my fault?
> 
> We have some tests that can be flaky but in this case the test is failing because the `ConfigPaths` aren't being set because of the argument parsing copy issue I commented on.

I fixed the pointer issues and the E2E test fails with another error:

```
compose_test.go:135: assertion failed: runContainerID is ""
```

The desired container apparently isn't running.

ulyssessouza:
This looks like the project name is not set respecting the `-f filepath` for the command run. It takes the name of the service to run as project name.
I'll propose a fix tomorrow so you can rebase on it.

ulyssessouza:
Hello @dominikbraun. [This PR](https://github.com/docker/compose-cli/pull/1087) should solve the problem.
You can rebase on it to check and maybe give me some feedback so I can unblock you.

ulyssessouza:
The problem is that in line https://github.com/docker/compose-cli/pull/1086/files#diff-e264c9ddb65c5a12fcab213cdb50c52c3875b9ce5ed2f0b018350d865ab335e3R93 `composeOpts` is passed as pointer and the `convertCommand` sets the flag `format` with the default value of `yaml` so it overwrites the default of `ps` command (that doesn't recognizes it). 
I think that passing it by value would fix it.

ulyssessouza:
This will also need a rebase on `main` branch

dominikbraun:
Guess we can close this, as #1165 has been merged.

PR #1106, opened 2021-01-06
introduce --remove-orphans option

ndeloof:
**What I did**
introduce --remove-orphans option
detect cotnainers which claim to be part of the project but we don't have a matching service in project model
warn user or remove them

**Related issue**
https://github.com/docker/compose-cli/issues/1070

/test-aci
/test-ecs
/test-windows

image

### DIFF
--- a/local/compose/create.go
+++ b/local/compose/create.go
@@ -23,12 +23,10 @@ import (
 	"strconv"
 	"strings"
 
-	convert "github.com/docker/compose-cli/local/moby"
-	"github.com/docker/compose-cli/progress"
-
 	"github.com/compose-spec/compose-go/types"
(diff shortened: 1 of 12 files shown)


local/compose/create.go
```diff
@@ -23,6 +23,14 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/sirupsen/logrus"
```
gtardif, line 26: Nit : remove empty lines and regroup compose-cli imports, etc.

gtardif (approved):

gtardif:
Nice functional programming on containers, Thx! :) Almost feels like java code on collections

Issue #934, opened 2020-11-20
Local compose: support secrets inlined


gtardif:
Using compose command in local context, I should be able to run `docker compose up` and have secrets mounted in my containers, when specifying inlined secrets in my compose file(s) using secret files next to my docker-compose.yaml file

PR #1113, opened 2021-01-07
Implement secrets via bind-mounts for local compose

aiordache:
Implement compose secrets via bind mounts. 
 
```yaml
services:
  test:
    image: nginx
    secrets:
      - my_secret
secrets:
  my_secret:
    file: ./my_secret.txt
```
```shell
$ cat my_secret.txt 
blabla bbla ba
```
```
docker compose up -d
[+] Running 2/1
 ⠿ Network "test_default"  Created                                                               0.5s
 ⠿ test_test_1             Created                                                               0.0s

```
```
$ docker exec -it test_test_1 cat /run/secrets/my_secret
blabla bbla ba

```

Closes https://github.com/docker/compose-cli/issues/934

### DIFF
--- a/local/compose/create.go
+++ b/local/compose/create.go
@@ -310,6 +310,16 @@ func buildContainerMountOptions(p types.Project, s types.ServiceConfig, inherit 
 		if definedSecret.External.External {
 			return nil, fmt.Errorf("unsupported external secret %s", definedSecret.Name)
 		}
+
+		if contains(inherited, target) {
+			// remove inherited mount
+			pos := indexOf(inherited, target)
+			if pos >= 0 {
+				mounts = append(mounts[:pos], mounts[pos+1])
+				inherited = append(inherited[:pos], inherited[pos+1])
+			}
+		}
+
 		mount, err := buildMount(p, types.ServiceVolumeConfig{
 			Type:   types.VolumeTypeBind,
 			Source: definedSecret.File,
--- a/local/compose/create.go
+++ b/local/compose/create.go
@@ -296,6 +296,31 @@ func buildContainerMountOptions(p types.Project, s types.ServiceConfig, inherit 
 		}
 		mounts = append(mounts, mount)
 	}
+
+	secretsDir := "/run/secrets"
+	for _, secret := range s.Secrets {
+		target := secret.Target
+		if secret.Target == "" {
+			target = filepath.Join(secretsDir, secret.Source)
+		} else if !filepath.IsAbs(secret.Target) {
+			target = filepath.Join(secretsDir, secret.Target)
+		}
+
+		definedSecret := p.Secrets[secret.Source]
+		if definedSecret.External.External {
+			return nil, fmt.Errorf("unsupported external secret %s", definedSecret.Name)
+		}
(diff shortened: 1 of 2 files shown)


local/compose/create.go
```diff
@@ -296,6 +296,32 @@ func buildContainerMountOptions(p types.Project, s types.ServiceConfig, inherit
...
+		}
+		target := "/var/run/secrets"
+		if secret.Target == "" {
+			target = fmt.Sprintf("%s/%s", target, secret.Source)
+		} else if !filepath.IsAbs(secret.Target) {
+			target = fmt.Sprintf("%s/%s", target, secret.Target)
+		}
+		defSecret, ok := p.Secrets[secret.Source]
+
+		if !ok {
```
ndeloof, line 311: this is already checked by compose-go loader
https://github.com/compose-spec/compose-go/blob/master/loader/validate.go#L51

local/compose/create.go
```diff
@@ -296,6 +296,32 @@ func buildContainerMountOptions(p types.Project, s types.ServiceConfig, inherit
 		}
 		mounts = append(mounts, mount)
 	}
+	for _, secret := range s.Secrets {
+		if contains(inherited, secret.Target) {
```
ndeloof, line 300: doing so, IIUC, will prevent changing the secret file name in compose file and get sercet bind updated on `up`.
IMHO inherited volumes from previous container should be default be overriden by config.
Not sure how docker-compose handle this, we should align

local/compose/create.go
```diff
@@ -296,6 +296,32 @@ func buildContainerMountOptions(p types.Project, s types.ServiceConfig, inherit
...
+			target = fmt.Sprintf("%s/%s", target, secret.Source)
+		} else if !filepath.IsAbs(secret.Target) {
+			target = fmt.Sprintf("%s/%s", target, secret.Target)
+		}
+		defSecret, ok := p.Secrets[secret.Source]
+
+		if !ok {
+			return nil, fmt.Errorf("secret %s is not defined", secret.Source)
+		}
+		mount, err := buildMount(p, types.ServiceVolumeConfig{
```
ndeloof, line 314: should first check this is _not_ `External`

ndeloof (changes requested):
some redundant code with compose-go and a potential distinct behaviour vs docker-compose to be confirmed

local/compose/create.go
```diff
@@ -296,6 +296,30 @@ func buildContainerMountOptions(p types.Project, s types.ServiceConfig, inherit
 		}
 		mounts = append(mounts, mount)
 	}
+	for _, secret := range s.Secrets {
+		if contains(inherited, secret.Target) {
+			continue
+		}
+		secretsDir := "/var/run/secrets"
```
gtardif, line 303: That should be`/run/secrets` shouldn't it ?

local/compose/create.go
gtardif: +1 to align with docker-compose behaviour (and indeed I would expect to override with config in case of clash)

ndeloof (approved):

Issue #1115, opened 2021-01-07
Support CloudFormation customization using kustomize


ndeloof:
`kubectl` adopted [Kustomize](https://kustomize.io/) to apply patches to a set of yaml file and avoid templating solutions.
To address the need to customize CloudFormation template for various needs/unsupported features (ssl termination, healthcheck, ...) we could offer a comparable solution, which would allow user to still benefit lifecycle management by `compose up` without being locked by unsupported features.

PR #1116, opened 2021-01-08
add support for overlays on generated cloudformation template

ndeloof:
**What I did**
Inspired by kustomize.io support in `kubectl`, introduce `x-aws-cloudformation` extension to pass patch files to apply on the converted CloudFormation template with unsupported features.

This allows user to apply custom attributes on components we manage without being locked by unsupported features which hardly find their way in the compose model, while still being able to manage the deployment lifecycle just by `docker compose up` command.

**usage**
I want to deploy Jenkins:
```yaml
services:
  test:
    image: jenkins/jenkins:lts
    ports:
      - "80:80"
```

Unfortunately, jenkins respond to `/` with a 403 error as users need to login. I need to tweak the LoadBalancer TargetGroupt healthCheck rules accordingly...

Here comes my updated compose file:
```yaml
services:
  test:
    image: jenkins/jenkins:lts
    ports:
      - "80:80"

x-aws-cloudformation:
  Resources:
    TestTCP80TargetGroup:
      Properties:
        Matcher: 
          HttpCode: 200-499
```
        
```console
docker compose up
```

**implementation details**

This relies on kustomize's [kyaml](https://github.com/kubernetes-sigs/kustomize/tree/master/kyaml) library to merge yaml trees, which ensure we get consistent behaviour / avoid dependecies hell if we want to adopt a comparable mechanism in a kubernetes backend

**Related issue**
https://github.com/docker/compose-cli/issues/1115

/test-ecs

image


ndeloof:
as suggested dy @rumpl : this would better be set by `x-aws-*` extension filed to avoid complex command line and confusion with contexts of distinct types

rumpl (approved):

gtardif (approved):

gtardif:
Great ! In terms of example, when writing docs, we should probably give an example with 2 compose files, the original and a second one with the cloudformation extension, that would look nicer ?

cch1:
I think this is an excellent idea.  Furthermore, using the same model and underlying tooling as Kubernetes is really good thinking.  I will be able to use this immediately to tweak my ALB health checks per the given example.  If I ever learn enough CF templating, I will attempt to graft on https listeners/certs as well.

nebuk89:
I like this, it is a simple solution to providing limitless overrides for now and should allow us to later on ask people what they find they are including here most often/programmatically check this in Github for a better idea of what we need to support.

robogeek:
Very good.  In my mind patching the Cloud Formation file -- since it's YAML that means it can be modified like a data object -- would be a good approach.  I see that this simple example of the health check is simple and achievable.

What I'm interested to learn is whether this would also support changing the load balancer configuration to support HTTPS and domain names using a similar patching approach.  That requires a more significant set of changes.

Also.. the _kyaml_ library didn't seem to have documentation.  Where will we know how to use this feature to its fullest?

ndeloof:
kaml is missing technical documentation, but you don't realy need to know implementation details so you can use it as a yaml overlay library (like used by `kubectl`).

ndeloof:
documentation proposal: https://deploy-preview-12080--docsdocker.netlify.app//cloud/ecs-integration/#tuning-the-cloudformation-template

cch1:
The documentation looks pretty good.  Frankly, it could use some editing for clarity and easy reading -but the essentials are in there.  For example `This is currently not supported by the ECS integration with lack for an equivalent abstraction in the Compose specification` should be written as `This is currently not supported by the ECS integration for lack of an equivalent abstraction in the Compose specification` or `This is currently not supported by the ECS integration due to the lack of an equivalent abstraction in the Compose specification`.

cch1:
Also, perhaps due to my lack of understanding of compose CLI and CloudFormation templates, I find the [SSL example](https://deploy-preview-12080--docsdocker.netlify.app//cloud/ecs-integration/#setting-ssl-termination-by-load-balancer) a bit confusing due to the _name_ of the auto-generated listener: why would a port 80 listener have an associated certificate?  My limited experience with SSL termination required a custom 443 Listener aimed at the auto-generated target group.  Perhaps some commentary about the specifics of the example could be provided?

cch1:
Any feedback on ^^, @ndeloof ?  Others?

ndeloof:
SSL does not **require** 443, just 443 is the default port for HTTP protocol.
The documentation is about exposing a HTTP service (so, container exposes port 80, but could be 5555) with SSL termination managed by AWS LoadBalancer. Without port translation set on Listener, same port is exposed with HTTPS, but you could also override the Listener configuration using overlays to expose 443 instead. Just I don't want docker documentation to become a Cloudformation tutorial, this is only for advanced users.

cch1:
I think I now see the source of my confusion your example "converts" the `WebAppTCP80Listener` into a listener that does not listen on port 80, but the (misleading) name remains.  I understand how this is the most expedient way of solving the problem.

ndeloof:
`WebAppTCP80Listener` route traffic to container's port 80. But also listen on port 80 if configuration is not modified to map to another external port.

robogeek:
It's technically correct that one could run HTTPS/SSL on any port.  But common practice is port 80 is only for HTTP, and port 443 is only for HTTPS.

I posted in the Slack an outline of what I've done to implement HTTPS on the AWS Load Balancer.  Namely -- the port 80 Listener is configured to redirect to the port 443 Listener, and the port 443 Listener is configured to terminate the SSL connection.  The port 443 listener is configured to default to providing an error message, BUT it is also configured with two rules to detect the domain name.  For the www-prefix name, the rule redirects to the base domain name.  For the base domain name, the rule forwards the connection to the port 80 target group.  There is no port 443 target group.

I think the SSL example you have in the online documentation is misleading because it does not follow usual practice for SSL.

Deninc:
@robogeek can you please update your solution somewhere on public? many thanks.

robogeek:
@Deninc - Since writing the above comment, I wrote a blog post describing how I have implemented HTTPS on the Docker ECS Context, using a Cloud Formation to customize an ALB.  

https://techsparx.com/software-development/docker/docker-ecs/load-balancer/https.html

The outline in that post is in turn derived from a book I've published -- "Deploying Docker Containers to AWS using Terraform".  The book has a long chapter showing several examples of using the Docker ECS Context.

PR #1118, opened 2021-01-08
Do not send metrics if executed as CLI backend

gtardif:

**What I did**
* stop sending metrics when compose-cli is executed as a CLI backend, as Docker CLI will handle metrics in this case

manually tested with local build of CLI including backend capabilities and metrics.
This is zero impact on the behavior of the compose cli if not used as a CLI backend

**Related issue**
https://github.com/docker/dev-tooling-team/issues/248

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -17,6 +17,7 @@
 package metrics
 
 import (
+	"os"
 	"strings"
 
 	"github.com/docker/compose-cli/utils"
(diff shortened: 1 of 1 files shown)


ndeloof (approved):
I assume this is a temporary solution and the whole metrics-gathering code will be removed once compose-cli becomes a backend-only implementation?

gtardif:
> I assume this is a temporary solution and the whole metrics-gathering code will be removed once compose-cli becomes a backend-only implementation?

Indeed !

PR #1136, opened 2021-01-13
Compose unit test networks & volumes

gtardif:
**What I did**
* move all network name normalizaiton into compose-go, same for volumes, secrets, configs
* add network e2e on networks with specified name
* small refactoring & unit tests

**Related issue**
https://github.com/docker/dev-tooling-team/issues/257

/test-aci
/test-ecs

### DIFF
--- a/ecs/testdata/simple/simple-cloudformation-conversion.golden
+++ b/ecs/testdata/simple/simple-cloudformation-conversion.golden
@@ -30,7 +30,7 @@ Resources:
       - Key: com.docker.compose.project
         Value: TestSimpleConvert
       - Key: com.docker.compose.network
-        Value: default
+        Value: TestSimpleConvert_default
       VpcId: vpc-123
     Type: AWS::EC2::SecurityGroup
   DefaultNetworkIngress:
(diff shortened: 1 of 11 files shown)


ecs/tags.go
```diff
@@ -44,15 +44,15 @@ func serviceTags(project *types.Project, service types.ServiceConfig) []tags.Tag
...
-func networkTags(project *types.Project, net types.NetworkConfig) []tags.Tag {
+func networkTags(project *types.Project, name string) []tags.Tag {
 	return []tags.Tag{
 		{
 			Key:   compose.ProjectTag,
 			Value: project.Name,
 		},
 		{
 			Key:   compose.NetworkTag,
-			Value: net.Name,
```
gtardif, line 55: @ndeloof I had to adapt this to keep cloudformation_test.go unchanged. Not sure what's best, set this tag to network key, or use network name that is now prefixed with project name (cf https://github.com/docker/compose-cli/runs/1696392558#step:7:59)

ecs/testdata/simple/simple-cloudformation-conversion.golden
```diff
@@ -30,7 +30,7 @@ Resources:
       - Key: com.docker.compose.project
         Value: TestSimpleConvert
       - Key: com.docker.compose.network
-        Value: default
+        Value: TestSimpleConvert_default
```
gtardif, line 33: Side effect on ECS volumes, this seems better than it was before (using the actual volume name rather than yaml key), just checking this sounds a sound change @ndeloof

ndeloof (approved):

PR #1137, opened 2021-01-13
Remove metrics and version hacks

ulyssessouza:
**What I did**
Remove metrics. 
The metrics will be sent from docker/cli instead

Remove version hacks, since this will be handled by `docker/cli`

**Related issue**
https://github.com/docker/dev-tooling-team/issues/252


gtardif:
LGTM, but we'll hold this a bit until compose-cli is actually used as a backend. Can you make this a draft PR for now ?

gtardif:
I think we can close this one and re-open when we have installed compose-cli as a backend of Docker-cli... or redo, depending on how rebasing works then

Issue #1141, opened 2021-01-14
compose down with file path


ndeloof:
**Description**
`compose.Down` uses project name as parameter, but on local backend has to resolve the compose.yaml file from container labels so it can handle service dependencies.

for the many cases where the user just pass a compose file (or the cli just guess which one to use) it would be simpler/safer to define a second API to just give backend the full compose model, so we don't have to rely on labels.

PR #1144, opened 2021-01-14
run "down" using project if it was set for command

ndeloof:
**What I did**

Pass `Project` to Down if it was set on command line, so we ddon't need to search for compose files

**Related issue**
https://github.com/docker/compose-cli/issues/1141

### DIFF
--- a/cli/cmd/compose/compose.go
+++ b/cli/cmd/compose/compose.go
@@ -50,16 +50,24 @@ func (o *composeOptions) toProjectName() (string, error) {
 		return o.ProjectName, nil
 	}
 
-	options, err := o.toProjectOptions()
-	if err != nil {
-		return "", err
-	}
-
-	project, err := cli.ProjectFromOptions(options)
+	project, err := o.toProject()
 	if err != nil {
 		return "", err
 	}
 	return project.Name, nil
+}
+
+func (o *composeOptions) toProject() (*types.Project, error) {
+	options, err := o.toProjectOptions()
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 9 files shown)


cli/cmd/compose/compose.go
```diff
@@ -45,21 +45,21 @@ func addComposeCommonFlags(f *pflag.FlagSet, opts *composeOptions) {
 	f.BoolVarP(&opts.Quiet, "quiet", "q", false, "Only display IDs")
 }
 
-func (o *composeOptions) toProjectName() (string, error) {
+func (o *composeOptions) toProjectName() (string, *types.Project, error) {
```
gtardif, line 48: Should we just return Project here, and name the function toProject() ? the signature is strange

gtardif (approved):
LGTM with minor improvement on toProjectName that now has a strange signature

cli/cmd/compose/compose.go
ndeloof: I'm not sure about adequate naming here, as this has to return either name or name _and_ project.
Maybe we could use `golang.org/x/functional/Either` :P

gtardif: My point was : we could return just the project and not the name (that's inside the project). And update the caller to use p.Name (that will not require `if err != nil` in this case :) ).

ndeloof: sure, but doing so we don't make it clear project has been loaded and is fully set or we just have a name. Backend code would have to check something like `if project.Services == nil` to distingish those scenarios.

gtardif (approved):

Issue #1140, opened 2021-01-14
Split repo to better isolate local vs aci vs ecs backends


ndeloof:
**Description**
Whenever a mono-repo make it easier to iterate over shared API and move code around, this doesn't mean we need to develop local+aci+ecs backends asa monolith.

Repo should be re-organized to create isolated components, not just packages. 

https://github.com/kubernetes-sigs/kustomize/ demonstrate a valid approach relying on go modules with relative paths `replace` directives, which allow cross-component dependencies for flexible development but well defined boundaries between components.

compose-cli must be split into:
- cli
- api (shared types, utility and go-plumbing code)
- local backend 
- ecs backend
- aci backend

as a further step, each backend could be packaged into a dedicated binary, using the cli module to opt-in for commands and flags being supported (to enforce an homogeneous UX).

gtardif:
Closing this, see discussion in #1174

PR #1150, opened 2021-01-15
Refacto folders

gtardif:
**What I did**
Move some root folders under either /cli, /api or /utils, preparing for separation of go modules, cf https://github.com/docker/compose-cli/issues/1140

**Related issue**
https://github.com/docker/compose-cli/issues/1140


gtardif:
Reading commit list will make it easier to review... :) 
Next step: remove example context & backend, then break into separate go modules

gtardif:
I moved context & config into /api, it's not really api, but common to other packages. We could move that to a common, including /utils as well... Happy to discuss this

chris-crone:
I don't have strong opinions here but I added @rumpl who might

rumpl (approved):

ndeloof (approved):

PR #1163, opened 2021-01-20
Stop project on Ctrl+C

ulyssessouza:
**What I did**
`Stop` project on Ctrl+C instead of `Down`

### DIFF
--- a/local/compose/stop.go
+++ b/local/compose/stop.go
@@ -1,0 +1,53 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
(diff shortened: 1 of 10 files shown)


local/compose/stop.go
```diff
@@ -0,0 +1,53 @@
...
+		All:     true,
+	})
+	if err != nil {
+		return err
+	}
+
+	err = InReverseDependencyOrder(ctx, project, func(c context.Context, service types.ServiceConfig) error {
+		serviceContainers, others := containers.split(isService(service.Name))
+		err := s.stopContainers(ctx, w, serviceContainers)
+		containers = others
```
gtardif, line 48: What is this for ? (things would more clear with no concurrent update on `containers`, and I don't think it's needed, since we re-filter on service.Name everytime)

ecs/context_test.go
```diff
@@ -43,7 +43,7 @@ func TestCreateContextDataFromEnv(t *testing.T) {
 	assert.Equal(t, desc, "credentials read from environment")
 }
 
-func TestCreateContextDataByKeys(t *testing.T) {
+func _TestCreateContextDataByKeys(t *testing.T) {
```
gtardif, line 46: I guess this is a temporary ignore, not to commit in the PR ?

ulyssessouza: Ops... Fixed

local/compose/stop.go
ulyssessouza: This just makes the `stop` pass through the services in the same order as the `down`, that means "respecting the dependencies order"

gtardif (approved):

PR #1162, opened 2021-01-20
Make `--file` and `--project-name` global compose options

ndeloof:
**What I did**
define `projectOption` and made `--project-name`, `--file`, `--workdir` and `--env-file` global compose options
to replicate docker-compose UX

most of the code was cherry-picked from @dominikbraun 's PR https://github.com/docker/compose-cli/pull/1086

**Related issue**
#1069

/test-windows

**(not mandatory) A picture of a cute animal, if possible in relation with what you did**

PR #1165, opened 2021-01-20
Make `--file` and `--project-name` global compose options

ndeloof:
**What I did**
define `projectOption` and made `--project-name`, `--file`, `--workdir` and `--env-file` global compose options
to replicate docker-compose UX

most of the code was cherry-picked from @dominikbraun 's PR https://github.com/docker/compose-cli/pull/1086

**Related issue**
#1069

/test-windows

this is same PR as https://github.com/docker/compose-cli/pull/1162 but ran from a branch, not a fork


gtardif:
re-running the windows test here... this is some flaky thing that is new since we run the metrics test with local backend rather than the example (hardcoded) backend. Need to fix this flakyness (in a separate PR) or not run the metrics test on windows as this seems to happen quite often

local/e2e/compose/compose_test.go
```diff
@@ -50,11 +50,11 @@ func TestLocalComposeUp(t *testing.T) {
 	const projectName = "compose-e2e-demo"
 
 	t.Run("up", func(t *testing.T) {
-		c.RunDockerCmd("compose", "up", "-d", "-f", "./fixtures/sentences/docker-compose.yaml", "--project-name", projectName, "-d")
+		c.RunDockerCmd("compose", "-f", "./fixtures/sentences/compose.yaml", "--project-name", projectName, "up", "-d")
```
gtardif, line 53: This change was made just to check we can put `-f` before or after `up`, right ? not because something is not supported anymore. Rebuilt and tester, seems OK, we still do support -f in both places

gtardif (approved):

local/e2e/compose/compose_test.go
ndeloof: I didn't tested we support `up -f` as I wanted to align with docker-compose which doesn't, but more flexibility doesn't hurt

PR #1172, opened 2021-01-21
introduce `--filter` option on `compose ls`

ndeloof:
**What I did**
Introduced `--filter` option on `compose ls` so one can limit output

removed support for project name which doesn't make much sense when listing stacks

/test-ecs
/test-aci
/test-windows

### DIFF
--- a/local/compose/ls.go
+++ b/local/compose/ls.go
@@ -27,7 +27,7 @@ import (
 	"github.com/docker/docker/api/types/filters"
 )
 
-func (s *composeService) List(ctx context.Context, projectName string) ([]compose.Stack, error) {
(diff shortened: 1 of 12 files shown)


local/compose/ls.go
```diff
@@ -27,15 +27,19 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
 
-	return containersToStacks(list)
+	stacks, err := containersToStacks(list)
+	if err != nil {
+		return nil, err
+	}
+	return stacks, nil
```
gtardif, line 42: Why changing `return containersToStacks(list)`?

gtardif (approved):
Just a question on returning stack,err, I guess there's a reason I'm missing ?

PR #1174, opened 2021-01-25
Separate go modules for API, cli & backends

gtardif:
**What I did**
* fixed a few dependency issues
* moved each backend to it's own go module, + go modules for api & cli
* updated Dockerfile & makefile to make all target work with multi-modules
* Tested with InterlliJ & VSCode, VSCode needs specific setting to support go multi-modules in workspace, updated BUILDING .md

Notes :
* go modules in this repo are importing each other locally, not using module versions. They all share the same version (based on the repo tag), and are moving together as we add more commits to the repo. It does not require to tag each module separately, nor increment or release one module before using new fields/API in another module.
* Separating go modules put light on some wrong dependencies we have (api depends on cli through utils [here](https://github.com/docker/compose-cli/pull/1174/commits/eadcdc56f7fe890805952c4df2373a420d0d4171#diff-598bf143e0a0756fbd24d9bf0b24144d62245e36a7d135e050f2a75436f90f87R24))
* Using this separation for Kubernetes backend also showed we didn't build/lint/test the kube backend properly
* For now I did not activate kube lint/test; as it is on main, this needs an additional PR
* ACK: Adding new modules requires a bit of attention, to add each module in root Dockerfile & makefile. Not complicated, but I'd love to make this more transparent
* Running `make commands` in VSCode terminal will output errors with relative paths from submodules, making links not clickable to go directly to VSCode files (probably the same in intelliJ). This is a bit problematic, not sure the module separation is worth loosing the ability to click in error logs and landing directly in the source code (although we have go "problems" integrated properly in VSCode and intelliJ, so it might not be too critical)

**Related issue**
https://github.com/docker/compose-cli/issues/1140

/test-aci
/test-ecs
/test-windows


gtardif:
ah, more rebase (kube...)

ndeloof (approved):

gtardif:
Discussing a bit more with @ndeoof, we see this IDE limitation as kind of a blocker for the moment moving forward with separate modules in sub folders : 
> Running make commands in VSCode terminal will output errors with relative paths from submodules, making links not clickable to go directly to VSCode files (probably the same in intelliJ). 

Putting this on the side for now, and will fix dependency issues / kube backend issues but without changing the go module setup.

gtardif:
Closing this, no more feedback on it, assuming we don't want to break IDE ergonomics. We can resurrect this later as needed, when we have more reasons to move to separate modules

PR #1175, opened 2021-01-25
Refactor kubernetes context create

aiordache:
Refactored the context create: 
- renamed `kubernetes` type to `k8s` (conflict with the legacy `kubernetes` endpoint from Docker contexts).
- parse kubeconfigs for the list of contexts
- `.FromEnvironment` option should enable the use of the current context in a kubeconfig passed via `$KUBECONFIG`

```
$ docker context create k8s test
? Create a Docker context using:  [Use arrows to move, type to filter]
> Context from kubeconfig file
  Kubernetes environment variables
```
```
$ docker context create k8s test
? Create a Docker context using: Context from kubeconfig file
? Select kubeconfig context  [Use arrows to move, type to filter]
> kind-kind
  kind-cluster2
```
```
$ docker context create k8s test
? Create a Docker context using: Context from kubeconfig file
? Select kubeconfig context kind-kind
Successfully created k8s context "test"
```
```
$ docker context create k8s test --from-env
Successfully created k8s context "test"
```
```
$ docker context create k8s test --kubeconfig ~/kubeconfig.json
? Select kubeconfig context  [Use arrows to move, type to filter]
> kind-kind
  kind-cluster2

```
Closes https://github.com/docker/dev-tooling-team/issues/271

### DIFF
--- a/api/context/store/store.go
+++ b/api/context/store/store.go
@@ -57,7 +57,7 @@ const (
 	LocalContextType = "local"
 	// KubeContextType is the endpoint key in the context endpoints for a new
 	// kube backend
-	KubeContextType = "k8s"
+	KubeContextType = "kube"
 )
 
 const (
--- a/api/context/store/store.go
+++ b/api/context/store/store.go
@@ -57,7 +57,7 @@ const (
 	LocalContextType = "local"
 	// KubeContextType is the endpoint key in the context endpoints for a new
 	// kube backend
-	KubeContextType = "kubernetes"
+	KubeContextType = "k8s"
 )
 
 const (
--- a/kube/charts/kubernetes/context.go
+++ b/kube/charts/kubernetes/context.go
@@ -1,0 +1,50 @@
+// +build kube
+
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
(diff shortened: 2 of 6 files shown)


kube/charts/kubernetes/context.go
```diff
@@ -0,0 +1,50 @@
+// +build kube
```
gtardif, line 1: I don't think this should be under charts, it could go next to kube/context.go

api/context/store/store.go
```diff
@@ -57,7 +57,7 @@ const (
 	LocalContextType = "local"
 	// KubeContextType is the endpoint key in the context endpoints for a new
 	// kube backend
-	KubeContextType = "kubernetes"
+	KubeContextType = "k8s"
```
gtardif, line 60: Is there really a conflict here ? if we know a context is of type kubernetes, the endpoint will also be named kubernetes, but won't be stored next to a "docker" endpoint ? (Happy to discuss this in more details)

gtardif (commented):
Need to discuss more in details the renaming of "kubernetes", and potential endpoint conflict. Appart from this, LGTM, tested the binary, deployed a 3 service stack (up/down) successfully. (will need to check in more details port exposure, but that's totally separated from this PR)

api/context/store/store.go
ndeloof: UX should use `kubernetes`, internal type can be `kube`, `k8s.io` or anything else

gtardif (approved):

Issue #1059, opened 2020-12-14
support --tail option


dolanor:
**Description**

**Steps to reproduce the issue:**
1. docker compose up -d
2. docker compose logs --tail=10

**Describe the results you received:**
`unknown flag: --tail`

**Describe the results you expected:**
the 10 last line in the log related to this `docker-compose.yml` file.

PR #1178, opened 2021-01-26
introduce compose logs --tail and --follow options

ndeloof:
**What I did**
introduce compose logs `--tail` and `--follow` options
`--tail` is only supported on local backend

**Related issue**
https://github.com/docker/compose-cli/issues/1059

/test-aci
/test-ecs


gtardif (approved):
LGTM. ACI red test i flaky (has happened previously on the main branch), we need to fix the flakyness in a separate PR, I think we're safe to merge this one

PR #1183, opened 2021-01-27
introduce `--pull` flag to set pull policy

ndeloof:
**What I did**
Allow user to pass `--pull=never` or `--pull=always` to force some image lifecycle

**Related issue**
see https://github.com/docker/compose/issues/6464 and related debates on pull vs build vs never pull vs pull before build etc

### DIFF
--- a/cli/cmd/compose/up.go
+++ b/cli/cmd/compose/up.go
@@ -35,7 +35,8 @@ import (
 // composeOptions hold options common to `up` and `run` to run compose project
 type composeOptions struct {
 	*projectOptions
-	Build bool
+	Build      bool
+	PullPolicy string
 	// ACI only
 	DomainName string
 }
@@ -70,6 +71,10 @@ func upCommand(p *projectOptions, contextType string) *cobra.Command {
 		Use:   "up [SERVICE...]",
 		Short: "Create and start containers",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if opts.Build {
+				opts.PullPolicy = types.PullPolicyBuild
+			}
+
 			switch contextType {
 			case store.LocalContextType, store.DefaultContextType, store.EcsLocalSimulationContextType:
 				if opts.forceRecreate && opts.noRecreate {
@@ -94,8 +99,8 @@ func upCommand(p *projectOptions, contextType string) *cobra.Command {
 		flags.BoolVar(&opts.forceRecreate, "force-recreate", false, "Recreate containers even if their configuration and image haven't changed.")
 		flags.BoolVar(&opts.noRecreate, "no-recreate", false, "If containers already exist, don't recreate them. Incompatible with --force-recreate.")
 		flags.BoolVar(&opts.noStart, "no-start", false, "Don't start the services after creating them.")
+		flags.StringVar(&opts.PullPolicy, "pull", "", `Define pull policy for service images ("always"|"missing"|"never"|"build").`)
 	}
-
 	return upCmd
 }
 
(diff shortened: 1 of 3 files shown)


cli/cmd/compose/up.go
```diff
@@ -74,7 +79,9 @@ func upCommand(p *projectOptions, contextType string) *cobra.Command {
 	if contextType == store.AciContextType {
 		flags.StringVar(&opts.DomainName, "domainname", "", "Container NIS domain name")
 	}
-
+	if contextType == store.LocalContextType {
+		flags.StringVar(&opts.PullPolicy, "pull", "", "Define pull policy for service images (always|never|if_not_present|build).")
```
thaJeztah, line 83: `docker run` uses "missing" for the `if_not_present` case (which is the default);

```
      --pull string                    Pull image before running ("always"|"missing"|"never") (default "missing")
```

For backward-compatibility (`--pull` used to be a "boolean" on `docker run`), specifying `--pull` without a value is equivalent to `--pull=always`. Not sure if we need to add that logic here (because this is a new flag).

What's the `build` option? It's a bit unclear to me what it's intended for; if that's meant for "only do a pull if we also have to do a build", should we keep it simple, and keep that for `docker compose build [--pull=<policy>]` ?

ndeloof: this is just plain https://github.com/compose-spec/compose-spec/blob/master/spec.md#pull_policy

ndeloof: that's a bit unfortunate defining this pull_policy as part of the compose-spec was not coordinated with introducion of the docker run `pull` flag for consistency :'(

thaJeztah: With the compose-spec not having a release yet, could we still change? I see I left a comment at the time in https://github.com/compose-spec/compose-spec/pull/78#discussion_r435800728 as well

TBH; Overall a bit on the fence of having it as an option in the compose-file itself, as it feels more of an option that should be set by the person doing the deploy, not a "declarative" option that specifies/defines the stack

gtardif (approved):
LGTM after updating with https://github.com/compose-spec/compose-go/pull/99

thaJeztah:
GitHub was a bit too eager closing

Screenshot 2021-01-29 at 16 47 16

gtardif:
Thx @thaJeztah ! 
https://github.com/docker/compose-cli/pull/1183 is now merged, @ndeloof we can bump version and use "missing" here

thaJeztah:
I'm still confused by the `--pull-policy=build` option (even more so because it's in the compose-file itself);

> build: Compose implementations SHOULD build the image. Compose implementations SHOULD rebuild the image if already present.

- `docker-compose up --build` is (currently) to trigger a build; the builder (if no changes) should then use its caching mechanism to determine if a rebuild is needed
- `--pull-policy=always` (note that _currently_ `docker (buildx) build` only supports `--pull` (boolean)) would in case of a "build" also update images used for the build itself

Note that the combination of `--build` and "pulling" can be tricky (what comes first?);

- a service that _builds_ an image (and has "just" rebuilt its image) would first _overwrite_ the fresh (locally built) image, then the build would overwrite it _again_ with a the new version
- having a pull-policy in the compose-file (in various situations) would now effectively "force" a user to also pass `--pull-policy` on the command-line to override the behavior (local development cycle) to prevent work being "un-done"

So I'm still quite uncomfortable with the pull policy being in the compose-file itself; this really feels like an option that should generally be passed on the commandline, as it's not declarative.

ndeloof:
> I'm still confused because it's in the compose-file itself

The compose spec actually defines an _model_, and attributes can be set by yaml _or_ command line flags, depending the implementation. This is up to users to decide if they want to hard code some behaviour in the yaml file or repeat `--pull` option. Basically, many users rely on compose to make `docker run` commands shorter and not have to remember tons of flags.


> Note that the combination of --build and "pulling" can be tricky (what comes first?);

this is what pull_policy has been designed for. `pull_policy: build` means image should be built from source, not pulled. Compose spec is not designed to follow buildx UX :D

thaJeztah:
> Basically, many users rely on compose to make docker run commands shorter and not have to remember tons of flags.

I understand how it's used, but still think there's a separation between describing/declaring a "stack", and runtime options / options for a specific deploy (I recall some discussion on that in https://github.com/docker/compose/issues/846)

ndeloof:
Most of the `deploy` section is actually about runtime options. It's way too late to make compose file format a clean, abstract declarative model.
From a pure usability point of view, users have asked for a way to get `compose up` to build first of pull first (on disctinct issues) with reasonable usage scenarios. With this attribute we give them an opportunity to achieve this. Obviously flags allow to override for other usages.

ndeloof:
closing this, will see later if we want to introduce this flag to compose command

PR #1188, opened 2021-01-28
introduce compose create command

ndeloof:
**What I did**
Introduce `docker compose create` command for parity with docker-compose

**Note**
This de-deprecates `create` command (https://github.com/docker/compose/commit/40c05cfc1e33077f7fe35ebabf64af3968ef58c8)
to maintain parity with `docker create` command and avoid ambiguous `--no-xyz=true` flags

### DIFF
--- a/cli/cmd/compose/compose.go
+++ b/cli/cmd/compose/compose.go
@@ -107,6 +107,7 @@ func Command(contextType string) *cobra.Command {
 			buildCommand(&opts),
 			pushCommand(&opts),
 			pullCommand(&opts),
+			createCommand(&opts),
 		)
 	}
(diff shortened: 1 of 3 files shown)


gtardif:
Makes sense to me to add `docker compose create` next to `compose start`, `compose stop` and be aligned with `docker create`. 
I don't know what was the historical reason to deprecate this in `docker-compose`, I don't see `docker-compose up --no-start` being more clear or intuitive. I'd prefer to not mark `compose create` as deprecated.

gtardif:
Ah, sorry, conflict (with start/stop I guess)... LGTM

gtardif (approved):

Issue #1182, opened 2021-01-27
Kube: expose ports and resolve service names internally


gtardif:
Services exposing public ports should be mapped to service type "LodBalancer"
This si not the case due to wrong heuristics on Compose project names (expecting project.Network to have zero entry, where it actually has 1 for "default"

Also, it seems services can't see each others, although service type "ClusterIP" is used by default as expected, if not LoadBalancer when exposing public ports.

PR #1196, opened 2021-01-28
Support exposing ports and cross-service communication

gtardif:

**What I did**
* Add headless service for internal cross-service communication
* Services exposing ports mapped to load-balancer
* add some unit tests
* Added a sample composefile that deploys successfuly on Desktop kube cluster

**Related issue**
Fixes https://github.com/docker/compose-cli/issues/1182

### DIFF
--- a/kube/charts/kubernetes/kube.go
+++ b/kube/charts/kubernetes/kube.go
@@ -33,6 +33,12 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const (
+	headlessPort      = 55555
+	headlessName      = "headless"
+	clusterIPHeadless = "None"
+)
+
 //MapToKubernetesObjects maps compose project to Kubernetes objects
 func MapToKubernetesObjects(project *types.Project) (map[string]runtime.Object, error) {
 	objects := map[string]runtime.Object{}
@@ -70,7 +76,12 @@ func MapToKubernetesObjects(project *types.Project) (map[string]runtime.Object, 
 
 func mapToService(project *types.Project, service types.ServiceConfig) *core.Service {
 	ports := []core.ServicePort{}
+	serviceType := core.ServiceTypeClusterIP
+	clusterIP := ""
 	for _, p := range service.Ports {
+		if p.Published != 0 {
+			serviceType = core.ServiceTypeLoadBalancer
+		}
 		ports = append(ports,
 			core.ServicePort{
 				Name:       fmt.Sprintf("%d-%s", p.Target, strings.ToLower(p.Protocol)),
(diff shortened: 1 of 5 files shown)


kube/charts/kubernetes/kube_test.go
```diff
@@ -0,0 +1,94 @@
...
+		},
+		Spec: core.ServiceSpec{
+			Selector:  map[string]string{"com.docker.compose.service": "nginx", "com.docker.compose.project": ""},
+			ClusterIP: "None",
+			Ports: []core.ServicePort{
+				{
+					Name:       headlessName,
+					Protocol:   core.ProtocolTCP,
+					Port:       headlessPort,
+					TargetPort: intstr.IntOrString{IntVal: 55555},
```
ndeloof, line 89: should use `headlessPort` constant here

kube/charts/kubernetes/kube.go
```diff
@@ -33,6 +34,12 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const (
+	headlessPort      = 55555
```
ndeloof, line 38: any reason using this specific value?

kube/charts/kubernetes/kube_test.go
```diff
@@ -0,0 +1,94 @@
...
+			Kind:       "Service",
+			APIVersion: "v1",
+		},
+		ObjectMeta: meta.ObjectMeta{
+			Name: "nginx",
+		},
+		Spec: core.ServiceSpec{
+			Selector:  map[string]string{"com.docker.compose.service": "nginx", "com.docker.compose.project": ""},
+			ClusterIP: "None",
+			Ports: []core.ServicePort{
```
ndeloof, line 84: This _should_  not be required with https://github.com/kubernetes/dns/issues/174

kube/charts/kubernetes/kube.go
gtardif: taken from compose-on-kube, no specific reason I think

kube/charts/kubernetes/kube_test.go
gtardif: `Service "web" is invalid: [spec.ports[0].port: Invalid value: 0: must be between 1 and 65535, inclusive, spec.ports[0].targetPort` when trying to deploy without this artificial port. 
versions in go.mod don't seem very old (helm 3.5.0, kube client 0.20.1, there's one minor version available, but https://github.com/kubernetes/dns/issues/174 is 3 years old, so I don't think we're behind versions fixing this...)

gtardif: was probably another issue, fixed this

ndeloof (approved):

PR #1211, opened 2021-02-01
Add HEALTH column to compose ps

### LINKED ISSUE #1198: Add healthcheck status to ps
I love depends_on + healthchecks and use them in every project (and teach them to everyone I can).

With `docker-compose ps` we see healthcheck status, if any. Today in trying to do an up with this new tool, the lack of healthcheck status prevented me from understanding why all services weren't starting. I had to use `docker-compose ps` to see that one db had a unhealthy status, preventing other containers from starting.

ndeloof:
**What I did**
Inspect containers on `compose ps` to get health status

**Related issue**
closes #1198

**sample output**
```
NAME                SERVICE             STATE               HEALTH              PORTS
nicolas_test_1      test                running             healthy             
nicolas_toto_1      toto                running             healthy             
nicolas_truc_1      truc                running             unhealthy           
```

**Note**
This PR doesn't propose ECS implementation, but AFAIK the `DescribeService` API would also return health status

image

### DIFF
--- a/local/compose/ps.go
+++ b/local/compose/ps.go
@@ -23,6 +23,7 @@ import (
 
 	moby "github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/filters"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/docker/compose-cli/api/compose"
 )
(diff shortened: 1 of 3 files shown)


ulyssessouza:
I'm just not sure if this should be on it's own column. I tend to see/prefer `Health` as a component of `Up` status when a healthcheck exists. Looking at `docker-compose`, that's the case.

ndeloof:
docker-compose indeed return state and health as a single column `Up (healthy)`. As we are listing containers with this command this seems to me to be inacurate and I'd prefer we get better aligned with `docker ps`.

I'm not sure we want to replicate this in compose-cli. Table output is distinct whatever we deciden so users will have to adjust scripts if they used to parse this from scripts. But if this is the case, probably better they use `format=json`

docker-compose ps:
```
     Name           Command           State        Ports
--------------------------------------------------------
nicolas_test_1   ping localhost   Up (healthy)          
nicolas_toto_1   ping localhost   Up (healthy)          
nicolas_truc_1   ping localhost   Up (unhealthy)  
```

docker compose ps:
```
NAME                SERVICE             STATE               HEALTH              PORTS
nicolas_test_1      test                running             healthy             
nicolas_toto_1      toto                running             healthy             
nicolas_truc_1      truc                running             unhealthy         
```

gtardif:
Generally speaking, at least on the visual aspect I’d prefer to align with docker ps and not replicate the output of docker-compose :
```
     Name           Command           State        Ports
--------------------------------------------------------
```

And more specifically regarding health, I also prefer to more cleanly separate info and not mix 2 things in 1 column : 
`up (healthy)` . 
Parsing json will be much cleaner if it’s 2 different fields, and that should also map easily to the human-readable version

ndeloof:
@gtardif json could use separate fields, while table output could use docker-compose `state (health)` style. But better avoid such an heterogeneous rendering?

gtardif:
@ndeloof Indeed, I'd rather keep pretty & json homogeneous (and this is also more aligned with the docker CLI

local/compose/ps.go
```diff
@@ -37,32 +38,49 @@ func (s *composeService) Ps(ctx context.Context, projectName string) ([]compose.
...
-		var publishers []compose.PortPublisher
-		for _, p := range c.Ports {
-			var url string
-			if p.PublicPort != 0 {
-				url = fmt.Sprintf("%s:%d", p.IP, p.PublicPort)
+	summary := make([]compose.ContainerSummary, len(containers))
+	eg, ctx := errgroup.WithContext(ctx)
+	for i, c := range containers {
+		container := c
+		i := i
```
gtardif, line 45: I'll never get used to it... :)

gtardif (approved):

local/compose/ps.go
ndeloof: I tend to like it now, sort of a running gag

PR #1216, opened 2021-02-02
Introduce --profile option

ndeloof:
**What I did**
Added support for --profiles according to compose-spec

**Related issue**

image

### DIFF
--- a/cli/cmd/compose/compose.go
+++ b/cli/cmd/compose/compose.go
@@ -29,12 +29,14 @@ import (
 
 type projectOptions struct {
 	ProjectName string
+	Profiles    []string
 	ConfigPaths []string
 	WorkingDir  string
 	EnvFile     string
 }
 
 func (o *projectOptions) addProjectFlags(f *pflag.FlagSet) {
+	f.StringArrayVar(&o.Profiles, "profile", []string{}, "Specify a profile to enable")
 	f.StringVarP(&o.ProjectName, "project-name", "p", "", "Project name")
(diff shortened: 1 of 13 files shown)


cli/cmd/compose/compose.go
```diff
@@ -47,14 +49,14 @@ func (o *projectOptions) toProjectName() (string, error) {
...
-	project, err := o.toProject()
+	project, err := o.toProject(nil)
 	if err != nil {
 		return "", err
 	}
 	return project.Name, nil
 }
 
-func (o *projectOptions) toProject() (*types.Project, error) {
+func (o *projectOptions) toProject(services []string) (*types.Project, error) {
```
gtardif, line 59: +1 (and simplification in all impacted calls)

local/compose/attach.go
```diff
@@ -42,6 +42,7 @@ func (s *composeService) attach(ctx context.Context, project *types.Project, con
 	if err != nil {
 		return nil, err
 	}
+	containers = Containers(containers).filter(isService(project.ServiceNames()...))
```
gtardif, line 45: Nit: could we add this filtering next to `projectFilter(project.Name)` for more clarity ? I guess it would require a nice way to express and OR condition, that we don't have ?

gtardif (approved):

local/compose/attach.go
ndeloof: `projectFilter` produces a moby API filter, unrelated to `Containers.filter` which use forceps to tweak golang into something functional-ish  :P

Issue #1192, opened 2021-01-28
Initial Kube e2e test with kind


gtardif:
We don't have any e2e test for kubernetes, using kind might help to setup a kube cluster for e2e tests without too much pain
Expected : a first e2e test deploying a compose file, accessing the app (curl)

PR #1217, opened 2021-02-02
First kube e2e. Adapted context create kubernetes command to allow non interactive mode.

gtardif:

**What I did**
* Start kind cluster, wait... 
* deploy compose sample sentence stack. Wait... 
* check all services are wired up and we can get access service exposed on port 80
* Added Kube CI workflow (need to install jq, kubectl, kind)

We should probably check if a specific kube cluster already exists locally and reuse it, for local tests that will make things a lot quicker (but more risks of starting in a non totally clean env locally).

**Related issue**
Fixes https://github.com/docker/compose-cli/issues/1192

/test-kube

### DIFF
--- a/.github/workflows/kube-tests.yml
+++ b/.github/workflows/kube-tests.yml
@@ -1,0 +1,63 @@
+name: Kube integration tests
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  check-optional-tests:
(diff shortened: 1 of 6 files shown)


ndeloof (approved):

ndeloof:
nice one :+1:

aiordache (approved):

Issue #937, opened 2020-11-20
Local compose: support config


gtardif:
Using compose command in local context, I should be able to run `docker compose up` and have config mounted in my containers, when specifying config in my compose file(s)

gtardif:
@ulyssessouza this ticket was actually about supporting the `config` field in compose files, next to secrets and volumes :)

PR #1219, opened 2021-02-02
Add config command

ulyssessouza:
**What I did**
Add config command

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/937

### DIFF
--- a/aci/compose.go
+++ b/aci/compose.go
@@ -203,6 +203,10 @@ func (cs *aciComposeService) Convert(ctx context.Context, project *types.Project
 	return nil, errdefs.ErrNotImplemented
 }
 
+func (cs *aciComposeService) Config(ctx context.Context, project *types.Project, options compose.ConfigOptions) ([]byte, error) {
+	return nil, errdefs.ErrNotImplemented
+}
+
 func (cs *aciComposeService) RunOneOffContainer(ctx context.Context, project *types.Project, opts compose.RunOptions) error {
 	return errdefs.ErrNotImplemented
 }
--- a/api/client/compose.go
+++ b/api/client/compose.go
@@ -76,6 +76,10 @@ func (c *composeService) Convert(context.Context, *types.Project, compose.Conver
 	return nil, errdefs.ErrNotImplemented
 }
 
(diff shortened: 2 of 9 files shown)


ndeloof:
Why didn't you just made `config` an alias for `convert`?

ulyssessouza:
@ndeloof IMO it should be the other way around. 
`convert` to be an alias of `config`, since `config` already exists in `docker-compose` and is well known.
Then we just have to add the option for `format` on it.

ndeloof:
I'm fine we make `config` the default command name and `convert` an alias.
I can't see the benefit of having a `--format` option other than selecting `json|yaml`. There's no mean to ask generating `cloudformation` format but on ECS.

That's why I liked `convert` as a command to produce the platform's canonical format. Would be cloudformation for ECS context, plain compose-spec on local engine

ulyssessouza:
@ndeloof Got it! The idea was to have a way to overwrite the context to generate something different from what the context says, but thinking again, we can always use global option `--context` to select another context/type to generate another format. I'll remove that option then

ndeloof:
By the way, #937 is about `config` entries in the compose file mounted into containers, not about `docker compose config` command

ndeloof (changes requested):
IMHO we don't need a new command in the API. Just rename `Convert` and make `convert` an alias for the config command.
`ProjectConfigView` is still relevant as the local implementation for `Convert` dump way too much things. Another way would be to fix compose-go so `Project` can be Loaded then marshalled without adding unexpected stuff to the yaml stream.

gtardif:
Discussing a bit with @ndeloof, it makes sense to keepp the `--format` option only for yaml / json, and use context for switching between cloudformation, composefile or other output. One argument for this specifically is that for cloudformation, we need an actual connection to ECS to generate the template, so this can't be done without an ECS context anyway. 
Indeed the marshalling issue needs to be addressed, but I'd rather make sure the compose.Project struct in compose-go is something we can marshall from anywhere properly and use as a reference for compose files, so hide relevant fields in compose-go. 
The config / convert alias is actually already in the main branch, from https://github.com/docker/compose-cli/pull/1216, so we can close this PR I think.

PR #1225, opened 2021-02-03
Add kill command

ndeloof:
**What I did**
Added `compose kill` command for parity with docker-compose

image

### DIFF
--- a/local/compose/containers.go
+++ b/local/compose/containers.go
@@ -70,3 +70,9 @@ func (containers Containers) names() []string {
 	}
 	return names
 }
+
+func (containers Containers) forEach(fn func(moby.Container)) {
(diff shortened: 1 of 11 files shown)


local/compose/containers.go
```diff
@@ -70,3 +70,9 @@ func (containers Containers) names() []string {
 	}
 	return names
 }
+
+func (containers Containers) forEach(fn func(moby.Container)) {
```
gtardif, line 74: forEach in go... don't push things too far :)

gtardif (approved):

local/compose/containers.go
ndeloof: (｡◕‿◕｡)

Issue #1166, opened 2021-01-20
add --remove-orphans on docker compose down


gtardif:
We added `--remove-orphans` to the `up` command, we need to also add it to `down`

PR #1231, opened 2021-02-03
Expose `—remove-orphans` in compose down command

gtardif:

**What I did**
* just expose the flag, everything was already wired to the backend

**Related issue**
Fixes https://github.com/docker/compose-cli/issues/1166

### DIFF
--- a/cli/cmd/compose/down.go
+++ b/cli/cmd/compose/down.go
@@ -31,6 +31,7 @@ import (
 
 type downOptions struct {
 	*projectOptions
+	removeOrphans bool
 }
 
 func downCommand(p *projectOptions) *cobra.Command {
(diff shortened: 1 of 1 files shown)


gtardif:
@ndeloof not sure what is the best thing to do with https://github.com/docker/compose-cli/blob/main/ecs/local/compose.go#L161 : 
* expose the option in ecs local and ignore it, force it to true (current code)
* respect the option (I guess we don't want to do this as we want to delete side-car container)
* not expose the flag for ecs. (only expose it for local compose). Maybe the best option ?
Not sure why we expose this for `compose up`, where we already have some flags sections specific to context types

ndeloof (approved):

PR #1241, opened 2021-02-04
Add configs bind mount support

ulyssessouza:
**What I did**
Add configs bind mount read-only support

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/937

### DIFF
--- a/local/compose/create.go
+++ b/local/compose/create.go
@@ -388,23 +388,37 @@ func buildContainerMountOptions(p types.Project, s types.ServiceConfig, img moby
 	}
 	if img.ContainerConfig != nil {
 		for k := range img.ContainerConfig.Volumes {
-			mount, err := buildMount(p, types.ServiceVolumeConfig{
+			m, err := buildMount(p, types.ServiceVolumeConfig{
 				Type:   types.VolumeTypeVolume,
 				Target: k,
 			})
 			if err != nil {
 				return nil, err
 			}
-			mounts[k] = mount
+			mounts[k] = m
 
 		}
 	}
+
+	mounts, err := fillBindMounts(p, s, mounts)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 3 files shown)


local/compose/create.go
```diff
@@ -388,36 +388,91 @@ func buildContainerMountOptions(p types.Project, s types.ServiceConfig, img moby
 	}
 	if img.ContainerConfig != nil {
 		for k := range img.ContainerConfig.Volumes {
-			mount, err := buildMount(p, types.ServiceVolumeConfig{
+			bindMount, err := buildMount(p, types.ServiceVolumeConfig{
```
ndeloof, line 391: why rename var as `bindMount`? this slice also contain volume mounts, not just bind mounts

local/compose/create.go
```diff
@@ -388,36 +388,91 @@ func buildContainerMountOptions(p types.Project, s types.ServiceConfig, img moby
...
+
+func fillBindMounts(p types.Project, s types.ServiceConfig, m map[string]mount.Mount) (map[string]mount.Mount, error) {
 	for _, v := range s.Volumes {
-		mount, err := buildMount(p, v)
+		bindMount, err := buildMount(p, v)
 		if err != nil {
 			return nil, err
 		}
-		mounts[mount.Target] = mount
+		m[bindMount.Target] = bindMount
```
ndeloof, line 421: I'll never get used with this golang habit to use single letter variables...

ndeloof (approved):

local/compose/create.go
ulyssessouza: I just renamed because `mount` is the name of an imported package. So my IDE is complaining about shadowing. I'll get a better name

local/compose/create.go
gtardif: same for me... Javaist thing probably :)

gtardif (approved):

PR #1246, opened 2021-02-04
Add environment_files label on service's `env_file` section

ulyssessouza:
**What I did**

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/965

### DIFF
--- a/cli/cmd/compose/up.go
+++ b/cli/cmd/compose/up.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/docker/compose-cli/api/client"
 	"github.com/docker/compose-cli/api/compose"
@@ -174,6 +175,23 @@ func setup(ctx context.Context, opts composeOptions, services []string) (*client
 		for _, service := range project.Services {
 			service.PullPolicy = types.PullPolicyBuild
 		}
+	}
+	if opts.EnvFile != "" {
+		var services types.Services
+		for _, s := range project.Services {
+			ef := opts.EnvFile
+			if ef != "" {
+				if !filepath.IsAbs(ef) {
+					ef = filepath.Join(project.WorkingDir, opts.EnvFile)
+				}
+				if s.Labels == nil {
+					s.Labels = make(map[string]string)
+				}
+				s.Labels[compose.EnvironmentFileLabel] = ef
+				services = append(services, s)
+			}
+		}
+		project.Services = services
 	}
 
 	return c, project, nil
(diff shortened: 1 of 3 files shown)


local/compose/labels.go
```diff
@@ -19,24 +19,26 @@ package compose
...
-
-	//ComposeVersion Compose version
+	containerNumberLabel  = "com.docker.compose.container-number"
+	oneoffLabel           = "com.docker.compose.oneoff"
+	slugLabel             = "com.docker.compose.slug"
+	projectLabel          = compose.ProjectTag
+	volumeLabel           = compose.VolumeTag
+	workingDirLabel       = "com.docker.compose.project.working_dir"
+	configFilesLabel      = "com.docker.compose.project.config_files"
+	environmentFilesLabel = "com.docker.compose.project.environment_files"
```
ndeloof, line 35: docker-compose label is `LABEL_ENVIRONMENT_FILE = 'com.docker.compose.project.environment_file'` (no `s`)

local/compose/create.go
```diff
@@ -180,6 +180,9 @@ func (s *composeService) getCreateOptions(ctx context.Context, p *types.Project,
 	labels[workingDirLabel] = p.WorkingDir
 	labels[configFilesLabel] = strings.Join(p.ComposeFiles, ",")
 	labels[containerNumberLabel] = strconv.Itoa(number)
+	if len(service.EnvFile) > 0 {
+		labels[environmentFilesLabel] = strings.Join(service.EnvFile, ",")
```
ndeloof, line 184: You're storing the service's env_file attribute in label doing so, while the expected value is the one set by user as `--env-file` on command line (if present)
see 
-> https://github.com/docker/compose/blob/master/compose/cli/command.py#L39
-> https://github.com/docker/compose/blob/master/compose/cli/command.py#L162
-> https://github.com/docker/compose/blob/master/compose/cli/command.py#L177

ndeloof (changes requested):
typo in label name
incorrect value stored

local/compose/create.go
ulyssessouza: Good catch!

cli/cmd/compose/up.go
```diff
@@ -32,6 +33,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const environmentFileLabel = "com.docker.compose.project.environment_file"
```
ndeloof, line 36: why not define this in https://github.com/docker/compose-cli/blob/main/api/compose/tags.go with other labels?

ulyssessouza: Done

ndeloof (approved):

Issue #1248, opened 2021-02-05
[local] inconsistent behaviour and output for "docker compose logs"


thaJeztah:
**Description**

There is an inconsistency between "docker-compose" and "docker compose" when viewing logs.

This was reproduced with the example in https://github.com/docker/compose/issues/8098, after doing a `docker-compose up`, and a couple of `docker-compose run stderr`, after which the following containers were present;

```console
$ docker ps -a --no-trunc --format 'table {{.ID}}\\t{{.Names}}'
CONTAINER ID                                                       NAMES
240a42ef96d5d8f55124dd8a2f41d96b6ef36a672e75011b8aaf2965402173cc   repro-8098_stderr_run_b023423a37e3
dc977442275aa246162b381934c2d7abfba7162a206b1009e313a38a937f0a7d   repro-8098_stderr_run_c4af82367ed4
8fd782949ee6fa84d22e1e369be0b1a2895d0f1c3f92bceb36a7a6ebe3bd94d7   repro-8098_stderr_1
```

```console
$ docker compose logs stderr
The new 'docker compose' command is currently experimental. To provide feedback or request new features please open issues at https://github.com/docker/compose-cli
45bd8f002e241b48c5d2508e2208e976f58b39ddbdda0aa08db6d80f17601891 | log to stdout
45bd8f002e241b48c5d2508e2208e976f58b39ddbdda0aa08db6d80f17601891 | log to stderr
45bd8f002e241b48c5d2508e2208e976f58b39ddbdda0aa08db6d80f17601891 | log to stderr
45bd8f002e241b48c5d2508e2208e976f58b39ddbdda0aa08db6d80f17601891 | log to stdout
7cf44ebf7bfeb406abd2a7caceb9e79f91dd1f77a5b4ef375ea6e42a37125c19 | log to stderr
7cf44ebf7bfeb406abd2a7caceb9e79f91dd1f77a5b4ef375ea6e42a37125c19 | log to stdout
```

In an equivalent situation with `docker-compose` (standalone);

```console
$ docker-compose logs stderr
Attaching to repro-8098_stderr_1
stderr_1  | log to stderr
stderr_1  | log to stdout
```

Notice that:

1. `docker-compose` only prints logs for the container that was started as part of `docker-compose up`, and skips the `docker-compose run` containers, whereas `docker compose` prints logs for both
2. decoration is different (name of container versus ID of container)

For `1.`, I think this may need some discussion;

[...]

PR #1250, opened 2021-02-05
Filter out one_off containers on `compose logs`

ndeloof:
**What I did**
Exclude one_off container as we collect project containers for `compose logs`
Ue container name for line prefix, not ID (sha)

**Related issue**
https://github.com/docker/compose-cli/issues/1248

### DIFF
--- a/local/compose/labels.go
+++ b/local/compose/labels.go
@@ -18,6 +18,8 @@ package compose
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


thaJeztah (approved):
❤️ you're fast! LGTM

gtardif (approved):

PR #1252, opened 2021-02-05
Allow compose common fags  (-f, -p, ...) to be used after compose command or after subcommand (as it used to be).

gtardif:
Needed to set flags on both command and subcommand. 
Used to work setting the flags only on the subcommand but without `TraverseChildren:true` (also set at root level BTW)

**What I did**
* Common flags are set both on parent `compose` command and on sub commands
* changed one local e2e test to fail not only on cloud backend tests if we change this 

**Related issue**
Red tests on main: https://github.com/docker/compose-cli/runs/1839474094#step:8:31 ; same in ECS, ACI

/test-kube

### DIFF
--- a/cli/cmd/compose/compose.go
+++ b/cli/cmd/compose/compose.go
@@ -105,7 +105,14 @@ func Command(contextType string) *cobra.Command {
 		},
 	}
 
-	command.AddCommand(
+	addComposeCommand := func(cmds ...*cobra.Command) {
+		for _, subCmd := range cmds {
+			opts.addProjectFlags(subCmd.Flags())
+			command.AddCommand(subCmd)
+		}
+	}
+
+	addComposeCommand(
 		upCommand(&opts, contextType),
 		downCommand(&opts),
 		startCommand(&opts),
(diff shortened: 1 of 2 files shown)


aiordache (approved):

ndeloof:
How does this handle flag conflicts ?
Are -f and -p definitely reserved ?
Not a big fan of this, so my +0

gtardif:
> How does this handle flag conflicts ?

Fail fast : `panic: unable to redefine 'f' shorthand in "logs" flagset: it's already used for "follow" flag`

> This typically prevent we support compose logs -f as a short notation for --follow and so full docker-compose parity

Agreed this is a good reason to not proceed like this. 
Still usure if we should try to support -f -p for main commands `up` and `down`, and if we can avoid regressions on the ACI/ECS side. Lets discuss this on Monday.

gtardif:
Searching a bit for "docker compose up -f", there are quite a few users trying and failing running the docker-compose command with `-f` *after* `up`. Not easy to find the balance between retro-compatibility and makeing the command more easy to use. 
Also, having these common flags not available on subcommands means `docker compose up --help` will not list common flags, finding help on those is not easy/intuitive

gtardif:
Holding to keep compatibility with docker-compose.

Issue #1209, opened 2021-02-01
Add "docker compose up --exit-code-from" to the local backend


gabrieldemarmiesse:
Hi! 

We use `docker-compose` a lot, and we would like to switch to `docker compose` because it's better aligned with swarm (supports [extention fields](https://stackoverflow.com/questions/38567355/docker-compose-global-level-logging) and fuse correctly container arguments).

The big thing missing for us to switch is the option `--exit-code-from` that we use heavily in our CI.

It allows us to up the whole stack, wait until all containers are up with the `depends-on` field, and then run the service doing the tests (cypress). The service doing the tests returns 0 if everything was ok, and a non-zero error code otherwise.

So that's the use case.

My 2c on it: Since we have the opportunity to rework the API, maybe it would be possible to remove the flag and return the exit code of a single container when doing `docker compose up my-test-service`. Of course that wouldn't work when declaring no services, or multiple ones. I'll leave the decision to the maintainers. It's just a quick idea.

gabrieldemarmiesse:
Thanks!!!

PR #1259, opened 2021-02-08
Capture container exit code and dump on console

ndeloof:
Introduce "--abort-on-container-exit"
use `ContainerWait` API to capture container termination and exit code, dump on console and cancel the context, which trigger `Stop` execution.

Note: 
The UX need some eextra improvement to remove duplicated stop-cause being reported
```
test4_1 exited with code 137 
Aborting on container exit...
Gracefully stopping...
[+] Running 4/4
 ⠿ Container truc_test2_1  Stopping                                                                                                                                    0.6s
 ⠿ Container truc_test4_1  Stopped                                                                                                                                      0.0s
 ⠿ Container truc_test3_1  Stopping                                                                                                                                     0.6s
 ⠿ Container truc_test1_1  Stopping      
```

**Related issue**
required to later implement https://github.com/docker/compose-cli/issues/1209

### DIFF
--- a/api/compose/api.go
+++ b/api/compose/api.go
@@ -68,7 +68,7 @@ type StartOptions struct {
 	// Attach will attach to container and pipe stdout/stderr to LogConsumer
 	Attach LogConsumer
 	// Listener will get notified on container events
-	Listener Listener
+	Listener chan ContainerExited
 }
 
 // UpOptions group options of the Up API
@@ -186,11 +186,8 @@ type LogConsumer interface {
 	Status(service, container, message string)
 }
 
-// Listener get notified on container Events
-type Listener chan Event
-
-// Event let us know a Container exited
-type Event struct {
+// ContainerExited let us know a Container exited
+type ContainerExited struct {
 	Service string
 	Status  int
 }
--- a/api/compose/api.go
+++ b/api/compose/api.go
@@ -65,10 +65,8 @@ type CreateOptions struct {
 
 // StartOptions group options of the Start API
 type StartOptions struct {
-	// Attach will attach to container and pipe stdout/stderr to LogConsumer
-	Attach LogConsumer
-	// Listener will get notified on container events
-	Listener chan ContainerExited
+	// Attach will attach to service containers and pipe stdout/stderr to channel
+	Attach chan ContainerEvent
 }
 
 // UpOptions group options of the Up API
@@ -183,11 +181,18 @@ type Stack struct {
 // LogConsumer is a callback to process log messages from services
 type LogConsumer interface {
 	Log(service, container, message string)
-	Status(service, container, message string)
+	Status(service, container, msg string)
 }
 
-// ContainerExited let us know a Container exited
-type ContainerExited struct {
-	Service string
-	Status  int
+type ContainerEvent struct {
+	Type     int
+	Source   string
+	Service  string
+	Line     string
+	ExitCode int
 }
+
+const (
+	ContainerEventLog = iota
+	ContainerEventExit
+)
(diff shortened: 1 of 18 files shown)


cli/cmd/compose/up.go
```diff
@@ -144,7 +149,25 @@ func runCreateStart(ctx context.Context, opts upOptions, services []string) erro
...
 	}
 
-	err = c.ComposeService().Start(ctx, project, formatter.NewLogConsumer(ctx, os.Stdout))
+	ctx, cancel := context.WithCancel(ctx)
+	listener := make(chan compose.Event)
+	go func() {
+		var aborting bool
+		for {
+			<-listener
+			if opts.cascadeStop && !aborting {
```
gtardif, line 158: We can avoid the whole goroutine and context change if `!opts.cascadeStop`, and then maybe remove the forever loop in the goroutine ?

api/compose/api.go
```diff
@@ -175,4 +183,14 @@ type Stack struct {
 // LogConsumer is a callback to process log messages from services
 type LogConsumer interface {
 	Log(service, container, message string)
+	Status(service, container, message string)
```
gtardif, line 186: Why do we need this callback in addition to the `Listener`, that could include both exit code and some message (or build message client-side, instead of hardcoding this message server-side). 
I just need to understand this a bit more with the various constraints...

local/compose/start.go
```diff
@@ -41,8 +49,31 @@ func (s *composeService) Start(ctx context.Context, project *types.Project, cons
...
+
+	eg, ctx := errgroup.WithContext(ctx)
+	for _, c := range containers {
+		c := c
+		eg.Go(func() error {
+			statusC, errC := s.apiClient.ContainerWait(ctx, c.ID, container.WaitConditionNotRunning)
+			select {
+			case status := <-statusC:
+				service := c.Labels[serviceLabel]
+				options.Attach.Status(service, getContainerNameWithoutProject(c), fmt.Sprintf("exited with code %d", status.StatusCode))
```
gtardif, line 65: We need to filter on which specific container we want to abort everything, or is this done in the listener somewhere ?

local/compose/start.go
```diff
@@ -41,8 +49,31 @@ func (s *composeService) Start(ctx context.Context, project *types.Project, cons
...
+				if options.Listener != nil {
+					options.Listener <- compose.Event{
+						Service: service,
+						Status:  int(status.StatusCode),
+					}
+				}
+				return nil
```
gtardif, line 72: Feels strange to just ignore exit code and return nil here. (In relation to not returning errors at https://github.com/docker/compose-cli/pull/1259/files#diff-057babf47f6c9d3e25ac4ca32267631c34ac2e0b816227d7278ef89418ac4818R95, not a good feeeling about this, intuitively)

local/compose/start.go
gtardif: I guess just still WIP

api/compose/api.go
ndeloof: due to service logs coloring in `logCusumer`

local/compose/start.go
ndeloof: `--abort-on-container-exit` as defined by docker-compose applies to **any** service to exit, not a specific one. This is a potential improvement to consider for compose future.

local/compose/start.go
ndeloof: there's no error to return here. `ContainerWait`'s `errChan` is checked for API errors alrady, and here we just capture container exit Code
Same on https://github.com/docker/compose-cli/pull/1259/files#diff-057babf47f6c9d3e25ac4ca32267631c34ac2e0b816227d7278ef89418ac4818R95: attach just starts goroutines, then return

local/compose/start.go
gtardif: indeed, I mixed this with `--exit-code-from`

api/compose/api.go
gtardif: yes, makes sense

Issue #1011, opened 2020-12-04
Local compose: support more network modes


gtardif:
compose networks (internal / external) are supported, but we don't yet support all network modes

PR #1262, opened 2021-02-09
Add service based network_mode

ulyssessouza:
**What I did**
Add service based network_mode

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/1011

Depends on https://github.com/compose-spec/compose-go/pull/103

### DIFF
--- a/local/compose/create.go
+++ b/local/compose/create.go
@@ -70,6 +70,8 @@ func (s *composeService) Create(ctx context.Context, project *types.Project, opt
 	if err != nil {
 		return err
 	}
+	containerState := NewContainersState(observedState)
+	ctx = context.WithValue(ctx, ContainersKey{}, containerState)
 
 	allServices := project.AllServices()
 	allServiceNames := []string{}
@@ -92,8 +94,10 @@ func (s *composeService) Create(ctx context.Context, project *types.Project, opt
 		}
 	}
 
+	prepareNetworkMode(project)
+
 	return InDependencyOrder(ctx, project, func(c context.Context, service types.ServiceConfig) error {
-		return s.ensureService(c, observedState, project, service, opts.Recreate)
+		return s.ensureService(c, project, service, opts.Recreate)
 	})
 }
 
@@ -126,6 +130,27 @@ func prepareNetworks(project *types.Project) {
 		network.Labels = network.Labels.Add(projectLabel, project.Name)
 		network.Labels = network.Labels.Add(versionLabel, ComposeVersion)
 		project.Networks[k] = network
+	}
+}
+
+func prepareNetworkMode(p *types.Project) {
+outLoop:
+	for i := range p.Services {
+		dependency := getDependentServiceByNetwork(p.Services[i].NetworkMode)
+		if dependency == "" {
+			continue
+		}
+		if p.Services[i].DependsOn == nil {
+			p.Services[i].DependsOn = make(types.DependsOnConfig)
+		}
+		for _, service := range p.Services {
+			if service.Name == dependency {
+				p.Services[i].DependsOn[service.Name] = types.ServiceDependency{
+					Condition: types.ServiceConditionStarted,
+				}
+				continue outLoop
+			}
+		}
 	}
 }
 
@@ -235,7 +260,11 @@ func (s *composeService) getCreateOptions(ctx context.Context, p *types.Project,
 	portBindings := buildContainerPortBindingOptions(service)
 
 	resources := getDeployResources(service)
-	networkMode := getNetworkMode(p, service)
+
+	networkMode, err := getNetworkMode(ctx, p, service)
+	if err != nil {
+		return nil, nil, nil, err
+	}
 	hostConfig := container.HostConfig{
 		AutoRemove:     autoRemove,
 		Binds:          binds,
@@ -333,6 +362,14 @@ func getVolumesFrom(project *types.Project, volumesFrom []string) ([]string, []s
 	}
 	return volumes, services, nil
 
+}
+
+func getDependentServiceByNetwork(networkMode string) string {
+	baseService := ""
+	if strings.HasPrefix(networkMode, types.NetworkModeServicePrefix) {
+		return networkMode[len(types.NetworkModeServicePrefix):]
+	}
+	return baseService
 }
 
 func (s *composeService) buildContainerVolumes(ctx context.Context, p types.Project, service types.ServiceConfig,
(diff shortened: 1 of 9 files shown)


ndeloof (approved):
a few notes:
* most posix tools I know use `-o`/`--out`
* "Save to path" would make more sense to me, so that this can ECS/local -backends can sa to file, and Helm would mkdir
* I don't see reasons to hide this flag, it also make sense for other backends (defaults to stdout: `-`)

ulyssessouza:
@ndeloof Looks like you commented in the wrong issue 😸

local/compose/create.go
```diff
@@ -129,6 +135,30 @@ func prepareNetworks(project *types.Project) {
...
+			for _, service := range p.Services {
+				if service.Name == dependency {
+					p.Services[i].DependsOn[service.Name] = types.ServiceDependency{
+						Condition: types.ServiceConditionStarted,
+					}
+					continue outLoop
+				}
+			}
+
+			return fmt.Errorf("service %q not found for network_mode 'service:%s'", dependency, dependency)
```
ndeloof, line 156: such a check should take place in compose-go

local/compose/create.go
```diff
@@ -602,32 +644,48 @@ func getAliases(s types.ServiceConfig, c *types.ServiceNetworkConfig) []string {
...
 
-	// FIXME incomplete implementation
+	serviceName := ""
 	if strings.HasPrefix(mode, "service:") {
-		panic("Not yet implemented")
-	}
-	if strings.HasPrefix(mode, "container:") {
-		panic("Not yet implemented")
+		_, _ = fmt.Sscanf(mode, "service:%s", &serviceName)
+		serviceContainers, err := c.ContainerList(ctx, moby.ContainerListOptions{
```
ndeloof, line 661: seems to me the need to lookup containers while we just did by caller methods applying convergence strategy demonstrates a design issue. If we can pass moby API client, we could also pass observed state with known containers.

local/compose/create.go
```diff
@@ -335,6 +368,15 @@ func getVolumesFrom(project *types.Project, volumesFrom []string) ([]string, []s
 
 }
 
+func getNetworkFrom(networkFrom string) string {
+	baseService := ""
+	if strings.HasPrefix(networkFrom, "service:") {
```
ndeloof, line 373: `"service:"` should be declared a const in compose-go

local/compose/create.go
```diff
@@ -335,6 +368,15 @@ func getVolumesFrom(project *types.Project, volumesFrom []string) ([]string, []s
 
 }
 
+func getNetworkFrom(networkFrom string) string {
```
ndeloof, line 371: unclear naming. parameter is actualy a network mode, `networkFrom` => `networkMode` 
method name is also unclear, doesn't return a network but dependent service
=> `getDependentServiceByNetwork` or any (maybe shorter :P) name

local/compose/create.go
```diff
@@ -335,6 +368,15 @@ func getVolumesFrom(project *types.Project, volumesFrom []string) ([]string, []s
...
+	baseService := ""
+	if strings.HasPrefix(networkFrom, "service:") {
+		_, _ = fmt.Sscanf(networkFrom, "service:%s", &baseService)
```
ndeloof, line 374: use of Sscanf is overkill, can just truncate string by `[len(servicePrefix):]`

local/compose/create.go
ndeloof, line 661: better use s.getContainers which also handle oneOff flag and return a `Containers` type we can extend for a (bit more) functional design

local/compose/create.go
```diff
@@ -129,6 +135,30 @@ func prepareNetworks(project *types.Project) {
 	}
 }
 
+func prepareNetworkFrom(p *types.Project) error {
+outLoop:
+	for i := range p.Services {
+		dependency := getNetworkFrom(p.Services[i].NetworkMode)
+		if dependency != "" {
```
ndeloof, line 142: `if dependency == "" { continue }` as a bypass would avoid too much nested conditional blocks

local/compose/create.go
```diff
@@ -129,6 +135,30 @@ func prepareNetworks(project *types.Project) {
...
+			if p.Services[i].DependsOn == nil {
+				p.Services[i].DependsOn = make(types.DependsOnConfig)
+			}
+
+			for _, service := range p.Services {
```
ndeloof, line 147: can just usse `p.GetService(dependency)`

ndeloof (changes requested):

local/compose/create.go
ulyssessouza: Done

local/compose/create.go
ulyssessouza: Done

local/compose/create.go
ulyssessouza: Done

local/compose/create.go
ulyssessouza: The idea here is to change the project's collection. If I use `Project::GetService` I will set the dependency into a copy of the service, not the one attached to the project

local/compose/create.go
ulyssessouza: By this do you mean to write a method `GetCotainers` in `ServiceConfig` because I couldn't find such method

local/compose/create.go
ulyssessouza: Depends on merge of https://github.com/compose-spec/compose-go/pull/109

local/compose/create.go
ulyssessouza: Included in https://github.com/compose-spec/compose-go/pull/109

local/compose/create.go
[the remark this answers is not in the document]
ulyssessouza: Depends on https://github.com/compose-spec/compose-go/pull/109

local/compose/create.go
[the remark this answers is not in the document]
ulyssessouza: Done

local/compose/create.go
ndeloof: https://github.com/docker/compose-cli/blob/9063c138ba13d154d03d3da5b399141ea6bb944f/local/compose/containers.go#L40

ndeloof: in terms of architecture, it seems to me this should be fully avoided as the caller is `createContainer` in convergence.go which already knows the "actual state" of services containers. 

Maybe we should make `convergence` a dedicated struct, which owns `actualState` and create/recreate functions

local/compose/create.go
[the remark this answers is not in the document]
ulyssessouza: Done

ulyssessouza:
@ndeloof PTAL

ndeloof (approved):
I dislike using ctx to pass hidden parameters around, looks to me like non-assumed global variables. Let's go with it anyway, we can revisit this later

PR #1263, opened 2021-02-09
Exit code from

ndeloof:
**What I did**

Introduce "--abort-on-container-exit" and "--exit-code-from"
use ContainerWait API to capture container termination and exit code
changed architecture to pull logs and events from cli, so that we can easily react to "container exit" event

**Related issue**
replace https://github.com/docker/compose-cli/pull/1259
fix https://github.com/docker/compose-cli/issues/1209

### DIFF
--- a/api/compose/api.go
+++ b/api/compose/api.go
@@ -65,10 +65,8 @@ type CreateOptions struct {
 
 // StartOptions group options of the Start API
 type StartOptions struct {
-	// Attach will attach to container and pipe stdout/stderr to LogConsumer
-	Attach LogConsumer
-	// Listener will get notified on container events
-	Listener chan ContainerExited
+	// Attach will attach to service containers and pipe stdout/stderr to channel
+	Attach chan ContainerEvent
 }
 
 // UpOptions group options of the Up API
@@ -185,11 +183,21 @@ type Stack struct {
 // LogConsumer is a callback to process log messages from services
 type LogConsumer interface {
 	Log(service, container, message string)
-	Status(service, container, message string)
+	Status(service, container, msg string)
 }
 
-// ContainerExited let us know a Container exited
-type ContainerExited struct {
-	Service string
-	Status  int
+// ContainerEvent notify an event has been collected on Source container implementing Service
+type ContainerEvent struct {
+	Type     int
+	Source   string
+	Service  string
+	Line     string
+	ExitCode int
 }
+
+const (
+	// ContainerEventLog is a ContainerEvent of type log. Line is set
+	ContainerEventLog = iota
+	// ContainerEventExit is a ContainerEvent of type exit. ExitCode is set
+	ContainerEventExit
+)
--- a/api/compose/api.go
+++ b/api/compose/api.go
@@ -68,7 +68,7 @@ type StartOptions struct {
 	// Attach will attach to container and pipe stdout/stderr to LogConsumer
 	Attach LogConsumer
 	// Listener will get notified on container events
-	Listener Listener
+	Listener chan ContainerExited
 }
 
 // UpOptions group options of the Up API
@@ -188,11 +188,8 @@ type LogConsumer interface {
 	Status(service, container, message string)
 }
 
-// Listener get notified on container Events
-type Listener chan Event
-
-// Event let us know a Container exited
-type Event struct {
+// ContainerExited let us know a Container exited
+type ContainerExited struct {
 	Service string
 	Status  int
 }
--- a/api/compose/api.go
+++ b/api/compose/api.go
@@ -67,8 +67,8 @@ type CreateOptions struct {
 type StartOptions struct {
 	// Attach will attach to container and pipe stdout/stderr to LogConsumer
 	Attach LogConsumer
-	// CascadeStop will run `Stop` on any container exit
-	CascadeStop bool
+	// Listener will get notified on container events
+	Listener Listener
 }
 
 // UpOptions group options of the Up API
@@ -185,5 +185,14 @@ type Stack struct {
 // LogConsumer is a callback to process log messages from services
 type LogConsumer interface {
 	Log(service, container, message string)
-	Exit(service, container string, exitCode int)
+	Status(service, container, message string)
+}
+
+// Listener get notified on container Events
+type Listener chan Event
+
+// Event let us know a Container exited
+type Event struct {
+	Service string
+	Status  int
 }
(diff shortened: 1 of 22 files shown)


cli/cmd/compose/up.go
```diff
@@ -73,6 +77,12 @@ func upCommand(p *projectOptions, contextType string) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			switch contextType {
 			case store.LocalContextType, store.DefaultContextType, store.EcsLocalSimulationContextType:
+				if opts.exitCodeFrom != "" {
```
gtardif, line 80: Nit: we could remove the switch case here and validate opts regardless of context type

cli/cmd/compose/start.go
```diff
@@ -61,10 +60,28 @@ func runStart(ctx context.Context, opts startOptions, services []string) error {
...
 		return err
 	}
 
-	return c.ComposeService().Start(ctx, project, formatter.NewLogConsumer(ctx, os.Stdout))
+	queue := make(chan compose.ContainerEvent)
+	printer := printer{
+		queue: queue,
+	}
+	err = c.ComposeService().Start(ctx, project, compose.StartOptions{
+		Attach: queue,
```
gtardif, line 73: To not block gRPC interfaces from working, we need to still use a callback, and have a ContainerEventConsumer that feeds into a channel, the channel not known from the backend side. (should not be too much of a change, since we have all the bits available here)

cli/cmd/compose/up.go
```diff
@@ -145,14 +157,35 @@ func runCreateStart(ctx context.Context, opts upOptions, services []string) erro
...
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		<-signalChan
+		fmt.Println("Gracefully stopping...")
+		stopFunc() // nolint:errcheck
+	}()
+
+	err = c.ComposeService().Start(ctx, project, compose.StartOptions{
+		Attach: queue,
```
gtardif, line 181: Same here, we need to keep a callback mechanism rather than pass the channel, so we can still build gRPC interfaces and remotely invoke backends

local/e2e/compose/cascade_stop_test.go
```diff
@@ -0,0 +1,38 @@
...
+	c := NewParallelE2eCLI(t, binDir)
+
+	const projectName = "compose-e2e-logs"
+
+	res := c.RunDockerCmd("compose", "-f", "./fixtures/cascade-stop-test/compose.yaml", "--project-name", projectName, "up", "--abort-on-container-exit")
+	res.Assert(t, icmd.Expected{Out: `PING localhost (127.0.0.1)`})
+	res.Assert(t, icmd.Expected{Out: `/does_not_exist: No such file or directory`})
+	res.Assert(t, icmd.Expected{Out: `should_fail_1 exited with code 1`})
+	res.Assert(t, icmd.Expected{Out: `Aborting on container exit...`})
+	// FIXME res.Assert(t, icmd.Expected{ExitCode: 1})
```
gtardif, line 37: nit: just remove this commented line

local/e2e/compose/cascade_stop_test.go
```diff
@@ -0,0 +1,38 @@
...
+func TestCascadeStop(t *testing.T) {
+	c := NewParallelE2eCLI(t, binDir)
+
+	const projectName = "compose-e2e-logs"
+
+	res := c.RunDockerCmd("compose", "-f", "./fixtures/cascade-stop-test/compose.yaml", "--project-name", projectName, "up", "--abort-on-container-exit")
+	res.Assert(t, icmd.Expected{Out: `PING localhost (127.0.0.1)`})
+	res.Assert(t, icmd.Expected{Out: `/does_not_exist: No such file or directory`})
+	res.Assert(t, icmd.Expected{Out: `should_fail_1 exited with code 1`})
+	res.Assert(t, icmd.Expected{Out: `Aborting on container exit...`})
```
gtardif, line 36: Nit: maybe a single assert with multi-line expected string to ensure order

gtardif (changes requested):
Mainly keep a callback for containerEvents, and feed into a chan much closer to the chan consumer, on the cli side (to avoid issues with gRPC). 
(This will likely make things easier to unit tests in some places)

local/e2e/compose/cascade_stop_test.go
ndeloof: That's mostly a reminder that feature is not ready yet

api/compose/api.go
```diff
@@ -177,4 +183,24 @@ type Stack struct {
...
+
+// ContainerEventListener is a callback to process ContainerEvent from services
+type ContainerEventListener func(event ContainerEvent)
+
+// ContainerEvent notify an event has been collected on Source container implementing Service
+type ContainerEvent struct {
+	Type     int
+	Source   string
+	Service  string
+	Line     string
```
gtardif, line 197: Should we have more typed logs (stdout/err) ? (Will be useful to add timestamps also later on)

ndeloof: Would be nice, and would also address https://github.com/docker/compose/issues/8098
`stdcopy.StdCopy` we use could distinguish container's stdout vs stderr, but I'm not sure we can do the same for containers configured to set a TTY

cli/main.go
```diff
@@ -241,6 +245,11 @@ $ docker context create %s <name>`, cc.Type(), store.EcsContextType), ctype)
 }
 
 func exit(ctx string, err error, ctype string) {
+	if exit, ok := err.(cmd.ExitCodeError); ok {
+		metrics.Track(ctype, os.Args[1:], metrics.SuccessStatus)
```
gtardif, line 249: not always `metrics.SuccessStatus` here

local/compose/attach.go
```diff
@@ -24,46 +24,36 @@ import (
...
 
-func (s *composeService) attach(ctx context.Context, project *types.Project, consumer compose.LogConsumer) (*errgroup.Group, error) {
-	containers, err := s.apiClient.ContainerList(ctx, moby.ContainerListOptions{
-		Filters: filters.NewArgs(
-			projectFilter(project.Name),
-		),
-		All: true,
-	})
+func (s *composeService) attach(ctx context.Context, project *types.Project, consumer compose.ContainerEventListener) (Containers, error) {
+	containers, err := s.getContainers(ctx, project)
```
gtardif, line 34: +1, thx !

gtardif (approved):
Last thing : metrics status should depend on exit code from requested container. 
Other than this, LGTM!

cli/main.go
gtardif: Actually not, the compose command did run fine

PR #1287, opened 2021-02-12
add support for up --scale

ndeloof:
**What I did**
introduced support for --scale

### DIFF
--- a/cli/cmd/compose/up.go
+++ b/cli/cmd/compose/up.go
@@ -123,7 +123,7 @@ func runUp(ctx context.Context, opts upOptions, services []string) error {
 		return err
 	}
 
-	err = applyScale(opts.scale, project)
+	err = applyScaleOpt(opts.scale, project)
 	if err != nil {
 		return err
 	}
@@ -142,7 +142,7 @@ func runCreateStart(ctx context.Context, opts upOptions, services []string) erro
 		return err
 	}
 
-	err = applyScale(opts.scale, project)
+	err = applyScaleOpt(opts.scale, project)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


cli/cmd/compose/up.go
```diff
@@ -201,6 +215,38 @@ func runCreateStart(ctx context.Context, opts upOptions, services []string) erro
 	return err
 }
 
+func applyScale(opts []string, project *types.Project) error {
```
gtardif, line 218: unit test on this one ?

cli/cmd/compose/up.go
```diff
@@ -201,6 +215,38 @@ func runCreateStart(ctx context.Context, opts upOptions, services []string) erro
...
+				if err != nil {
+					return err
+				}
+				if service.Deploy == nil {
+					service.Deploy = &types.DeployConfig{}
+				}
+				count := uint64(replicas)
+				service.Deploy.Replicas = &count
+				project.Services[i] = service
+				continue SCALE
```
gtardif, line 242: errf... I will need to spend some time to look for a nicer solution, but ok (with some tests around it for now)

gtardif (commented):
LGTM with unit tests on applyScale()

cli/cmd/compose/up.go
ndeloof: I like using `for` labels for such scenarios. The other ways are either define sub-functions that would hide this `goto` by doing `return` inside the loop, or include a `var found bool` as a guard.

gtardif:
https://github.com/docker/compose-cli/pull/1287/checks?check_run_id=1886378763#step:9:84 should be fixed with a rebase

gtardif (approved):

Issue #1289, opened 2021-02-12
up restarting container collect logs from previous run


ndeloof:
**Description**
As I run a compose file with `compose up` and stop it by Ctrl/C, then restart I can read logs from previous run.

sample compose.yaml:
```yaml
services:
  ping:
    image: busybox:1.27.2
    command: ping localhost -c 1
```

Then run twice in detached mode, and a last time attached

**Describe the results you received:**
```
➜  docker compose -f compose.yaml up -d
The new 'docker compose' command is currently experimental. To provide feedback or request new features please open issues at https://github.com/docker/compose-cli
[+] Running 1/1
 ⠿ Container truc_ping_1  Started                                                                                                                                              0.5s
➜  docker compose -f compose.yaml up -d
The new 'docker compose' command is currently experimental. To provide feedback or request new features please open issues at https://github.com/docker/compose-cli
[+] Running 1/1
 ⠿ Container truc_ping_1  Restarted                                                                                                                                            0.4s
➜  docker compose -f compose.yaml up 
The new 'docker compose' command is currently experimental. To provide feedback or request new features please open issues at https://github.com/docker/compose-cli
[+] Running 1/1
 ⠿ Container truc_ping_1  Restarted                                                                                                                                            0.4s
Attaching to truc_ping_1
ping_1  | PING localhost (127.0.0.1): 56 data bytes
ping_1  | 64 bytes from 127.0.0.1: seq=0 ttl=64 time=0.044 ms
ping_1  | --- localhost ping statistics ---
ping_1  | 1 packets transmitted, 1 packets received, 0% packet loss
ping_1  | round-trip min/avg/max = 0.044/0.044/0.044 ms
ping_1  | PING localhost (127.0.0.1): 56 data bytes
ping_1  | 64 bytes from 127.0.0.1: seq=0 ttl=64 time=0.044 ms
ping_1  | --- localhost ping statistics ---
ping_1  | 1 packets transmitted, 1 packets received, 0% packet loss
ping_1  | round-trip min/avg/max = 0.044/0.044/0.044 ms
ping_1  | PING localhost (127.0.0.1): 56 data bytes
ping_1  | 64 bytes from 127.0.0.1: seq=0 ttl=64 time=0.043 ms
ping_1  | --- localhost ping statistics ---
ping_1  | 1 packets transmitted, 1 packets received, 0% packet loss
ping_1  | round-trip min/avg/max = 0.043/0.043/0.043 ms
truc_ping_1 exited with code 0

[...]

PR #1290, opened 2021-02-12
detect stopped containers as "Created"

ndeloof:
**What I did**
consider "exited" container from a previous run as "Created" so we just need to invoke _start_ 

**Related issue**
https://github.com/docker/compose-cli/issues/1289

### DIFF
--- a/local/compose/attach.go
+++ b/local/compose/attach.go
@@ -114,7 +114,7 @@ func (s *composeService) getContainerStreams(ctx context.Context, container moby
 			Stdin:  true,
 			Stdout: true,
 			Stderr: true,
-			Logs:   true,
+			Logs:   false,
 		})
 		if err != nil {
 			return nil, nil, err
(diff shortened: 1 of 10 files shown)


local/compose/attach.go
```diff
@@ -114,7 +114,7 @@ func (s *composeService) getContainerStreams(ctx context.Context, container moby
 			Stdin:  true,
 			Stdout: true,
 			Stderr: true,
-			Logs:   true,
+			Logs:   false,
```
gtardif, line 117: Looking back at the history of this (3714ab7d41af64f456e110e572681d570a598c1d), this was set to true because we were loosing some logs in `run` (with a race probably). Is this not needed anymore ?

ndeloof: yes indeed.
according to API docs `Logs: Replay previous logs from the container` is the main reason we get logs from previous runs as described in #1289
this option is usefull to attach to a container that's already running, but here we attach before we call `start`, so not needed

gtardif (approved):

PR #1299, opened 2021-02-15
introduce compose rm command

ndeloof:
**What I did**
Reproduced `docker-compose rm` command and flags

/test-windows

image

### DIFF
--- a/cli/cmd/compose/remove.go
+++ b/cli/cmd/compose/remove.go
@@ -18,10 +18,13 @@ package compose
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"github.com/docker/compose-cli/api/client"
 	"github.com/docker/compose-cli/api/compose"
 	"github.com/docker/compose-cli/api/progress"
+	"github.com/docker/compose-cli/utils/prompt"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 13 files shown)


cli/cmd/compose/remove.go
```diff
@@ -0,0 +1,84 @@
...
+	stop    bool
+	volumes bool
+}
+
+func removeCommand(p *projectOptions) *cobra.Command {
+	opts := removeOptions{
+		projectOptions: p,
+	}
+	cmd := &cobra.Command{
+		Use:   "rm [SERVICE...]",
```
aiordache, line 41: Could we add `remove` as alias?

ndeloof: there's no such thing as a `remove` alias on `docker rm` nor `docker-compose rm`, so to prevent heterogeneous UX I'd prefer we keep such "UX improvement" a separate debate

local/compose/remove.go
```diff
@@ -0,0 +1,83 @@
...
+		names = append(names, getCanonicalContainerName(c))
+	})
+
+	if len(stoppedContainers) == 0 {
+		fmt.Println("No stopped containers")
+		return nil
+	}
+	msg := fmt.Sprintf("Going to remove %s", strings.Join(names, ", "))
+	if options.Force {
+		fmt.Println(msg)
```
aiordache, line 54: Not sure about it but this may be the cause for the inconsistent output with the progress writer:
```
$ ./bin/docker compose rm --stop --force
[+] Running 0/1
[+] Running 1/1mpose-cli_web_1  Stopping                                                         0.3s
 ⠿ Container compose-cli_web_1  Removed                                                          0.4s
```

local/compose/remove.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		if !confirm {
+			return nil
+		}
+	}
+
+	w := progress.ContextWriter(ctx)
+	eg, ctx := errgroup.WithContext(ctx)
+	for _, c := range stoppedContainers {
+		c := c
```
aiordache, line 68: remove this?

local/compose/remove.go
ndeloof: oh yes indeed, need to get user interaction to take place before the progresswriter is started. But can't do this without breaking the backend API abstraction. Damned :'(

local/compose/remove.go
ndeloof: required. go iteration re-assigning `c` variable. such a stupid idiom (ㆆ _ ㆆ)

aiordache (approved):

Issue #1257, opened 2021-02-08
Weird logging output


ndeloof:
Reported on community slack

using 
```yaml
services:
  stderr:
    image: mcr.microsoft.com/mssql/server
```

docker-compose output:
```console
Building with native build. Learn about native build in Compose here: https://docs.docker.com/go/compose-native-build/
Recreating truc_stderr_run_b35a87e9d6aa ... done
Attaching to truc_stderr_1
stderr_1  | SQL Server 2019 will run as non-root by default.
stderr_1  | This container is running as user mssql.
stderr_1  | To learn more visit https://go.microsoft.com/fwlink/?linkid=2099216.
stderr_1  | The SQL Server End-User License Agreement (EULA) must be accepted before SQL
Server can start. The license terms for this product can be downloaded from
http://go.microsoft.com/fwlink/?LinkId=746388.
stderr_1  | 
You can accept the EULA by specifying the --accept-eula command line option,
setting the ACCEPT_EULA environment variable, or using the mssql-conf tool.
truc_stderr_1 exited with code 1
```

migrating to docker compose:
```console
The new 'docker compose' command is currently experimental. To provide feedback or request new features please open issues at https://github.com/docker/compose-cli
[+] Running 1/1
 ⠿ Container truc_stderr_1  Recreated                                                                                                                                   0.1s
Attaching to truc_stderr_1
truc_stderr_1 | SQL Server 2019 will run as non-root by default.
truc_stderr_1 | This container is running as user mssql.
truc_stderr_1 | To learn more visit https://go.microsoft.com/fwlink/?linkid=2099216.
truc_stderr_1 | The SQL 
truc_stderr_1 | S
truc_stderr_1 | e
truc_stderr_1 | r
truc_stderr_1 | v
truc_stderr_1 | e
truc_stderr_1 | r 
truc_stderr_1 | E
truc_stderr_1 | n
truc_stderr_1 | d
...
truc_stderr_1 | t
truc_stderr_1 | o
truc_stderr_1 | o
truc_stderr_1 | l
truc_stderr_1 | .
truc_stderr_1 | 
truc_stderr_1 | 

```

ndeloof:
Need to revisit logCusomer to fully replicate https://github.com/docker/compose/blob/bba8cd0322343750ae8f4aacdc46e9390a117089/compose/utils.py#L39

PR #1305, opened 2021-02-15
join log stream before we split by line

ndeloof:
**What I did**
Reproduced https://github.com/docker/compose/blob/bba8cd0322343750ae8f4aacdc46e9390a117089/compose/utils.py#L32 logic to first join log stream then split by `\n` to produce the compose log events.

includes test case

**Related issue**
https://github.com/docker/compose-cli/issues/1257

image


local/compose/logs.go
```diff
@@ -88,31 +88,41 @@ func (s *composeService) Logs(ctx context.Context, projectName string, consumer
...
 
 type splitBuffer struct {
+	buffer    bytes.Buffer
 	service   string
 	container string
 	consumer  compose.ContainerEventListener
 }
 
 // getWriter creates a io.Writer that will actually split by line and format by LogConsumer
 func getWriter(service, container string, events compose.ContainerEventListener) io.Writer {
```
aiordache, line 98: I moved this method and struct to utils/logconsumer.go as I needed it in the kube backend too(to avoid having duplicated code). Not sure why this code was back here. 

And above we call utils.GetWriter.

ndeloof: https://github.com/docker/compose-cli/issues/1311

aiordache: Should this method be called on line 78 above?

aiordache (approved):

PR #1313, opened 2021-02-16
introduce --quiet option

### LINKED ISSUE #1280: Support compose build --quiet
```
  -q, --quiet             Don't print anything to STDOUT

```
Also add the same flag to `config` and `pull` 
(note `docker-compose up` has `--quiet-pull` to disable output only related to pulling images)
We should also add this to the `up` command (but still attach to logs of requested)

ndeloof:
**What I did**
introduce --quiet option

**Related issue**
close #1280

image

### DIFF
--- a/cli/cmd/compose/convert.go
+++ b/cli/cmd/compose/convert.go
@@ -89,6 +89,11 @@ func runConvert(ctx context.Context, opts convertOptions, services []string) err
 	if err != nil {
 		return err
 	}
+
+	if opts.quiet {
+		return nil
+	}
+
 	var out io.Writer = os.Stdout
 	if opts.Output != "" {
 		file, err := os.Create(opts.Output)
--- a/cli/cmd/compose/convert.go
+++ b/cli/cmd/compose/convert.go
@@ -17,8 +17,11 @@
 package compose
 
 import (
+	"bufio"
 	"context"
 	"fmt"
+	"io"
+	"os"
 
 	"github.com/docker/compose-cli/api/compose"
 
@@ -31,6 +34,7 @@ type convertOptions struct {
 	*projectOptions
 	Format string
 	Output string
+	quiet  bool
 }
 
 var addFlagsFuncs []func(cmd *cobra.Command, opts *convertOptions)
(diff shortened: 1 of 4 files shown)


cli/cmd/compose/convert.go
```diff
@@ -39,22 +43,33 @@ func convertCommand(p *projectOptions) *cobra.Command {
...
 		projectOptions: p,
 	}
-	convertCmd := &cobra.Command{
+	cmd := &cobra.Command{
 		Aliases: []string{"config"},
 		Use:     "convert SERVICES",
 		Short:   "Converts the compose file to platform's canonical format",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if opts.quiet {
+				if opts.quiet {
```
chris-crone, line 52: Guessing these shouldn't both be `opts.quiet`?

aci/compose.go
```diff
@@ -100,7 +101,7 @@ func (cs aciComposeService) warnKeepVolumeOnDown(ctx context.Context, projectNam
 		if v.AzureFile == nil || v.AzureFile.StorageAccountName == nil || v.AzureFile.ShareName == nil {
 			continue
 		}
-		fmt.Printf("WARNING: fileshare \"%s/%s\" will NOT be deleted. Use 'docker volume rm' if you want to delete this volume\n",
+		fmt.Fprintf(os.Stderr, "WARNING: fileshare \"%s/%s\" will NOT be deleted. Use 'docker volume rm' if you want to delete this volume\n",
```
chris-crone, line 104: Do we still want to print out the stderr contents if we're in quiet mode? What does docker/cli do?

cc @thaJeztah

chris-crone (changes requested):

cli/cmd/compose/convert.go
ndeloof: not sure what I had in mond dogin this ( ဖ ͜ စ)

aci/compose.go
ndeloof: good point

cli/cmd/compose/convert.go
chris-crone: Making _really_ sure it's enabled :)

aci/compose.go
thaJeztah: Hm... will have to check. Not sure what would be most expected. 🤔. I guess it's a "warning", so not a _critical_ message, but not sure how important it is to let the user know

aiordache (approved):

cli/cmd/compose/convert.go
```diff
@@ -76,9 +89,14 @@ func runConvert(ctx context.Context, opts convertOptions, services []string) err
 	if err != nil {
 		return err
 	}
+	var out io.Writer = os.Stdout
 	if opts.Output != "" {
```
chris-crone, line 93: Should this still work with quiet mode? The help says that it only does validation

chris-crone (approved):
LGTM with one question needing an answer

cli/cmd/compose/convert.go
ndeloof: Indeed. --ouput might be used with a path and then we would override file, and not just "validate and quit".
fixed

Issue #1310, opened 2021-02-16
Unable to deploy simple example to ECS


metasoarous:
Hello; Thanks for your time.

**Description**

Over at #1308 I mentioned that I was able to get `docker compose` working locally using the example here: https://docs.docker.com/compose/gettingstarted/. However, when I try to witch contexts to a freshly minted `ecs` context, it fails to deploy.

**Steps to reproduce the issue:**
1. `docker context create ecs myecscontext`
2. go through setup in https://docs.docker.com/compose/gettingstarted/
3. run `docker compose up` on default context to ensure it works
4. run `docker context use myecscontext` to switch to our new context
5. run `docker compose up` to deploy to new context

**Describe the results you received:**

```
% docker compose up
redis:alpine resolved to docker.io/library/redis:alpine@sha256:d2b1b0bf521f55381c81b1bc0218aec4ddc92fb545e9e8492e8ff324462eed58
invalid reference format
```

**Describe the results you expected:**

Expected the app to deploy under my credentials.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
% docker version                                                :( 1 21-02-15 - 15:48:53
Client: Docker Engine - Community
 Cloud integration: 1.0.7
 Version:           20.10.3
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        48d30b5
 Built:             Fri Jan 29 14:33:25 2021
 OS/Arch:           linux/amd64
 Context:           polis-preprod
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.3
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       46229ca
  Built:            Fri Jan 29 14:31:35 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```
myecscontext
```

**Output of `docker info`:**

```
% docker info                                                        21-02-15 - 16:15:18
Client:
 Context:    default
 Debug Mode: false
[...]

PR #1316, opened 2021-02-16
let user know service require image attribute

ndeloof:
**What I did**
Don't try to resolve image digest for services without an image attribute

**Related issue**
#1310

/test-ecs


ndeloof:
already included in https://github.com/compose-spec/compose-go/pull/104/files#diff-738ef8ba46c21b0df2d53bc70d0250e4a85dde5efa730e06be0bb843c9be7f47R252

Issue #1268, opened 2021-02-10
Support compose down --volumes


gtardif:
help from `docker-compose down`: 
```
 -v, --volumes           Remove named volumes declared in the `volumes`
                            section of the Compose file and anonymous volumes
                            attached to containers.
```

PR #1317, opened 2021-02-17
introduce --rmi and --volumes options

ndeloof:
**What I did**
introduce --rmi and --volumes options on compose down

**Related issue**
https://github.com/docker/compose-cli/issues/1268
https://github.com/docker/compose-cli/issues/1269


princemaple:
Has this been released? Still seeing it on docker desktop 3.3.3

ndeloof:
`--volumes` is supported since 1.0.15, soon to be bundled by Docker Desktop. Alternatively you can install docker-compose CLI plugin from this repository's release tar.gz

princemaple:
Thanks.

Issue #1286, opened 2021-02-12
Support compose config --resolve-image-digests


ndeloof:
```
--resolve-image-digests  Pin image tags to digests.
```

Note: code is already available in ECS backend: https://github.com/docker/compose-cli/blob/b39208ec9281e91d4f66f05b5a153dfd19f83c69/ecs/cloudformation.go#L106

ndeloof:
initial step here: https://github.com/compose-spec/compose-go/pull/104

PR #1318, opened 2021-02-17
introduce --resolve-image-digests on config command

ndeloof:
**What I did**
introduce --resolve-image-digest on `compose config` command based on compose-go ResolveImage

**Related issue**
close https://github.com/docker/compose-cli/issues/1286


cli/cmd/compose/convert.go
```diff
@@ -23,18 +23,23 @@ import (
 	"io"
 	"os"
 
-	"github.com/docker/compose-cli/api/compose"
-
+	"github.com/cnabio/cnab-to-oci/remotes"
+	"github.com/distribution/distribution/v3/reference"
+	cliconfig "github.com/docker/cli/cli/config"
+	"github.com/docker/compose-cli/api/config"
```
chris-crone, line 29: This should be in the other import section

ecs/cloudformation.go
```diff
@@ -19,6 +19,8 @@ package ecs
 import (
 	"context"
 	"fmt"
+	reference2 "github.com/distribution/distribution/v3/reference"
+	"github.com/opencontainers/go-digest"
```
chris-crone, line 23: Should be in the lower import section

ecs/cloudformation.go
```diff
@@ -19,6 +19,8 @@ package ecs
 import (
 	"context"
 	"fmt"
+	reference2 "github.com/distribution/distribution/v3/reference"
```
chris-crone, line 22: Don't need a `reference2` here.

chris-crone (approved):
Minor fixes then LGTM

PR #1323, opened 2021-02-18
introduce logs --timestamp

ndeloof:
**What I did**
Introduced `--timestamps` option on `compose logs`

**Related issue**
https://github.com/docker/compose-cli/issues/1272


chris-crone (approved):
LGTM

Issue #1108, opened 2021-01-06
(feat): Suppress warn message


lukasmrtvy:
Would be nice to be able somehow suppress warn message: `The new 'docker compose' command is currently experimental. To provide feedback or request new features please open issues at https://github.com/docker/compose-cli` for better machine parsing..
Thanks

ndeloof:
IMHO it would make sense indeed for user to has some acknowledge mechanism so they don't get this warning, comparable to CLI plugins enablement. Maybe we could skip this message when some environment variable is set?
cc @nebuk89

chris-crone:
+1 We could also not show it after x runs or something.

thaJeztah:
We could have a look at reinstating the global `--log-level` flag that's present on the `docker` CLI (and add it to the `~/.docker/config.json`);

```
  -l, --log-level string   Set the logging level ("debug"|"info"|"warn"|"error"|"fatal") (default "info")
```

That option was mostly inherited from when the CLI and Daemon were a single binary (many/most parts of the cli currently ignore it, but perhaps it's a good one for this?)

chris-crone:
Yep, that's an option too.

We should also decide what is required for us to be comfortable removing the warning all together. I think everyone wants this to just be GA :)

gabrieldemarmiesse:
One easy solution to allow parsing of outputs would be to print this warning to stderr instead of stdout (it's currently printed to stdout). 
```
$ docker compose -f dummy_compose.yml  ps --quiet 2>/dev/null
The new 'docker compose' command is currently experimental. To provide feedback or request new features please open issues at https://github.com/docker/compose-cli
809a979544d62b8777823baadbd1424a4be5f2ccf20641cc01ec7439ea459b39
2b109f69822e570f024f4419140a6817b08f0a43cf432323396834341914767e
```

As far as I understand @lukasmrtvy  just wants to parse stdout. It's easier to redirect the warning to stderr than to make a new option in the config.json or moving out of experimental.

chris-crone:
That's a good idea @gabrieldemarmiesse. We [write warnings to stderr on the docker/cli](https://github.com/docker/cli/blob/1e54c5d67c536ff4e6cf3efa5aa66999c0d4e11f/cli/config/config.go#L141).

thaJeztah:
Yes, they should be on stderr; see https://github.com/docker/cli/pull/1859

chris-crone:
Let's aim to have this for the next release:
- stderr instead of stdout
[...]

PR #1339, opened 2021-02-23
only display experimental warning on command error

ndeloof:
**What I did**
only display experimental warning when compose command fails

note: cobra directly return on error, PersistentPostRun is not executed, so I eventually used a global variable :'(

**Related issue**
https://github.com/docker/compose-cli/issues/1108

### DIFF
--- a/cli/cmd/compose/compose.go
+++ b/cli/cmd/compose/compose.go
@@ -17,8 +17,6 @@
 package compose
 
 import (
-	"fmt"
-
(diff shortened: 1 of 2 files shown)


gtardif (approved):
LGTM. 
I noticed that running `docker compose up -f titi` did output only `unknown shorthand flag: 'f' in -f`, without the experimental warning, but this was already the case before: cobra fails parsing flags before running he compose preRun. 
We should merge this PR and check if we can improve the flag error scenario

Issue #1291, opened 2021-02-12
Support compose up --no-build


gtardif:
` --no-build Don't build an image, even if it's missing.` 
Same option on `up` and `create`

PR #1340, opened 2021-02-23
introduce --no-build option

ndeloof:
**What I did**
Introduced --no-build option to disable building images from source

**Related issue**
https://github.com/docker/compose-cli/issues/1291

### DIFF
--- a/cli/cmd/compose/create.go
+++ b/cli/cmd/compose/create.go
@@ -17,6 +17,8 @@
 package compose
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


cli/cmd/compose/up.go
```diff
@@ -284,10 +289,18 @@ func setup(ctx context.Context, opts composeOptions, services []string) (*client
 		project.Services[0].DomainName = opts.DomainName
 	}
 	if opts.Build {
-		for _, service := range project.Services {
+		for i, service := range project.Services {
 			service.PullPolicy = types.PullPolicyBuild
+			project.Services[i] = service
```
gtardif, line 294: did we have a bug here not reaffecting modified service ?

cli/cmd/compose/create.go
```diff
@@ -34,10 +35,14 @@ func createCommand(p *projectOptions) *cobra.Command {
 		Use:   "create [SERVICE...]",
 		Short: "Creates containers for a service.",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if opts.Build && opts.noBuild {
+				return fmt.Errorf("--build and --no-build are incompatible")
+			}
```
gtardif, line 40: Nit: we should also add a check for force-recreate and no-recreate, next to this one. (Check already exists in the up command)

gtardif (approved):
LGTM, small improvement on missing check for force-recreate / no-recreate flag next to no-build

cli/cmd/compose/up.go
ndeloof: seems so

PR #1341, opened 2021-02-23
introduce a few more `compose run` options

ndeloof:
**What I did**
extended compose run with a few more options

*  --name NAME Assign a name to the container
* --entrypoint CMD Override the entrypoint of the image.
* -e KEY=VAL Set an environment variable (can be used multiple times)
* -l, --label KEY=VAL Add or override a label (can be used multiple times)
* -u, --user="" Run as specified username or uid
* -T Disable pseudo-tty allocation. By default docker-compose run allocates a TTY.
* -w, --workdir="" Working directory inside the container

### DIFF
--- a/api/compose/api.go
+++ b/api/compose/api.go
@@ -19,6 +19,7 @@ package compose
 import (
 	"context"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/compose-spec/compose-go/types"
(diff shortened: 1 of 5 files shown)


api/compose/api.go
```diff
@@ -130,20 +131,38 @@ type RemoveOptions struct {
...
+		parts := strings.SplitN(s, "=", 2)
+		key := parts[0]
+		switch {
+		case len(parts) == 1:
+			environment[key] = nil
+		default:
+			environment[key] = &parts[1]
+		}
+	}
+	return environment
```
gtardif, line 165: some unit test on this maybe ?

gtardif (approved):
LGTM, with a small unit test added

PR #1343, opened 2021-02-23
add --timeout option to compose up

ndeloof:
**What I did**
added `--timeout` option to `docker compose up` by overriding `StopGracePeriod` in compose model

added `--no-deps` option to `docker compose up` by marking all dependent services as "disabled"

### DIFF
--- a/cli/cmd/compose/up.go
+++ b/cli/cmd/compose/up.go
@@ -25,6 +25,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/docker/compose-cli/api/client"
 	"github.com/docker/compose-cli/api/compose"
@@ -60,6 +61,8 @@ type upOptions struct {
 	scale         []string
 	noColor       bool
 	noPrefix      bool
+	timeChanged   bool
+	timeout       int
 }
 
 func (o upOptions) recreateStrategy() string {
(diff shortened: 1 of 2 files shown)


cli/cmd/compose/up.go
```diff
@@ -117,6 +122,8 @@ func upCommand(p *projectOptions, contextType string) *cobra.Command {
 		flags.BoolVar(&opts.noStart, "no-start", false, "Don't start the services after creating them.")
 		flags.BoolVar(&opts.cascadeStop, "abort-on-container-exit", false, "Stops all containers if any container was stopped. Incompatible with -d")
 		flags.StringVar(&opts.exitCodeFrom, "exit-code-from", "", "Return the exit code of the selected service container. Implies --abort-on-container-exit")
+		flags.IntVarP(&opts.timeout, "timeout", "t", 10, "Use this timeout in seconds for container shutdown when attached or when containers are already running.")
```
ulyssessouza, line 125: IMO `time.Duration` would be better here. So you can let the user use `1m`, `1s` and so on.

ndeloof: this aligns with Docker CLI and other places where we defined timeout
I'd prefer we focus on docker-compose parity before we debate UX improvements :-P

ulyssessouza (approved):
LGTM

PR #1346, opened 2021-02-23
Update all headers and its checkers for year 2021

ulyssessouza:
**What I did**
Update all files with:
```sh
$ find . -type f \( -name "*Dockerfile" -o -name "*Makefile" -o -name "*.go" -o -name "*.sh" \) | xargs sed -i 's/Copyright 2020 Docker Compose CLI authors/Copyright 2021 Docker Compose CLI authors/g'
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:experimental
 
 
-#   Copyright 2020 Docker Compose CLI authors
+#   Copyright 2021 Docker Compose CLI authors
 
 #   Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-#   Copyright 2020 Docker Compose CLI authors
+#   Copyright 2021 Docker Compose CLI authors
 
 #   Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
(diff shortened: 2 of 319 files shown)


gtardif:
is there a valid form that does not require updating all files every year ? 
Something like "since 2020" ?

ndeloof:
Copyright date should reflect the time work took place on code, IMHO it doesn't make sense to update all files every year. Only new components/packages may use a refreshed copyright header to reflect work in progress.
That being said, this whole copyright headers are a non-sense to me as LICENSE is already set on repo, but INAL.

thaJeztah:
Yes, changing 2020 to 2021 is definitely wrong. 

Copyright should either:

- "not" include a year
- include Year of first publication (2020 for this code?)
- only update year if significant changes are made to that file (in which case a range should be used `<year of first publication>` - `<year of last significant change>`)

Some projects also choose to omit the Year (as, especially with Git), year of "first publication" can be proven based on git history (e.g. https://github.com/containerd/containerd/blob/master/client.go#L2)
That said, I know there are some in the "camp" that including a Year makes it slightly "stronger"

gtardif:
cf https://docker.slack.com/archives/C0111G2RCE4/p1614177320044400

ndeloof (changes requested):
Copyright header should reflect the year code has been created, not legit to update it as a whole each and every year

gtardif:
Closing for now as the change will not be updating year in copyright anyway. Need to confirm if we can remove year from all files and only keep it in the NOTICE at repo root

Issue #1279, opened 2021-02-10
Support compose common flag  --project-directory


gtardif:
```
 --project-directory PATH    Specify an alternate working directory
                              (default: the path of the Compose file)
```

To be linked with --workdir we already have (alias ? ) 
⚠️ `compose run`and `compose exec` also have a `--workdir` subcommand flag...

PR #1353, opened 2021-02-25
Fix --project-directory mix with --workdir

ulyssessouza:
This fixes the ambiguity between `--project-directory` and `--workdir`

`--project-directory` is the one passed to `compose-go` to resolve compose files
`--workdir` is passed to commands `run` and `exec` to provide $PWD to the container before running the command.

Resolves https://github.com/docker/compose-cli/issues/1279

### DIFF
--- a/cli/cmd/compose/compose.go
+++ b/cli/cmd/compose/compose.go
@@ -32,7 +32,7 @@ type projectOptions struct {
 	ProjectName string
 	Profiles    []string
 	ConfigPaths []string
-	WorkingDir  string
+	ProjectDir  string
 	EnvFile     string
 }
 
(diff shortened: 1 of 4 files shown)


cli/cmd/compose/compose.go
```diff
@@ -41,8 +41,7 @@ func (o *projectOptions) addProjectFlags(f *pflag.FlagSet) {
 	f.StringVarP(&o.ProjectName, "project-name", "p", "", "Project name")
 	f.StringArrayVarP(&o.ConfigPaths, "file", "f", []string{}, "Compose configuration files")
 	f.StringVar(&o.EnvFile, "env-file", "", "Specify an alternate environment file.")
-	f.StringVar(&o.WorkingDir, "workdir", "", "Specify an alternate working directory")
-	// TODO make --project-directory an alias
+	f.StringVar(&o.ProjectDir, "project-directory", "", "Specify an alternate working directory\n(default: the path of the Compose file)")
```
gtardif, line 44: unfortunately we can't just break existing usage on ACI/ECS like this I think, we'll need to keep workdir an alias of this flag (confusing with the exec one I agree, but... we can put project-directory in front more visible of course, just support running with the --workdir alias)

ulyssessouza: I understand it's a breaking change, but that's actually a bug fix.
The problem is that in the case of `exec` and `run` we can use both at the same time. So in this case it's broken.

ndeloof (changes requested):
Don't remove `--workdir` or we will break ECS/ACI use (whihc is G.A., we can't just change our mind like this). Make it a hidden flag, warn user about deprecated attribute and use of `--project-directory`

ulyssessouza:
@PTAL @ndeloof @gtardif

ndeloof (approved):

Issue #1325, opened 2021-02-18
compose CLI ignores platform value from compose spec


BretFisher:
**Description**

On my Apple M1 Mac, I noticed that the new `docker compose pull` ignores the `platform:` key for services. If I specify a platform in YAML and try a pull, it will still try to pull the local platform's image. I wanted to get mysql working on a arm64 machine by pulling the amd64 image.

The old `docker-compose pull` works correctly for pulling different platform images.

**Steps to reproduce the issue:**
1. Create a compose file with the following:

```yaml
services:
  db:
    image: mysql
    platform: linux/amd64
```

2. Run a `docker-compose pull` and see it pull the correct x86-64 image.

3. Run a `docker compose pull` and it will error with `no matching manifest for linux/arm64/v8 in the manifest list entries`.

**Output of `docker version`:**

Docker Desktop M1 Preview 3.1.0 (60984)

```
Client/Server: 20.10.3
```

PR #1354, opened 2021-02-25
Take `platform` in account on pulling and building images

ulyssessouza:
Resolves https://github.com/docker/compose-cli/issues/1325

### DIFF
--- a/local/compose/build.go
+++ b/local/compose/build.go
@@ -26,12 +26,14 @@ import (
 	"github.com/docker/compose-cli/api/compose"
 
 	"github.com/compose-spec/compose-go/types"
+	"github.com/containerd/containerd/platforms"
 	"github.com/docker/buildx/build"
 	"github.com/docker/buildx/driver"
 	_ "github.com/docker/buildx/driver/docker" // required to get default driver registered
 	"github.com/docker/buildx/util/progress"
 	"github.com/docker/docker/errdefs"
 	bclient "github.com/moby/buildkit/client"
+	specs "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 func (s *composeService) Build(ctx context.Context, project *types.Project, options compose.BuildOptions) error {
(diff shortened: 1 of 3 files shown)


gtardif (changes requested):
Change on pull LGTM. 
You'll also need a `FROM --platform xxx` at https://github.com/docker/compose-cli/blob/main/local/compose/build.go#L83, because we use buildkit to pull images when doing compose up.

gtardif:
Adding Chris as reviewer for some M1 double checking. (Did some tests yesterday with `FROM --platform`, should be all fine)

local/compose/build.go
```diff
@@ -161,7 +176,37 @@ func (s *composeService) toBuildOptions(service types.ServiceConfig, contextPath
...
+		}, nil
+	}
+	tokens := strings.Split(p, "/")
+	if len(tokens) != 2 {
+		return specs.Platform{}, errdefs.InvalidParameter(errors.New("could not parse platform " + p))
 	}
+	return specs.Platform{
+		OS:           tokens[0],
+		Architecture: tokens[1],
+	}, nil
```
ndeloof, line 209: for consistency, this should be implemented as a `Transform` func by compose-go, like other "short syntax" strings vs dict.

local/compose/build.go
```diff
@@ -161,7 +176,37 @@ func (s *composeService) toBuildOptions(service types.ServiceConfig, contextPath
...
+		platforms = append(platforms, platform)
+	}
+	return platforms, nil
+}
+
+func getPlatform(p string) (specs.Platform, error) {
+	if p == "" {
+		return specs.Platform{
+			OS:           runtime.GOOS,
+			Architecture: runtime.GOARCH,
```
ulyssessouza, line 199: For this lines, I've tested on Linux and that works. But something tells me that it doesn't work for MacOS's case, since the CLI runs on the `darwin` and the actual image will run on a `linux`.
@chris-crone Could you confirm that?

chris-crone: `runtime.{GOOS,GOARCH}` isn't safe to use on macOS M1 machines as the binary could be being emulated (see: [go-rosetta](https://github.com/tonistiigi/go-rosetta))

We should look at what's done in docker/cli and on moby/moby to see how this is handled

local/compose/build.go
ulyssessouza: I could actually remove this, since it's already implemented in `containerd`.
PTAL

gtardif (approved):
LGTM. Could not test on a M1 mac, just tried to specify platforms in compose file and inspect resulting images (pull & build)

gtardif (changes requested):
Ah, we also need to use the specified platform when running the containers, at https://github.com/docker/compose-cli/blob/main/local/compose/convergence.go#L281 (we actually specify the platform for local `docker run` command, we just ignored it in local compose

ndeloof (approved):

gtardif (approved):

Issue #1338, opened 2021-02-22
Command-line completion?


sbellem:
I just wonder if there's a command completion like there is for the [docker-compose](https://docs.docker.com/compose/completion/) command.

ndeloof:
Should be easy to generate one, as Cobra command library includes support for generating completion scripts https://github.com/spf13/cobra/blob/master/shell_completions.md

mat007:
Are we looking to reach parity with docker-compose or is this not a requirement?

glenjamin:
In the mean time, does anyone know if there's a way to configure `docker compose` completion to fall back to the old `docker-compose` completion?

PR #1355, opened 2021-02-25
Add completion to shells

ulyssessouza:
This provides shell completion for `bash`, `fish`, `zsh` and `Powershell`

Resolves https://github.com/docker/compose-cli/issues/1338

### DIFF
--- a/cli/cmd/compose/completion.go
+++ b/cli/cmd/compose/completion.go
@@ -1,0 +1,86 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+func completionCommand() *cobra.Command {
+	return &cobra.Command{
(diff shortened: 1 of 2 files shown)


ndeloof:
How to use completion script generated for a subcommand?
How to combine this with docker/cli completion?

cli/cmd/compose/completion.go
```diff
@@ -0,0 +1,86 @@
...
+  $ docker compose completion fish > ~/.config/fish/completions/docker_compose.fish
+
+PowerShell:
+
+  PS> docker compose completion powershell | Out-String | Invoke-Expression
+
+  # To load completions for every new session, run:
+  PS> docker compose completion powershell > docker_compose.ps1
+  # and source this file from your PowerShell profile.
+`,
```
ulyssessouza, line 67: @ndeloof 
> How to use completion script generated for a subcommand?
This comes in the help by running `docker compose completion --help`.

> How to combine this with docker/cli completion?
I'm not that sure if I understood well. Do we need to combine that?

thaJeztah:
Wondering if instead of a subcommand, we should use a `--completion` flag (`--completion=bash`); this command is very unlikely to be used by the user itself (more a "set and forget" options somewhere in a `.profile`), but being a `command` could by itself make it end up in completion suggestions

gtardif:
As discussed previously with @ulyssessouza , we need to find a better way to integrate with completion scripts shipped / installed with Docker Desktop. (Desktop actually includes completion scripts for the docker CLI and for `docker-compose`)

Completion scripts for the docker CLI include many examples of completion for subcommands, we can adapt this. 
for the cli & docker-compose, completion scripts are packages next to the binary and Desktop re-packages this for it's install)

ndeloof (changes requested):
need to clarify how this would impact the docker CLI completion scripts

ulyssessouza:
PTAL at https://github.com/docker/compose-cli/pull/1384

This alternative PR builds an external binary to generate the completion files. This avoid polluting the "compose business" with technical stuff.
It also integrates the generation to the normal build so that when running `make` it also generates the completion files bin the `bin` directory side-by-side with the binary.

Issue #1361, opened 2021-03-01
Compose : support compose pull --include-deps


gtardif:
compose pull option: 
`--include-deps Also pull services declared as dependencies`

PR #1377, opened 2021-03-02
introduce pull --include-deps

ndeloof:
**What I did**
introduce pull --include-deps
disable all dependency services so that they are ignored by Pull

**Related issue**
https://github.com/docker/compose-cli/issues/1361

### DIFF
--- a/cli/cmd/compose/pull.go
+++ b/cli/cmd/compose/pull.go
@@ -64,7 +64,11 @@ func runPull(ctx context.Context, opts pullOptions, services []string) error {
 		if err != nil {
 			return err
 		}
-		project.DisabledServices = append(project.DisabledServices, project.Services...)
+		for _, s := range project.Services {
+			if !contains(services, s.Name) {
+				project.DisabledServices = append(project.DisabledServices, s)
+			}
+		}
 		project.Services = enabled
 	}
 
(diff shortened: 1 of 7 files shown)


cli/cmd/compose/pull.go
```diff
@@ -57,6 +59,15 @@ func runPull(ctx context.Context, opts pullOptions, services []string) error {
 		return err
 	}
 
+	if !opts.includeDeps {
+		enabled, err := project.GetServices(services...)
+		if err != nil {
+			return err
+		}
+		project.DisabledServices = append(project.DisabledServices, project.Services...)
```
gtardif, line 67: DisabledServices will contain all services including the enabled ones here. Not affecting the current pull execution path probably, but very confusing, we should remove enabled services from there

gtardif (changes requested):
LGTM except filling DisabledServices with some that are actually enabled

cli/cmd/compose/pull.go
ndeloof: ok, I'll fix the other places we do the same for "noDeps"

ndeloof: see https://github.com/docker/compose-cli/pull/1377/commits/42d1dd2e0d0d7ce3d77e384445b79eb93905fda7

gtardif (approved):

PR #1384, opened 2021-03-03
Generated completion

ulyssessouza:
Resolves https://github.com/docker/compose-cli/issues/1338

This generates the four supported completions along with the `compose-cli` binary in the `bin` directory.


Makefile
```diff
@@ -37,13 +37,18 @@ protos: ## Generate go code from .proto files
...
+cli: completion ## Compile the cli
 	@docker build . --target cli \
 	--platform local \
 	--build-arg BUILD_TAGS=e2e,kube \
 	--build-arg GIT_TAG=$(GIT_TAG) \
 	--output ./bin
 
+completion: ## Compile completion binary to generate completion scripts
+	@docker build . --target completion \
+	--platform local \
```
gtardif, line 49: for completion generation, --platform is useless

gtardif (commented):
Actually, in order to have meaningful completion, we'll need to have specific completion logic for each compose subcommand, checking current docker-compose completion, and adapting that to match `docker compose` command, as is done for the docker cli subcommands. 
(Just generating completion from the cobra command provides quite low added value)

ndeloof (approved):

Makefile
ulyssessouza: I actually realized that building the completion binary is not necessary.
This artifact is of no servitude. A simple `go run` to generate the completions is the simplest  solution to this avoiding the whole docker build and platform related things.

ulyssessouza:
@ndeloof @gtardif PTAL

Now this can be tested with
```
$ make -f builder.Makefile package
```

The resulting tarballs will be placed in the `dist` directory with it's respective completion files inside.

ebriney:
Screenshot 2021-03-04 at 10 11 13

It's quite similar names then what we have for python version, but except that, it will be easy to integrate.
On windows we don't embed completions.

ndeloof:
obsolete

xnaas:
In what way is this obsolete? Was bash completion solved somewhere? Is it coming in a `docker-ce` or `docker-ce-cli` package update or something?

Issue #1368, opened 2021-03-01
Support compose up --quiet-pull


gtardif:
compose up option: 
`--quiet-pull               Pull without printing progress information`

ndeloof:
not sure how we can tweak buildx to be quiet pulling images but still report build progress

PR #1390, opened 2021-03-04
Add `quietPull` option to `up`

ulyssessouza:
**What I did**
Add `quietPull` option to `up`

Resolves https://github.com/docker/compose-cli/issues/1368
Depends on https://github.com/docker/buildx/pull/558

### DIFF
--- a/local/compose/build.go
+++ b/local/compose/build.go
@@ -56,7 +56,7 @@ func (s *composeService) Build(ctx context.Context, project *types.Project, opti
 	return err
 }
 
-func (s *composeService) ensureImagesExists(ctx context.Context, project *types.Project) error {
+func (s *composeService) ensureImagesExists(ctx context.Context, project *types.Project, quietPull bool) error {
 	opts := map[string]build.Options{}
 	imagesToBuild := []string{}
 	for _, service := range project.Services {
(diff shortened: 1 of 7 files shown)


local/compose/build.go
```diff
@@ -98,7 +98,12 @@ func (s *composeService) ensureImagesExists(ctx context.Context, project *types.
 
 	}
 
-	err := s.build(ctx, project, opts, "auto")
+	mode := "auto"
+	if quietPull {
+		mode = "quiet"
```
ulyssessouza, line 103: Little hack to be checked in https://github.com/docker/compose-cli/pull/1390/files#diff-477fda25fb05d92fe2c8e3cb31fdc228a8f5597b9e26b22f7feeff8246404db7R146-R156 to silence the `pull`

local/compose/build.go
```diff
@@ -98,7 +98,12 @@ func (s *composeService) ensureImagesExists(ctx context.Context, project *types.
...
+	}
+
+	err := s.build(ctx, project, opts, mode)
```
gtardif, line 106: I would not mix mode (auto/tty/...) with this pull-specific option. This will lead to unwanted side effects or hard to understand/evolve code. I'd prefer to have a separate param here, a lot more explicit

gtardif (changes requested):
Just add more explicit param to service.build rather than mixing behaviour between tty display and quiet pull option, otherwise LGTM

local/compose/build.go
ndeloof: I understand your PoV but on the other hand, as `mode` is responsible to manage output, using `quiet` here sounds legitimate. Maybe we could contribute this special value to https://github.com/docker/buildx/blob/080e9981c7332513158f0deff7d4057c370395d7/util/progress/printer.go ?

ulyssessouza: @ndeloof PTL https://github.com/docker/buildx/pull/558

gtardif: 👍

gtardif:
This is waiting on https://github.com/docker/buildx/pull/558

gtardif (approved):

Issue #1283, opened 2021-02-11
Docker-compose command compatibility list


gtardif:
Check list of commands and flags, from docker-compose help

- [x]   build              Build or rebuild services
  - [x]   #1360 --build-arg key=val     Set build-time variables for services.
  - [x]   #1440 --compress              Compress the build context using gzip
  - [x]   #1440 --force-rm              Always remove intermediate containers.
  - [x]   #1468 -m, --memory MEM        Set memory limit for the build container. - not yet supported in buildkit
  - [x]   #1468 --no-cache              Do not use cache when building the image.
  - [x]   #1468 --no-rm                 Do not remove intermediate containers after a successful build.
  - [x]  #1440 --parallel              Build images in parallel.
  - [x]  #1358 --progress string       Set type of progress output (auto, plain, tty).
  - [x]  #1359 --pull                  Always attempt to pull a newer version of the image.
  - [x] #1280  -q, --quiet             Don't print anything to STDOUT

- [x]   config             Validate and view the Compose file
  - [x] #1286  --resolve-image-digests  Pin image tags to digests.
  - [x]   --no-interpolate         Don't interpolate environment variables.
  - [x] #1280  -q, --quiet              Only validate the configuration, don't print anything.
  - [x]   --services               Print the service names, one per line.
  - [x]   --volumes                Print the volume names, one per line.
  - [x]   --hash="*"               Print the service config hash, one per line.
  - [x]  --profiles does not exist yet in docker-compose but has been requested
- [x]   create             Create services
  - [x]  #1277 --force-recreate       Recreate containers even if their configuration and image haven't changed. Incompatible with --no-recreate.
  - [x] #1277  --no-recreate          If containers already exist, don't recreate them. Incompatible with --force-recreate.
  - [x] #1291  --no-build             Don't build an image, even if it's missing.
  - [x] #1292  --build                Build images before creating containers.

- [x]   down               Stop and remove resources
  - [x] #1269 --rmi type              Remove images. 
  - [x] #1268  -v, --volumes           Remove named volumes declared in the `volumes` section of the Compose file and anonymous volumes attached to containers.
  - [x]  --remove-orphans        Remove containers for services not defined in the Compose file
  - [x] #1275 -t, --timeout TIMEOUT   Specify a shutdown timeout in seconds.
- [x]  #1396 events             Receive real time events from containers.
  - [x]  --json      Output events as a stream of json objects

- [x] #1266  exec               Execute a command in a running container
[...]

PR #1396, opened 2021-03-05
introduce docker compose events

ndeloof:
**What I did**
Introduce `compose events` command

**Related issue**
https://github.com/docker/compose-cli/issues/1283
we considered this one to be left until requested by user, but I actually need the backend code to implement attaching to containers on restart, so ¯\_(ツ)_/¯

### DIFF
--- a/api/compose/api.go
+++ b/api/compose/api.go
@@ -18,6 +18,7 @@ package compose
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"strings"
 	"time"
@@ -65,6 +66,8 @@ type Service interface {
 	UnPause(ctx context.Context, project *types.Project) error
 	// Top executes the equivalent to a `compose top`
 	Top(ctx context.Context, projectName string, services []string) ([]ContainerProcSummary, error)
+	// Events executes the equivalent to a `compose events`
+	Events(ctx context.Context, project string, options EventsOptions) error
 }
 
 // BuildOptions group options of the Build API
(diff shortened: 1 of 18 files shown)


api/compose/api.go
```diff
@@ -169,6 +171,12 @@ type RunOptions struct {
 	Index int
 }
 
+// EventsOptions group options of the Events API
+type EventsOptions struct {
+	Services []string
+	Consumer func(timestamp time.Time, service, container, status string, attributes map[string]string) error
```
gtardif, line 177: We should add a struct for all these params to avoid changing signature if we change this (long) list of params

ecs/sdk.go
```diff
@@ -565,7 +565,7 @@ func (s sdk) ListStackServices(ctx context.Context, stack string) ([]string, err
 			return nil, err
 		}
 		for _, r := range response.StackResourceSummaries {
-			if aws.StringValue(r.ResourceType) == "AWS::ECS::Service" {
+			if aws.StringValue(r.ResourceType) == "AWS::ECS::Services" {
```
gtardif, line 568: is this a bug fix (unrelated to events) ?

gtardif (approved):
LGTM, with minor improvement on consumer

ecs/sdk.go
ndeloof: oups

gtardif (commented):
Sorry I didn't check before, an e2e tests (maybe next to start/stop e2e) would be cool.

gtardif (changes requested):
an e2e tests (maybe next to start/stop e2e) would be cool.

gtardif (approved):

PR #1397, opened 2021-03-05
[WIP] compose local as a cli plugin

gtardif:
**What I did**
Keep ACI/ECS backend but ship compose command in docker cli (default context) as a cli plugin to make it easier in terms of deployment esp. on linux. 

Need some additional changes : packaging as a cli plugin in Desktop & CE-packaging (linux)
This PR includes deploying the compose CLI plugin as part of the linux install script.

**Related issue**

/test-aci
/test-ecs
/test-windows
/test-kube

### DIFF
--- a/cli/cmd/version.go
+++ b/cli/cmd/version.go
@@ -40,7 +40,11 @@ func VersionCommand() *cobra.Command {
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			err := runVersion(cmd)
 			if err != nil {
-				return cli.StatusError{StatusCode: 1, Status: err.Error()}
+				errMsg := ""
+				if err != nil {
+					errMsg = err.Error()
+				}
+				return cli.StatusError{StatusCode: 1, Status: errMsg}
 			}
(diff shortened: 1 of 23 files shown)


ndeloof (approved):
♥‿♥

gtardif:
We'll need to check CI build time, as it builds most of the code twice. Now with arm64, v6, v7, the cross build really takes time

cli/cmd/version.go
```diff
@@ -40,7 +40,11 @@ func VersionCommand() *cobra.Command {
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			err := runVersion(cmd)
 			if err != nil {
-				return cli.StatusError{StatusCode: 1, Status: err.Error()}
+				errMsg := ""
+				if err != nil {
+					errMsg = err.Error()
+				}
+				return cli.StatusError{StatusCode: 1, Status: errMsg}
```
ndeloof, line 47: Maybe I missed something, but afaict this is strictly equivalent

PR #1400, opened 2021-03-05
Add connection flags to root command

aiordache:
**What I did**
Added common flags to the root command: tls, cert flags etc.

Initialized the API client with the docker/cli `NewAPIClientFromFlags` which processes the config and tls options. This way we avoid duplicating the code that is already in the original cli. 

Set up a secured Docker engine following https://docs.docker.com/engine/security/protect-access/#use-tls-https-to-protect-the-docker-daemon-socket

```
$ docker -H=127.0.0.1:2376 compose ps
Error response from daemon: Client sent an HTTP request to an HTTPS server.

```
```
$ docker --tlsverify --tlscacert=/tmp/certs/ca.pem --tlscert=/tmp/certs/cert.pem --tlskey=/tmp/certs/key.pem -H=127.0.0.1:2376 compose ps
NAME                SERVICE             STATUS              PORTS
```

```
$ docker --tlsverify --tlscacert=/tmp/certs/ca.pem --tlscert=/tmp/certs/cert.pem --tlskey=/tmp/certs/key.pem -H=127.0.0.1:2376 compose up -d
[+] Running 1/1
 ⠿ Container compose-cli_web_1  Started                                                                                                                                      0.8s
```
```
$ docker --tlsverify --tlscacert=/tmp/certs/ca.pem --tlscert=/tmp/certs/cert.pem --tlskey=/tmp/certs/key.pem -H=127.0.0.1:2376 compose ps
NAME                SERVICE             STATUS              PORTS
compose-cli_web_1   web                 running             0.0.0.0:8080->80/tcp
```
```
$ docker --tlsverify --tlscacert=/tmp/certs/ca.pem --tlscert=/tmp/certs/cert.pem --tlskey=/tmp/certs/key.pem -H=127.0.0.1:2376 compose down
[+] Running 2/2
 ⠿ Container compose-cli_web_1    Removed                                                                                                                                    0.5s
 ⠿ Network "compose-cli_default"  Removed                                                                                                                                    0.5s


```

```
$ docker context use sshcontext
sshcontext

$ docker context ls
NAME                TYPE                DESCRIPTION                               DOCKER ENDPOINT               KUBERNETES ENDPOINT                 ORCHESTRATOR
default             moby                Current DOCKER_HOST based configuration   unix:///var/run/docker.sock   https://127.0.0.1:42745 (default)   swarm
sshcontext *        moby                My HOME server                            ssh://host                                                         
```

```
$ docker compose up
[+] Running 2/1
 ⠿ Network "compose-cli_default"  Created       0.3s
 ⠿ Container compose-cli_web_1    Created       0.0s
Attaching to compose-cli_web_1
...
```

```
$ docker -H ssh://host compose ps
NAME                SERVICE             STATUS              PORTS
compose-cli_web_1   web                 running             0.0.0.0:8080->80/tcp
```
```
$ docker compose ps
NAME                SERVICE             STATUS              PORTS
compose-cli_web_1   web                 running             0.0.0.0:8080->80/tcp

```
 

/test-kube
/test-aci
/test-ecs
/test-windows

### DIFF
--- a/api/context/context.go
+++ b/api/context/context.go
@@ -20,9 +20,14 @@ import (
 	gocontext "context"
 
 	"golang.org/x/net/context"
+
+	configfile "github.com/docker/cli/cli/config/configfile"
+	cliflags "github.com/docker/cli/cli/flags"
 )
 
 type currentContextKey struct{}
+type cliOptionsKey struct{}
+type configKey struct{}
 
 // WithCurrentContext sets the name of the current docker context
 func WithCurrentContext(ctx gocontext.Context, contextName string) context.Context {
@@ -32,5 +37,27 @@ func WithCurrentContext(ctx gocontext.Context, contextName string) context.Conte
 // CurrentContext returns the current context name
 func CurrentContext(ctx context.Context) string {
 	cc, _ := ctx.Value(currentContextKey{}).(string)
+	return cc
+}
+
+// WithCliOptions sets CLI options
+func WithCliOptions(ctx gocontext.Context, options cliflags.CommonOptions) context.Context {
+	return context.WithValue(ctx, cliOptionsKey{}, options)
+}
+
+// CliOptions returns the current context name
+func CliOptions(ctx context.Context) cliflags.CommonOptions {
+	cc, _ := ctx.Value(cliOptionsKey{}).(cliflags.CommonOptions)
+	return cc
+}
+
+// WithConfig sets docker config
+func WithConfig(ctx gocontext.Context, config configfile.ConfigFile) context.Context {
+	return context.WithValue(ctx, configKey{}, config)
+}
+
+// Config returns the docker config
+func Config(ctx context.Context) configfile.ConfigFile {
+	cc, _ := ctx.Value(cliOptionsKey{}).(configfile.ConfigFile)
 	return cc
 }
--- a/api/context/context.go
+++ b/api/context/context.go
@@ -21,13 +21,11 @@ import (
 
 	"golang.org/x/net/context"
 
-	configfile "github.com/docker/cli/cli/config/configfile"
 	cliflags "github.com/docker/cli/cli/flags"
 )
 
 type currentContextKey struct{}
 type cliOptionsKey struct{}
-type configKey struct{}
 
 // WithCurrentContext sets the name of the current docker context
 func WithCurrentContext(ctx gocontext.Context, contextName string) context.Context {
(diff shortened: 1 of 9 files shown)


api/context/context.go
```diff
@@ -34,3 +37,14 @@ func CurrentContext(ctx context.Context) string {
 	cc, _ := ctx.Value(currentContextKey{}).(string)
 	return cc
 }
+
+// WithCliOptions sets CLI options
+func WithCliOptions(ctx gocontext.Context, options cliflags.CommonOptions) context.Context {
+	return context.WithValue(ctx, cliOptionsKey{}, options)
+}
+
+// CliOptions returns the current context name
```
gtardif, line 46: update comment :)

cli/cmd/version.go
```diff
@@ -37,7 +37,9 @@ func VersionCommand() *cobra.Command {
 		Short: "Show the Docker version information",
 		Args:  cobra.MaximumNArgs(0),
 		Run: func(cmd *cobra.Command, _ []string) {
-			runVersion(cmd)
+			if err := runVersion(cmd); err != nil {
```
gtardif, line 40: why explicitely exit here and not let the error be managed in main.go ?? (this will prevent metrics working on this command)

cli/main.go
```diff
@@ -213,7 +212,26 @@ func main() {
 		compose.Command(ctype),
 		volume.Command(ctype),
 	)
+	if ctype == store.DefaultContextType || ctype == store.LocalContextType {
+		if len(opts.Hosts) > 0 {
```
gtardif, line 216: why change the current context ? I'd rather error with an explicit message if users are changing the host and also specifying -H, if we don't support overriding the local context. 
And definitely error if context is ACI or ECS

cli/cmd/version.go
```diff
@@ -48,34 +50,38 @@ func VersionCommand() *cobra.Command {
...
+		versionString, err = getOutFromMoby(cmd, fixedJSONArgs(os.Args[1:])...)
+		versionString = strings.Replace(versionString,
 			`"Version":`, fmt.Sprintf(`"CloudIntegration":%q,"Version":`, displayedVersion), 1)
 	default:
-		versionString = getOutFromMoby(cmd)
+		versionString, err = getOutFromMoby(cmd)
 	}
 
 	fmt.Print(versionString)
+	return err
```
gtardif, line 73: why do we need to return error (from moby exec) after displaying the version ?

local/backend.go
```diff
@@ -42,7 +46,13 @@ func init() {
 }
 
 func service(ctx context.Context) (backend.Service, error) {
-	apiClient, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	options := apicontext.CliOptions(ctx)
```
ndeloof, line 49: I really hate this way of passing hidden parameters through context and pretend we have a neutral API.
I'd prefer we adopt a more exeplicit `switch contextType` to initialize the adequate backend and parse command flags

cli/cmd/version.go
aiordache: Moby cli returns a non-zero exit code when it fails to retrieve the engine version. Compose CLI ignores it.

local/backend.go
gtardif: yes, we discussed this with Anca earlier, decided to get a first step like this to check everything is working fine, as changing the backend init API will impact a lot of places. But we should be doing this now with separate commit (change signature of backend init() func to pass what's required there). 
Using the context like this will also not work nicely with https://github.com/docker/compose-cli/pull/1397

ndeloof (approved):

PR #1401, opened 2021-03-06
Add arm64 support to linux install script

AaronDewes:
**What I did**

I added aarch64 support to the linux installl script, so it doesn't install an incompatible binary on arm64.

**Related issue**
There is no issue, should I open one?


gtardif (approved):

gtardif:
Thx for the PR, no specific need to open an issue first, this is well self-contained.

PR #1407, opened 2021-03-08
generate reference documentation

ndeloof:
**What I did**
adapted reference documentation generator from https://github.com/docker/app/tree/master/docs/yaml to produce compose-cli Ref

This is a WiP: need automation and improvement in commands definition (Long description, Examples, ...)

**Related issue**
https://github.com/docker/compose-cli/issues/1404


go.mod
```diff
@@ -60,6 +60,7 @@ require (
 	google.golang.org/grpc v1.33.2
 	google.golang.org/protobuf v1.25.0
 	gopkg.in/ini.v1 v1.62.0
+	gopkg.in/yaml.v2 v2.4.0
```
gtardif, line 63: I suppose this is needed, something we use is not available in v3 but only in v2 ?

gtardif (commented):
LGTM, a simple target in Makefile specifying the generate command would be useful. 
Improvements in command definition & examples can come as follow up PRs.

PR #1413, opened 2021-03-10
refactoring: create backend in main then pass as command target

ndeloof:
**What I did**
Changed architecture: backend is created from `main`(where init can depend on actual CLI flags) then pass to commands as target to execute

/test-ecs
/test-aci
/test-windows
/test-kube

### DIFF
--- a/api/backend/backend.go
+++ b/api/backend/backend.go
@@ -17,7 +17,6 @@
 package backend
 
 import (
-	"context"
 	"errors"
 	"fmt"
 
@@ -38,7 +37,7 @@ var (
 	errTypeRegistered = errors.New("backend: already registered")
 )
 
-type initFunc func(context.Context) (Service, error)
+type initFunc func() (Service, error)
 type getCloudServiceFunc func() (cloud.Service, error)
 
 type registeredBackend struct {
@@ -51,6 +50,18 @@ type registeredBackend struct {
 var backends = struct {
 	r []*registeredBackend
 }{}
+
+var instance Service
+
+// Current return the active backend instance
+func Current() Service {
+	return instance
+}
+
+// WithBackend set the active backend instance
+func WithBackend(s Service) {
+	instance = s
+}
 
 // Service aggregates the service interfaces
 type Service interface {
@@ -85,10 +96,10 @@ func Register(name string, backendType string, init initFunc, getCoudService get
 
 // Get returns the backend registered for a particular type, it returns
 // an error if there is no registered backends for the given type.
-func Get(ctx context.Context, backendType string) (Service, error) {
+func Get(backendType string) (Service, error) {
 	for _, b := range backends.r {
 		if b.backendType == backendType {
-			return b.init(ctx)
+			return b.init()
 		}
 	}
 
@@ -97,7 +108,7 @@ func Get(ctx context.Context, backendType string) (Service, error) {
 
 // GetCloudService returns the backend registered for a particular type, it returns
 // an error if there is no registered backends for the given type.
-func GetCloudService(ctx context.Context, backendType string) (cloud.Service, error) {
+func GetCloudService(backendType string) (cloud.Service, error) {
 	for _, b := range backends.r {
 		if b.backendType == backendType {
 			return b.getCloudService()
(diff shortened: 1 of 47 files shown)


api/backend/backend.go
```diff
@@ -52,6 +51,18 @@ var backends = struct {
...
 
+var instance Service
+
+// Current return the active backend instance
+func Current() Service {
+	return instance
+}
+
+// WithBackend set the active backend instance
+func WithBackend(s Service) {
```
gtardif, line 62: For these singleton get methods, maybe we should have an explicit error saying this is not initialised, the corresponding set method should be called first, rather than let the code continue with uninitialised values

api/client/client.go
```diff
@@ -25,38 +25,25 @@ import (
...
 
-	backendName := cc.Type()
-	if backendName == store.DefaultContextType && defaultBackend != "" {
-		backendName = defaultBackend
-	}
-
-	service, err := backend.Get(ctx, backendName)
-	if err != nil {
-		return nil, err
+	service := backend.Current()
```
gtardif, line 44: confirmed : would be more explicit (and more go-like) returning a potential error here, rather than letting the caller check if nil

cli/main.go
```diff
@@ -191,50 +194,69 @@ func main() {
...
-			TLSVerify: opts.TLSVerify,
+
+	if err = root.ExecuteContext(ctx); err != nil {
+		handleError(ctx, err, ctype, currentContext, cc, root)
+	}
+	metrics.Track(ctype, os.Args[1:], metrics.SuccessStatus)
+}
+
+func getBackend(ctype string, configDir string, opts cliopts.GlobalOpts) (backend.Service, error) {
+	switch ctype {
```
gtardif, line 234: why have this specific case for local / default backend, and not also use the init() mechanism, same as for ACI/ECS (using singletons to initialise cliOptions, same as for configDir, etc)
This would also move the api instanciation in the local backend code, that seems more logical

cli/main.go
```diff
@@ -191,50 +194,69 @@ func main() {
...
+		}
+		return local.NewService(apiClient), nil
 	}
-	ctx = apicontext.WithCurrentContext(ctx, currentContext)
-	ctx = store.WithContextStore(ctx, s)
-
-	if err = root.ExecuteContext(ctx); err != nil {
-		handleError(ctx, err, ctype, currentContext, cc, root)
+	service, err := backend.Get(ctype)
+	if errdefs.IsNotFoundError(err) {
```
gtardif, line 256: is there a specific reason to return nil without error here ?

gtardif: or do we have cases returning a non-nil service with IsNotFoundError ??

api/client/client.go
```diff
@@ -25,38 +25,25 @@ import (
...
-
-// NewWithDefaultLocalBackend returns a backend client associated with current context or local backend if on default context type
-func NewWithDefaultLocalBackend(ctx context.Context) (*Client, error) {
-	return newWithDefaultBackend(ctx, store.LocalContextType)
-}
-
-func newWithDefaultBackend(ctx context.Context, defaultBackend string) (*Client, error) {
-	currentContext := apicontext.CurrentContext(ctx)
-	s := store.ContextStore(ctx)
+	currentContext := apicontext.Current()
```
gtardif, line 36: This method centralises checks on singleton to ensure all is initialised ; need to also check if currentContext and store are properly initlised, and return clear errors if not, as with backend

gtardif (changes requested):
Mainly move the clientAPI instanciation in local backend code, and remove the switch case specific to local context type, otherwise LGTM

cli/main.go
ndeloof: because local backend require a docker.API which we build by CLI options (otherwise need to create another singleton for this purpose).

cli/main.go
ndeloof: This happens for `context create` command
Backend is actually optional for a few commands

cli/main.go
gtardif: OK, as discussed it will be easier to merge this and we can try follow-up improvements on top of this work

gtardif (approved):

PR #1415, opened 2021-03-10
Add restart command

ulyssessouza:
**Related issue**
Resolves https://github.com/docker/compose-cli/issues/1389

### DIFF
--- a/local/compose/convergence.go
+++ b/local/compose/convergence.go
@@ -390,3 +390,26 @@ func (s *composeService) startService(ctx context.Context, project *types.Projec
 	}
 	return eg.Wait()
 }
+
+func (s *composeService) restartService(ctx context.Context, serviceName string, timeout *time.Duration) error {
+	containerState, err := GetContextContainerState(ctx)
+	if err != nil {
+		return err
+	}
+	containers := containerState.GetContainers().filter(isService(serviceName))
+	w := progress.ContextWriter(ctx)
+	eg, ctx := errgroup.WithContext(ctx)
+	for _, c := range containers {
+		container := c
+		eg.Go(func() error {
+			eventName := getContainerProgressName(container)
+			w.Event(progress.RestartingEvent(eventName))
+			err := s.apiClient.ContainerRestart(ctx, container.ID, timeout)
+			if err == nil {
+				w.Event(progress.StartedEvent(eventName))
+			}
+			return err
+		})
+	}
+	return eg.Wait()
+}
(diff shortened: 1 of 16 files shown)


local/compose/convergence.go
```diff
@@ -390,3 +390,31 @@ func (s *composeService) startService(ctx context.Context, project *types.Projec
...
+			serviceFilter(service.Name),
+		),
+		All: true,
+	})
+	if err != nil {
+		return err
+	}
+	w := progress.ContextWriter(ctx)
+	eg, ctx := errgroup.WithContext(ctx)
+	for _, c := range containers {
```
ndeloof, line 407: I wonder we need to run this in respect to depedencies order(especially with wait-condition: healthy
If we don't, then passing the project name would be enough and avoid requirement to get the original compose file

ndeloof (changes requested):

local/compose/convergence.go
gtardif: Also, can we do a container restart on each container, or do we need to stop the whole project, and restart it ? 
not sure how this is done in `docker-compose`.

ndeloof:
don't forget to add a reference doc for this new command
=> https://github.com/docker/compose-cli/tree/main/docs/reference

local/compose/convergence.go
ulyssessouza: @ndeloof And it does https://github.com/docker/compose-cli/pull/1415/files#diff-d6629ccbd8641e5ff4ab3b8cccf9b8a4c7921f765f3bc0bc9ba8ee47bc715c8eR27-R35
Actually `docker-compose` seems to get the containers in an unsorted way, see https://github.com/docker/compose/blob/5b7851f55b4ae9533f42eed282ffad7c442778f5/compose/project.py#L471 but as you stated, I preferred to visit it in dependencies order.

@gtardif `docker-compose` restart is an API call and not a combination of `stop` followed by `start`. Also doing so would lead to possible data synch problems, since the compose file could be different in between operations. Container ID's and other things could be differ too.

ndeloof: Please don't invoke `ContainerList` for each service. Collect project containers once, then filter to get the service ones.
Command should not require a `project` but use the same logic you designed for `down` to rebuild model from project name (using labels)
We could later add labels for service depednencies so that we don't even need to retrieve and parse the compose file

gtardif (approved):

PR #1416, opened 2021-03-11
reference documentation

ndeloof:
**What I did**

Created reference documentation to address https://github.com/docker/docker.github.io/pull/12466#issuecomment-794225737

This is basically a copy/paste from reference documentation in docker-compose


docs/reference/compose.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Description
+
+You can use compose subcommand, `docker compose [-f <arg>...] [options] [COMMAND] [ARGS...]`, to build and manage
+multiple services in Docker containers.
+
+### Use `-f` to specify name and path of one or more Compose files
+Use the `-f` flag to specify the location of a Compose configuration file.
+
+#### Specifying multiple Compose files
```
gtardif, line 11: @Usha, I seem to remember there are limitations in how many ## we support in docs tooling, is this OK ?

docs/reference/compose.md
```diff
@@ -0,0 +1,86 @@
...
+from the command line or by setting up a `COMPOSE_FILE` environment variable in your shell or in an environment file.
+
+For an example of using the `-f` option at the command line, suppose you are running the Compose Rails sample, and 
+have a `compose.yaml` file in a directory called `sandbox/rails`. You can use a command like `docker compose pull` to 
+get the postgres image for the db service from anywhere by using the `-f` flag as follows: 
+```
+docker compose -f ~/sandbox/rails/compose.yaml pull db
+```
+
+### Use `-p` to specify a project name
```
gtardif, line 63: In the -p section, maybe add examples of `compose -p xxx down` without a compose file (or linked to the result of  `compose ls`)

gtardif: (could be done as a follow up PR if we want ot mainly cc/paste from docker-compose help here)

ndeloof: I indeed would like we get a reasonable reference doc based on docker-compose one, then iterate on improvements / docker-cli specific topics
About -p, imho `down` is not a relevant example as this one actually need the original compose file (even we retrieve it from labels). `logs` would be a better one

PR #1420, opened 2021-03-12
Add "docs" target to generate reference docs

ulyssessouza:
Also add this as dependency to `pre-commit` target

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ validate-go-mod: ## Validate go.mod and go.sum are up-to-date
 
 validate: validate-go-mod validate-headers ## Validate sources
 
-pre-commit: validate import-restrictions check-dependencies lint cli test e2e-local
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -132,10 +132,13 @@ clean-aci-e2e: ## Make sure no ACI tests are currently runnnig in the CI when in
 	@ az group list | jq '.[].name' | grep -i E2E-Test
 	az group list | jq '.[].name' | grep -i E2E-Test | xargs -n1 az group delete -y --no-wait -g
 
+docs:
```
ndeloof, line 135: already there: https://github.com/docker/compose-cli/blob/main/builder.Makefile#L100

ndeloof: could maybe be containerized from this Makefile?

ndeloof: /rebase


just testing the rebase trigger

Issue #1427, opened 2021-03-14
Subsequent calls to multiple services compose files randomly makes Cloudformation to update services


flaviostutz:
When doing subsequent calls for deploying a non changing docker-compose.yml file, sometimes Cloudformation redeploys some services because it identifies a change in cloudformation file, which makes the deploy much longer.

1. Create a docker-compose.yml file with two services
2. Deploy the compose file to ECS using compose
3. Deploy the same compose file again using compose
4. Check if a changeset was identified. It not, redeploy again until you see the changes. Files with more than 4 services almost everytime causes changes.

**Describe the results you expected:**
I expected those untouched services not to be redeployed, because it takes a lot of time.

**Insights:**
Investigating further, I found that the generated CloudFormation file randomly changes the order of the services, causing CF to create an active changeset. Maybe forcing a sorting in services before template generation would fix this.

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Cloud integration: 1.0.7
 Version:           20.10.2
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        2291f61
 Built:             Mon Dec 28 16:12:42 2020
 OS/Arch:           darwin/amd64
 Context:           ecs3
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.2
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       8891c58
  Built:            Mon Dec 28 16:15:28 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  scan: Docker Scan (Docker Inc., v0.5.0)

Server:
 Containers: 3
  Running: 0
  Paused: 0
  Stopped: 3
 Images: 2082
 Server Version: 20.10.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
[...]

PR #1428, opened 2021-03-14
Sorting services in order to make cloudformation template file generation idempotent

flaviostutz:
Sorted services in order to make cloudformation template generation the same for the same set of services

**Related issue**
#1427 

image


ndeloof:
Closing according to https://github.com/docker/compose-cli/issues/1427#issuecomment-799542844

Issue #1405, opened 2021-03-08
inconsistent container name reported on console


ndeloof:
to align compose-cli with docker-compose we reproduced a debatable behaviour on console output:
- container logs are prefixed by container name _without project name prefix_
- container events are reported with full container name:

```console
 ⠿ Container project_simple_1   Created                                    0.0s
Attaching to project_simple_1
simple_1  | hello
project_simple_1 exited with code 1
```

I propose we update this behaviour to offer a more homogeneous UX by only use the "container name without project prefix".
Please comment if you heavily depend on this legacy behaviour !

ndeloof:
closing as #1433 has been merged

PR #1433, opened 2021-03-16
use "container name witout project" for both logs and events

ndeloof:
**What I did**

adopted "container name without project" as the canonical name for both logs prefixes and lifecycle event

**Related issue**
https://github.com/docker/compose-cli/issues/1405


aci/convert/volume.go
```diff
@@ -139,7 +139,7 @@ type volumeInput struct {
 
 // parse takes a candidate string and creates a volumeInput
 // Candidates take the form of <source>[:<target>][:<permissions>]
-// Source is of the form `<storage account>/<fileshare>`
+// Container is of the form `<storage account>/<fileshare>`
```
gtardif, line 142: a find/replace with too large scope ? (and I suppose many more instances following)

api/compose/api.go
```diff
@@ -314,22 +314,21 @@ type Stack struct {
...
+	Log(service, container, message string)
+	Status(container, msg string)
+	Register(container string)
 }
 
 // ContainerEventListener is a callback to process ContainerEvent from services
 type ContainerEventListener func(event ContainerEvent)
 
-// ContainerEvent notify an event has been collected on Source container implementing Service
+// ContainerEvent notify an event has been collected on Container container implementing Service
```
gtardif, line 325: same here

gtardif (approved):
Just fix unwanted replace of source/ Container. Otherwise LGTM

aci/convert/volume.go
ndeloof: ouch :'(

Issue #1441, opened 2021-03-19
Add compose pull parallel flags for compatibility


gtardif:
support pull --parallel / --no-parallel flags for backward compatibility (deprecated, without behaviour change)

PR #1443, opened 2021-03-19
introduce --parallel  and --no-parallel (hidden, deprecated) flags

ndeloof:
**What I did**
introduce --parallel  and --no-parallel (hidden, deprecated) flags

this prevent backward compatibility issues for users migrating from docker-compose
marked deprecated, implemented as no-op

close https://github.com/docker/compose-cli/issues/1441

### DIFF
--- a/cli/cmd/compose/pull.go
+++ b/cli/cmd/compose/pull.go
@@ -18,7 +18,10 @@ package compose
 
 import (
 	"context"
+	"fmt"
+	"os"
 
+	"github.com/morikuni/aec"
 	"github.com/spf13/cobra"
 
 	"github.com/docker/compose-cli/api/client"
@@ -29,6 +32,8 @@ type pullOptions struct {
 	*projectOptions
 	composeOptions
 	quiet       bool
+	parallel    bool
+	noParallel  bool
 	includeDeps bool
 }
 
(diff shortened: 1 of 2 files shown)


cli/cmd/compose/pull.go
```diff
@@ -40,11 +43,19 @@ func pullCommand(p *projectOptions) *cobra.Command {
 		Use:   "pull [SERVICE...]",
 		Short: "Pull service images",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if opts.noParallel {
+				logrus.Warn("`--no-parallel` flag is deprecated and will be ignored")
```
chris-crone, line 47: Do we want a log here or just to write something to stderr?

ndeloof: I would assume logrus "warn" is designed for this purpose, but no strong opinion here
what does docker/cli for such cases ?

chris-crone: Looks like it uses logrus: https://github.com/docker/cli/blob/88c6089300a82d3373892adf6845a4fed1a4ba8d/cli/command/container/opts.go#L345

I was thinking that a log with timestamp, etc. wasn't really ideal but consistency is more important.

chris-crone (approved):
LGTM

cli/cmd/compose/pull.go
ndeloof: we already have a comparable DEPRECATION message on compose command, will align to it

gtardif (approved):

PR #1444, opened 2021-03-19
introduce config --services, --volumes, --hash for backward compatibility

ndeloof:
**What I did**
Implemented --services, --volumes, --hash pseudo-commands on `compose config`

this will prevent compatibility issues with docker-compose

**Related issue**
close https://github.com/docker/compose-cli/issues/1442

### DIFF
--- a/utils/hash.go
+++ b/utils/hash.go
@@ -14,27 +14,21 @@
    limitations under the License.
 */
 
-package compose
+package utils
 
 import (
 	"encoding/json"
 
+	"github.com/compose-spec/compose-go/types"
 	"github.com/opencontainers/go-digest"
 )
 
-func jsonHash(o interface{}) (string, error) {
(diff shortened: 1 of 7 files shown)


utils/hash.go
```diff
@@ -14,27 +14,21 @@
...
-}
-
-func contains(slice []string, item string) bool {
-	for _, v := range slice {
-		if v == item {
-			return true
-		}
-	}
-	return false
+	return digest.SHA256.FromBytes(bytes).Hex(), nil
```
gtardif, line 33: why this change to Hex ? 
And then I just noticed `// Hex is deprecated. Please use Digest.Encoded.`

ndeloof: updated to use Encoded()
String() do include the algorithm (`sha256:...`), so aligned with docker-compose

PR #1445, opened 2021-03-19
Deprecated flag --no-ansi

gtardif:

**What I did**
Add deprecated `--no-ansi` (hidden) for backward compatibility

**Related issue**

### DIFF
--- a/cli/cmd/compose/compose.go
+++ b/cli/cmd/compose/compose.go
@@ -103,18 +103,26 @@ func (o *projectOptions) toProjectOptions() (*cli.ProjectOptions, error) {
 func Command(contextType string) *cobra.Command {
 	opts := projectOptions{}
 	var ansi string
+	var noAnsi bool
 	command := &cobra.Command{
 		Short:            "Docker Compose",
 		Use:              "compose",
 		TraverseChildren: true,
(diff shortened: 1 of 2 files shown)


cli/cmd/compose/compose.go
```diff
@@ -103,18 +103,26 @@ func (o *projectOptions) toProjectOptions() (*cli.ProjectOptions, error) {
...
 	var ansi string
+	var noAnsi bool
 	command := &cobra.Command{
 		Short:            "Docker Compose",
 		Use:              "compose",
 		TraverseChildren: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if noAnsi {
+				if ansi != "auto" {
+					return errors.New(aec.Apply(`cannot specify DEPRECATED "--no-ansi" and "--ansi". Please use only "--ansi".`, aec.RedF))
```
ndeloof, line 114: is it relevant to apply colors to golang error message?
I'm fine we do when we directly display message on stderr, but errors we forward to caller should be plain text

gtardif: no indeed, an error in CC/paste

ndeloof (approved):

PR #1448, opened 2021-03-22
--no-interpolate

ndeloof:
**What I did**
introduce --no-interpolate flag to align with docker-compose
require https://github.com/compose-spec/compose-go/pull/117

**Related issue**


gtardif (approved):
LGTM, just merge the compose-go and update the go.mod

Issue #1409, opened 2021-03-10
Support compose pull --ignore-pull-failures


gtardif:
`compose pull --ignore-pull-failures (Pull what it can and ignores images with pull failures.)`

Follow up from https://github.com/docker/compose-cli/pull/1399

gtardif:
Actually already supported

PR #1454, opened 2021-03-22
Add --ignore-pull-failures to command pull

ulyssessouza:
**What I did**
Add --ignore-pull-failures to command pull

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/1409

### DIFF
--- a/local/compose/pull.go
+++ b/local/compose/pull.go
@@ -21,22 +21,25 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"strings"
 
 	"github.com/compose-spec/compose-go/types"
 	"github.com/distribution/distribution/v3/reference"
+	"github.com/docker/buildx/driver"
 	cliconfig "github.com/docker/cli/cli/config"
 	moby "github.com/docker/docker/api/types"
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/docker/docker/registry"
 	"golang.org/x/sync/errgroup"
 
+	"github.com/docker/compose-cli/api/compose"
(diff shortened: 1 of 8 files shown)


local/compose/pull.go
```diff
@@ -64,73 +67,89 @@ func (s *composeService) Pull(ctx context.Context, project *types.Project) error
...
-				Text:   "Pulling",
-			})
-			ref, err := reference.ParseNormalizedNamed(service.Image)
+			err := s.pullServiceImage(ctx, service, info, configFile, w)
 			if err != nil {
-				return err
+				if !opts.IgnoreFailures {
+					return err
+				}
+				w.Event(progress.Event{
```
ndeloof, line 75: progress event should be also created when err == nil

ndeloof, line 75: event should be created before returning error, so user see it reported on progress before the command fails

ndeloof (approved):
LGTM with a minor UX change

local/compose/pull.go
ulyssessouza: This is done inside `err := s.pullServiceImage(ctx, service, info, configFile, w)` just some lines before. It calls https://github.com/docker/compose-cli/pull/1454/files#diff-49c56ffdcd4d1b6c21e27379db32f411e24665476483f4901e99359d2eb83177R155

gtardif: So if all event display is done inside `pullServiceImage` (that makes sense), we should not have this event displayed here, just check the ignoreFailures option and return the err or nil ?

gtardif: OK, I understand now, this is a workaround to still display the error message when discarding the error with --ignore-failure. And symetrical to what we do with push

Issue #1304, opened 2021-02-15
`down` invoke `ContainerList` twice


ndeloof:
**Description**

`down` command invokes `ContainerList` with same parameters **twice**
(as unit test demonstrate https://github.com/docker/compose-cli/blob/main/local/compose/down_test.go#L41)

implementation could load containers (use `getContainers` with adequate filters), then pass containers to load project from labels

ndeloof:
I also noticed we invoke `ImageInspectWithRaw` twice:
- once to check service image is present (localImagePresent)
- then again looking for image's "anonymous" volume (buildContainerVolumes)

ulyssessouza:
@ndeloof I've just created an [issue](https://github.com/docker/compose-cli/issues/1466) to address your comment.

PR #1455, opened 2021-03-23
Reuse container list on command down

ulyssessouza:
**What I did**
Use cached container list on command down

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/1304

### DIFF
--- a/local/compose/down.go
+++ b/local/compose/down.go
@@ -38,22 +38,23 @@ func (s *composeService) Down(ctx context.Context, projectName string, options c
 	w := progress.ContextWriter(ctx)
 	resourceToRemove := false
 
+	var containers Containers
+	containers, err := s.apiClient.ContainerList(ctx, moby.ContainerListOptions{
+		Filters: filters.NewArgs(projectFilter(projectName)),
+		All:     true,
+	})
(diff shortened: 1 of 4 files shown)


local/compose/down.go
```diff
@@ -35,6 +35,10 @@ import (
 )
 
 func (s *composeService) Down(ctx context.Context, projectName string, options compose.DownOptions) error {
+	ctx, err := s.getUpdatedContainersStateContext(ctx, projectName)
```
gtardif, line 38: Ah please do not add more stored magical things in context, for #1304 we wanted to add an *explicit* parameter `containers` to projectFromContainerLabels(), and load containers once before invoking [projectFromContainerLabels](https://github.com/docker/compose-cli/blob/d3dbc90fdeb2f4fdcd48e1a1c711d5bb58c2c960/local/compose/down.go#L42). 
For the inspect part I have not checked the details, but I'd also rather not cache magical things in the context and pass explicit parameters.

gtardif (changes requested):
Too much magical things store in context, don't help following code as things are not explicit in method signatures

gtardif (approved):

Issue #1453, opened 2021-03-22
exec and run shell's have strange character output bugs


BretFisher:
**Description**

Quirks with exec shell and character output.

**Steps to reproduce the issue:**
1. Create a simple compose file:

```
services:
  ruby:
    image: ruby
```

2. Run an exec while overwriting entrypoint `docker compose run --entrypoint bash ruby`
    1. Try to ctrl-C (I see the ^C printed on screen, but no typical shell ctrl-C experience)
3. Run a `run` and get proper irb output `docker compose run ruby`
    1. notice the various `^[[30;` output.

Both of those commands show proper output in `docker-compose.

I tried this both on iTerm and Terminal on mac with both bash and zsh, and saw the same issues each time.

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Cloud integration: 1.0.9
 Version:           20.10.5
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        55c4c88
 Built:             Tue Mar  2 20:13:00 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.5
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       363e9a8
  Built:            Tue Mar  2 20:15:47 2021
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker context show`:**  

```
default
```

ndeloof:
"weird" console output is caused by compose cli not setting the terminal in character mode when attaching to container, we need to fix this.

second issue with Ctrl+C signal is unclear to me, maybe related.

RKushnir:
Hi. Since the latest update of Docker Desktop to 3.5.2 (66501) I started seeing an issue that completely fits this description. 

I'm on a mac using iterm. After executing each command the shell prompt contains characters like `^[[25;12R` and Ctrl-C prints `^C` then ignores input until I press Ctrl-C again, at which point it says `ERRO[<4 digits>] got 3 SIGTERM/SIGINTs, forcing shutdown` and exits.

[...]

PR #1462, opened 2021-03-24
set terminal in character mode when attached to a container

ndeloof:
**What I did**
adopted docker/cli code to set terminal is "raw" character mode when attached to container's stdin

**Related issue**
https://github.com/docker/compose-cli/issues/1453


gtardif (approved):
Also fixes the command you type being re-displayed...

Issue #1457, opened 2021-03-23
docker compose up --build forces build but does not restart container 


gtardif:
**Description**

`docker compose up --build -d` forces rebuild of images if composefile includes build section for images. 
However, if the containers are already started, this command does not restart the container. It will only restart the container if I add the `--force-recreate` option, where `docker-compose` will restart containers after rebuilding images.

**Steps to reproduce the issue:**
1. `docker compose up -d` a project
2. Update content
3. `docker compose up --build -d` to update and deploy changes
4. See that updates are not deployed
5. `docker compose up --build -d --force-recreate` to see updates deployed

**Describe the results you expected:**
No need to add the force-recreate option to restart container

PR #1464, opened 2021-03-25
Refactor `up --build` behavior

ulyssessouza:
**What I did**
This makes `--build` imply `--force-recreate` in command `up`

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/1457

### DIFF
--- a/local/compose/build.go
+++ b/local/compose/build.go
@@ -24,7 +24,10 @@ import (
 	"path"
 	"strings"
 
+	moby "github.com/docker/docker/api/types"
+
 	"github.com/docker/compose-cli/api/compose"
+	composeprogress "github.com/docker/compose-cli/api/progress"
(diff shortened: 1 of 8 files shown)


local/compose/build.go
```diff
@@ -187,6 +187,18 @@ func (s *composeService) build(ctx context.Context, project *types.Project, opts
 	if err == nil {
 		err = errW
 	}
+
+	for imageName := range opts {
+		for _, c := range observedState {
+			if c.Image == imageName {
+				err := s.restartContainer(ctx, c)
```
gtardif, line 194: Are we sure we don't need to recreate the container, to use the updated image ? (this is invoking `s.apiClient.ContainerRestart()` here)

gtardif (approved):
Just needs a rebase

PR #1465, opened 2021-03-25
redirect `docker-compose *` to `docker compose ...`

ndeloof:
**What I did**
Created a thin go binary to parse `docker-compose` flags and re-order those so that we can actually run `docker compose ...`

**Related issue**
https://github.com/docker/dev-tooling-team/issues/281


v1/main.go
```diff
@@ -17,14 +17,17 @@
...
-	"syscall"
-
-	"github.com/spf13/cobra"
-
+	"github.com/docker/compose-cli/api/config"
+	"github.com/docker/compose-cli/cli/metrics"
 	"github.com/docker/compose-cli/cli/mobycli/resolvepath"
 	"github.com/docker/compose-cli/utils"
+	"github.com/spf13/cobra"
+	"io/ioutil"
```
chris-crone, line 27: Import ordering

PR #1467, opened 2021-03-25
Add plain tail messages on progress writer

ulyssessouza:
**What I did**
This allows progress writer to print plain text after printing all the `tty` messages.

### DIFF
--- a/local/compose/push.go
+++ b/local/compose/push.go
@@ -70,12 +70,7 @@ func (s *composeService) Push(ctx context.Context, project *types.Project, optio
 				if !options.IgnoreFailures {
 					return err
 				}
-				w.Event(progress.Event{
-					ID:         fmt.Sprintf("Pushing %s:", service.Name),
-					Text:       fmt.Sprintf("%v", err),
-					Status:     progress.Error,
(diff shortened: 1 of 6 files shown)


local/compose/push.go
```diff
@@ -70,12 +70,7 @@ func (s *composeService) Push(ctx context.Context, project *types.Project, optio
 				if !options.IgnoreFailures {
 					return err
 				}
-				w.Event(progress.Event{
-					ID:         fmt.Sprintf("Pushing %s:", service.Name),
-					Text:       fmt.Sprintf("%v", err),
-					Status:     progress.Error,
-					StatusText: fmt.Sprintf("%s", err),
-				})
+				w.TailMsgf("Pushing %s: %s", service.Name, err.Error())
```
ndeloof, line 73: Doing so, we will show the error as tail message 👍  but the progress won't complete. Shouldn't we do both?

ulyssessouza: From what I see this Error event is unique. It has no `progress.Working` preceding it.

ndeloof (approved):

Issue #1491, opened 2021-04-04
compose up deadlock with no enabled services


mxk:
**Description**

Running `docker compose up` with the following docker-compose.yaml results in a deadlock:

```yaml
services:
  init:
    profiles: [init]
    image: library/ubuntu:20.04
    command: echo 'hello, world'
```

**Steps to reproduce the issue:**

1. Create docker-compose.yaml with the above contents.
2. Run `docker compose up` (without enabling the init profile).
3. Observe the deadlock.
4. Press Ctrl-C and note that the process does not terminate.

**Describe the results you received:**

Comparing `docker-compose` vs `docker compose`:

```
$ docker-compose up
Attaching to
$ docker compose up
Attaching to
# Press Ctrl-C
Gracefully stopping...
# Deadlock
```

**Describe the results you expected:**

CLI should terminate after printing "Attaching to". If it doesn't, Ctrl-C should stop it.

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Cloud integration: 1.0.9
 Version:           20.10.5
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        55c4c88
 Built:             Tue Mar  2 20:14:53 2021
 OS/Arch:           windows/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.5
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       363e9a8
  Built:            Tue Mar  2 20:15:47 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker context show`:**  

```
default
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  scan: Docker Scan (Docker Inc., v0.5.0)

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 17
 Server Version: 20.10.5
[...]

PR #1495, opened 2021-04-06
deadlock waiting for services when none selected

ndeloof:
**What I did**
Report an error when there's no service attached to project after profiles and options have been applied. This prevents a deadlock waiting for containers to attach

**Related issue**
https://github.com/docker/compose-cli/issues/1491

### DIFF
--- a/cli/cmd/compose/up.go
+++ b/cli/cmd/compose/up.go
@@ -235,6 +235,10 @@ func runCreateStart(ctx context.Context, opts upOptions, services []string) erro
 		return err
 	}
 
+	if len(project.Services) == 0 {
+		return fmt.Errorf("no service selected")
+	}
+
 	_, err = progress.Run(ctx, func(ctx context.Context) (string, error) {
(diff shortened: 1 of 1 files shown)


cli/cmd/compose/up.go
```diff
@@ -396,7 +400,7 @@ type printer struct {
 	queue chan compose.ContainerEvent
 }
 
-func (p printer) run(ctx context.Context, cascadeStop bool, exitCodeFrom string, consumer compose.LogConsumer, stopFn func() error) (int, error) { //nolint:unparam
+func (p printer) run(cascadeStop bool, exitCodeFrom string, consumer compose.LogConsumer, stopFn func() error) (int, error) { //nolint:unparam
```
chris-crone, line 403: Can we remove the unparam now that we have removed the unused `context.Context`?

gtardif, line 403: Do we still need //nolint:unparam here ?

gtardif (approved):

PR #1497, opened 2021-04-06
don't use filepath to process remote bind paths

ndeloof:
**What I did**
Removed use of `filepath` methods to compute bind mount target, as target OS != source and source's platform-specific format would then be applied.

**Related issue**
https://github.com/docker/compose-cli/issues/1493

### DIFF
--- a/local/compose/create.go
+++ b/local/compose/create.go
@@ -618,9 +618,9 @@ func buildContainerConfigMounts(p types.Project, s types.ServiceConfig) ([]mount
 	for _, config := range s.Configs {
 		target := config.Target
 		if config.Target == "" {
-			target = filepath.Join(configsBaseDir, config.Source)
-		} else if !filepath.IsAbs(config.Target) {
-			target = filepath.Join(configsBaseDir, config.Target)
+			target = configsBaseDir + config.Source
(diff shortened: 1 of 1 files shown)


ndeloof:
need to apply the same fix for other mounts, but would welcome review and feedback before

gtardif (approved):

local/compose/create.go
```diff
@@ -649,13 +649,13 @@ func buildContainerConfigMounts(p types.Project, s types.ServiceConfig) ([]mount
...
+	secretsDir := "/run/secrets/"
 	for _, secret := range s.Secrets {
 		target := secret.Target
 		if secret.Target == "" {
-			target = filepath.Join(secretsDir, secret.Source)
-		} else if !filepath.IsAbs(secret.Target) {
-			target = filepath.Join(secretsDir, secret.Target)
+			target = secretsDir + secret.Source
+		} else if !strings.HasPrefix(secret.Target, "/") /* Is Unix absolute patZZ */ {
+			target = secretsDir + secret.Target
```
gtardif, line 658: Maybe join the 2 if, making it more clear both result in the same target

Issue #1498, opened 2021-04-06
Implement docker compose images for compatibility with docker-compose


gtardif:
only option: 
` -q, --quiet Only display IDs`

PR #1505, opened 2021-04-07
Add `compose images` cmd

aiordache:
Added the `compose images` command.

```
[example]$ docker-compose images
    Container           Repository       Tag       Image Id       Size  
------------------------------------------------------------------------
example_backend_1    nginx              latest   35c43ace9216   133.1 MB
example_backend_2    nginx              latest   35c43ace9216   133.1 MB
example_backend_3    nginx              latest   35c43ace9216   133.1 MB
example_frontend_1   example_frontend   latest   ca1d0f22403f   22.59 MB
```
```
[example]$ docker compose images
Container            Repository          Tag                 Image Id            Size
example_backend_1    nginx               latest              35c43ace9216        133MB
example_backend_2    nginx               latest              35c43ace9216        133MB
example_backend_3    nginx               latest              35c43ace9216        133MB
example_frontend_1   example_frontend    latest              ca1d0f22403f        22.6MB
```
```
[example]$ docker-compose images frontend
    Container           Repository       Tag       Image Id       Size  
------------------------------------------------------------------------
example_frontend_1   example_frontend   latest   ca1d0f22403f   22.59 MB
```
```
[tower example]$ docker compose images frontend
Container            Repository          Tag                 Image Id            Size
example_frontend_1   example_frontend    latest              ca1d0f22403f        22.6MB
```
```
[example]$ docker-compose images -q
35c43ace9216212c0f0e546a65eec93fa9fc8e96b25880ee222b7ed2ca1d2151
ca1d0f22403f0c178aaba1cfaf5c2ccbed00fc768605fe98a01917a076ab5b89
```
```
[example]$ docker compose images -q
35c43ace9216212c0f0e546a65eec93fa9fc8e96b25880ee222b7ed2ca1d2151
ca1d0f22403f0c178aaba1cfaf5c2ccbed00fc768605fe98a01917a076ab5b89

```
Closes https://github.com/docker/compose-cli/issues/1498

### DIFF
--- a/cli/cmd/compose/images.go
+++ b/cli/cmd/compose/images.go
@@ -1,0 +1,107 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compose
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/docker/compose-cli/api/client"
+	"github.com/docker/compose-cli/api/compose"
+	"github.com/docker/compose-cli/cli/formatter"
+	"github.com/docker/compose-cli/utils"
+	"github.com/docker/docker/pkg/stringid"
+
+	units "github.com/docker/go-units"
+)
+
+type imageOptions struct {
+	*projectOptions
+	Quiet bool
+}
+
+func imagesCommand(p *projectOptions) *cobra.Command {
+	opts := imageOptions{
+		projectOptions: p,
+	}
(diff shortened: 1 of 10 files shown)


cli/cmd/compose/images.go
```diff
@@ -0,0 +1,103 @@
...
+	return imgCmd
+}
+
+func runImages(ctx context.Context, opts imageOptions, services []string) error {
+	c, err := client.New(ctx)
+	if err != nil {
+		return err
+	}
+
+	project, err := opts.toProject(services)
```
ndeloof, line 63: better use `toProjectName`, as backend don't need an actual project, just name

ndeloof (approved):
LGTM
no need to pass a full types.project, can just use project name to avoid relying on original compose.yaml file

cli/cmd/compose/images.go
aiordache: In the backend we need the sevice names from the project to filter the container list. I could pass the project name and the service list in the api method if you think it would be better.

aiordache: In the backend we need the sevice names from the project to filter the container list. I could pass the project name and the service list in the api method if you think it would be better.

ndeloof: indeed, see https://github.com/docker/compose-cli/pull/1509/files#diff-bf7596d83ffc37f2610a3a7b826911de05f2c7996bdf3f6ad4add0a88de35195R29-R30

gtardif (commented):
A small E2E test would be nice for new commands (maybe add to an existing e2e, once a stack if running). Otherwise LGTM, after changing API to pass `project string` as suggested by Nicolas

gtardif (approved):

Issue #1500, opened 2021-04-07
Add Support to use existing IAM role by referring to it by its ARN


rosven:
It is currently possible to create policies before hand and refer to those as `x-aws-policies` in the configuration. This dynamically creates a IAM Role with a role name consisting of `myApplicationTaskRole-someHash`. Likewise, it's supported to create a tailored IAM role with `x-aws-role`. 

A problem arises when you want to refer to the IAM role elsewhere. For example in a terraform flow where the ARN to the IAM Role needs to be defined. In my specific use case, I need to define my IAM role inside of the access policies in an Elasticsearch service. This can be done by looking up the created role ARN and manually added to access policies. However it's not possible in an automated flow because the name is not deterministic.

Would it be possible to add support to refer to an existing IAM Role by ARN inside of the config and that it is automated so that role is used for the created task instead?

Grateful for non-hacky work arounds or added support for this!

rosven:
Seems that if this part in the file: https://github.com/docker/compose-cli/blob/6a0269f7b93b8e9228a1ed009e6d0a1144d83651/ecs/cloudformation.go#L453
Lines: https://github.com/docker/compose-cli/blob/6a0269f7b93b8e9228a1ed009e6d0a1144d83651/ecs/cloudformation.go#L174-L183

```
taskRole := b.createTaskRole(project, service, template, resources)

definition, err := b.createTaskDefinition(project, service, resources)
if err != nil {
     return err
}
definition.ExecutionRoleArn = cloudformation.Ref(taskExecutionRole)
if taskRole != "" {
    definition.TaskRoleArn = cloudformation.Ref(taskRole)
}
```

Checking for taskRole ARN in config before creating it and setting it to config ARN `definition.TaskRoleArn = configArn`. Could work?

defyrlt:
@rosven: i haven't tried this, and not sure if it fits the description of a "non-hacky" solution, but you can probably define an output via `x-aws-cloudformation` and use it in terraform.

rosven:
@defyrlt Thank you for the suggestion. Unfortunately It sounds like a similar solution to what I already found (which I consider hacky), where you simply search for the role name in the aws cli and get ARNS matching that. I would like to know deterministically what the ARN is to the role without looking it up so that I can have the role ARN locked and checked in, inside of my other configurations. I also dont want to terraform every time I make changes and deploy services to ECS, thats kind of the point of using this feature.

This is the other way of doing it: (I have not tested the one you suggest)
```
[...]

Issue #1411, opened 2021-03-10
[Proposal] define AWS infrastructure in a separate config file


ndeloof:
**Description**
to support AWS deployment we define custom extensions `x-aws-vpc`, `x-aws-loadbalancer`, `x-aws-roles` ... This comes with a few drawbacks:

1. `x-aws-*` attributes are not validated, and users might missuse them without being warned (see https://github.com/compose-spec/compose-go/issues/66) 
1. AWS infrastructure is way more complex than the compose one, especially considering networking components. Many user request a way to define subnets assignment to services, which would require extra extensions. Those extensions actually define a full model on top of the compose one, which is entangled in the yaml file
1. Last but not least, custom extensions intent by compose-spec is to allow experimentation with new attributes to be proposed to the compose-spec for normalization. But obviously those aws-specific ones won't become standard attributes.

I suggest we adopt the same approach proposed by [aws copilot](https://aws.amazon.com/fr/containers/copilot/) to define the AWS infrastructure within a dedicated file (with schema to validate content). By default, ECS intregration would behave like it does today in absence of `x-aws-*` extensions (i.e, use AWS account's default VPC) to make newcomer experience easy. Passing an AWS infrastructure file with explicit resources ARN will allow to tweak resource assignment to services. This approach can help keep the comopse file a developer-centric one while the AWS-specific one would be more operator-centric

ndeloof:
cc @CarmenAPuccio

flaviostutz:
Great. So in this proposal if I want to use an existing LB listener, for example, you would pass its ARN by using this dedicated file, right? Do you have more details on the format of this file? I looked at CoPilot but could not get it...

blampe:
Could these attributes be inferred from `ecs-params.yaml`? That would give you a [well-defined](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/cmd-ecs-cli-compose-ecsparams.html) schema as a starting point.

I stumbled on this issue by way of #921, and while I would prefer to use `docker compose` it seems I'll need to use `ecs-cli` for the time being. If `docker compose` understood how to consume `ecs-cli` params, that would give me a smooth onramp to come back to docker. I could understand not wanting to couple yourself to that schema, though!

mikelhamer:
Yes please!

EricBorland:
What's the status of this? ETA for custom subnets?

komatom:
[...]

PR #1506, opened 2021-04-07
ADDS Support to define IAM roles in yml by ARN

rosven:
**What I did**
Added extension to support IAM policys in config to be referred to by ARN.

**Related issue**
resolves: [1500](https://github.com/docker/compose-cli/issues/1500)

### DIFF
--- a/docs/ecs-compose-examples.md
+++ b/docs/ecs-compose-examples.md
@@ -70,6 +70,15 @@ services:
           Resource: arn:aws:sqs:us-east-1:12345678:myqueue
 ```
 
+Use an existing IAM role to a task:
+
+NOTE: This will override any as above mentioned in-line IAM policys.
+```yaml
+services:
+  test:
+    x-aws-iam: arn:aws:iam::123456789:role/applicationTaskRole
+```
+
 ###### Logging
 Pass options to awslogs driver
 ```yaml
--- a/ecs/cloudformation.go
+++ b/ecs/cloudformation.go
@@ -170,16 +170,21 @@ func (b *ecsAPIService) convert(ctx context.Context, project *types.Project) (*c
 }
 
 func (b *ecsAPIService) createService(project *types.Project, service types.ServiceConfig, template *cloudformation.Template, resources awsResources) error {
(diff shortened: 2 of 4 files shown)


ndeloof:
As discussed on https://github.com/docker/compose-cli/issues/1411 I'm not in favour to introduce more `x-aws-` extensions that will never get into the compose specification.
On the other hand I understand your use case.

A possible way to address this would be to rely on the already supported `x-aws-role`, but let this one define a plain ARN in addition to a policy document. You can use `arn.IsARN()` to check the attribute value defines an ARN, vs a nested document.

rosven:
Found another solution to this problem! :) Pm me if you ever have a similar issue

ndeloof:
@rosven could you please add a note on #1500 to explain the solution you adopted? Could be usefull for others, and maybe we could include this in the documentation

rosven:
@ndeloof I simply bypassed it by looking over my role and extended the role policys instead. There seems to be a way to overlay a specific rolename in `x-aws-cloudformation` but it wont be possible due to how the stack is updated/created.

https://github.com/docker/compose-cli/blob/b6df84f56a270cf3fb35bc67d4e8f0eca258f903/ecs/sdk.go#L363-L371

Seems that `capabilities` should be set to `CAPABILITY_NAMED_IAM` if that should work.

https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateStack.html
https://github.com/docker/compose-cli/blob/b6df84f56a270cf3fb35bc67d4e8f0eca258f903/ecs/sdk.go#L369-L371

But since its hardcoded to `CapabilityCapabilityIam`
https://github.com/aws/aws-sdk-go/blob/3961e6ef50ca64ae09f0e935f7aa7ffbffde5f5b/service/cloudformation/api.go#L17784-L17788

if it was made possible to choose one of:

https://github.com/aws/aws-sdk-go/blob/3961e6ef50ca64ae09f0e935f7aa7ffbffde5f5b/service/cloudformation/api.go#L17783-L17792

It would probably be possible to set an explicit RoleName when it is created if I don't misunderstand how this works.

PR #1513, opened 2021-04-08
Avoid panic: concurrent map writes

gtardif:

**What I did**
use mutex to sync concurrent access to map (quick global pass looking for `eg.Go(` )

**Related issue**
https://github.com/docker/compose-cli/runs/2295818799 (first time seen)

### DIFF
--- a/local/compose/images.go
+++ b/local/compose/images.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 
 	moby "github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/filters"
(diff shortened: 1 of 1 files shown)


local/compose/ps.go
```diff
@@ -61,6 +63,7 @@ func (s *composeService) Ps(ctx context.Context, projectName string, options com
 				health = inspect.State.Health.Status
 			}
 
+			l.Lock()
```
ndeloof, line 66: not necessary AFAIK as we pre-allocated the slice and each gortouine will run with a distinct index

local/compose/images.go
```diff
@@ -66,7 +68,9 @@ func (s *composeService) Images(ctx context.Context, projectName string, options
 			if err != nil {
 				return err
 			}
+			l.Lock()
```
karolz-ms, line 71: Why not just use https://golang.org/pkg/sync/#Map ? ☺️

gtardif: Yes, actually a good suggestion... I was about to do this, and then Map is totally untyped, so I prefer the 3 lines of explicit locking here, and not have to cast things further down (and the doc says "Most code should use a plain Go map instead, with separate locking" 🙄  )

ndeloof (approved):

Issue #1515, opened 2021-02-25
`docker version -f json` not (only) returning json if the daemon is not running


chevdor:
**Description**

The expectation when running `docker version -f json` is to get an answer formatted as json.
Currently, it does work well as long as the daemon is running. If the daemon is not working however, the output is polluted by `Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?`, preventing json parsing of the response.

**Steps to reproduce the issue:**
1. Make sure docker is installed...
2. STOP the daemon
3. Run `docker version -f json`

**Describe the results you received:**

```
{"Client":{"Platform":{"Name":"Docker Engine - Community"},"CloudIntegration":"1.0.7","Version":"20.10.2","ApiVersion":"1.41","DefaultAPIVersion":"1.41","GitCommit":"2291f61","GoVersion":"go1.13.15","Os":"darwin","Arch":"amd64","BuildTime":"Mon Dec 28 16:12:42 2020","Context":"default","Experimental":true},"Server":null}
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

**Describe the results you expected:**

```
{"Client":{"Platform":{"Name":"Docker Engine - Community"},"CloudIntegration":"1.0.7","Version":"20.10.2","ApiVersion":"1.41","DefaultAPIVersion":"1.41","GitCommit":"2291f61","GoVersion":"go1.13.15","Os":"darwin","Arch":"amd64","BuildTime":"Mon Dec 28 16:12:42 2020","Context":"default","Experimental":true},"Server":null}
```

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

I let you parse the output :)
```
{"Client":{"Platform":{"Name":"Docker Engine - Community"},"CloudIntegration":"1.0.7","Version":"20.10.2","ApiVersion":"1.41","DefaultAPIVersion":"1.41","GitCommit":"2291f61","GoVersion":"go1.13.15","Os":"darwin","Arch":"amd64","BuildTime":"Mon Dec 28 16:12:42 2020","Context":"default","Experimental":true},"Server":null}
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
[...]

PR #1516, opened 2021-04-09
don't include stderr in moby exec output

ndeloof:
**What I did**
only capture stdout when we run a moby command, so that we don't get stderr mixed

**Related issue**
close https://github.com/docker/compose-cli/issues/1515

### DIFF
--- a/cli/mobycli/exec.go
+++ b/cli/mobycli/exec.go
@@ -136,5 +136,6 @@ func ExecSilent(ctx context.Context, args ...string) ([]byte, error) {
 		args = os.Args[1:]
 	}
 	cmd := exec.CommandContext(ctx, ComDockerCli, args...)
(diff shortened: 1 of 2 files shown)


cli/mobycli/exec.go
```diff
@@ -136,5 +137,9 @@ func ExecSilent(ctx context.Context, args ...string) ([]byte, error) {
 		args = os.Args[1:]
 	}
 	cmd := exec.CommandContext(ctx, ComDockerCli, args...)
-	return cmd.CombinedOutput()
+	var b bytes.Buffer
```
gtardif, line 140: Can't we just replace `cmd.CombinedOutput()` by `cmd.CombinedOutput()` here to do this ? 
In terms of impact, it's OK as version is the only user of `ExecSilent()`, OK

PR #1522, opened 2021-04-09
Ignore previous CFM events on update to avoid false positive errors

### LINKED ISSUE #1112: Compose CLI sometimes returns 1 instead of 0 on success
**Description**
The Compose CLI sometimes returns 1 instead of 0 on a successful `docker compose up` update of an ECS stack.

Initial report was on the community Slack:
> I’m surprised by the error codes reported by docker compose -D up .  I would expect a successful deploy to return zero unless there is a failure.  My experience after several experiments is that when successfully deploying an update (such as simply changing the tag of a container) then the return code is 1.  Is this expected?  Is it documented somewhere?

> Hmm.. Sounds like a bug. Do you have a specific flow that you run and always get a 1 return code for?

> Yes.  I have consistently seen a one returned by docker compose -D -context mycontext up when the only change is the tag on one of the service images.  The deploy succeeds and the scope of changes reported by -D seems roughly appropriate, but the return code is always 1.

**Describe the results you expected:**
Return 0 on success and another value on error with a relevant error message.

defyrlt:
ECS `up` is pulling all events for CFM stack, which may be having some errors in
the history. For example, if CFM stack update was ever cancelled, `up` will be
exiting with code 1 on every run after that cancellation.

Fixes #1112

### DIFF
--- a/ecs/up.go
+++ b/ecs/up.go
@@ -93,6 +93,16 @@ func (b *ecsAPIService) Up(ctx context.Context, project *types.Project, options 
 	if err != nil {
 		return err
 	}
+
+	var previousEvents []string
+	if update {
+		var err error
+		previousEvents, err = b.previousStackEvents(ctx, project.Name)
+		if err != nil {
+			return err
+		}
+	}
+
 	operation := stackCreate
 	if update {
 		operation = stackUpdate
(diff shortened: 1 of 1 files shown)


ndeloof:
please sign-off your commit so we can accept this contribution (git commit -s)

defyrlt:
@ndeloof i believe the commit is already [signed off](https://github.com/docker/compose-cli/pull/1522/commits/dc6bf1a0f3597e97f1e48d0e2d05f9169a1c8422), DCO check is [failing](https://github.com/docker/compose-cli/pull/1522/checks?check_run_id=2309885727) because of the name mismatch

ndeloof:
indeed, then please fix your git username (or manually edit sign-off-by) so that DCO check can validate your commits

defyrlt:
@ndeloof done, all passing

ndeloof (approved):

defyrlt:
@ndeloof thanks :) do you know when this will be released?

Issue #1526, opened 2021-04-12
docker compose up command failed to use credential info from docker


albertjone:
**Description**
i was use `docker compose up` to create containers following https://docs.stackstorm.com/install/docker.html, but it failed with below error:
failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to build LLB: failed to load cache key: failed to copy: httpReaderSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/stackstorm/st2auth/manifests/sha256:dbc9b1bc01725ca17d01e22873287899c0def080bf8f21574a093a2fbc6547a2: 429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
And i was behind a lb.

**Steps to reproduce the issue:**
1. git clone https://github.com/stackstorm/st2-docker
2. cd st2-docker
3. docker-compose up
4. then you get the above error

**Describe the results you received:**
got the error info, even i login in. But i use `docker-compose up` is ok.

**Describe the results you expected:**
After i docker login with my user name and password, i am able to docker compose up the st2.

**Additional information you deem important (e.g. issue happens only occasionally):**
it always happen.
**Output of `docker version`:**

```
Docker version 20.10.0, build 7287ab3
```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```
[]
context "context-name" does not exist
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.4.2-docker)
  scan: Docker Scan (Docker Inc., v0.5.0)

Server:
 Containers: 35
  Running: 19
  Paused: 0
  Stopped: 16
 Images: 53
 Server Version: 19.03.12
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
[...]

PR #1529, opened 2021-04-12
pass config file to buildkit for authentication

ndeloof:
**What I did**
Pass client's config file as Auth to buildkit

**Related issue**
close https://github.com/docker/compose-cli/issues/1526

### DIFF
--- a/local/compose/build.go
+++ b/local/compose/build.go
@@ -30,12 +30,14 @@ import (
 	"github.com/docker/buildx/driver"
 	_ "github.com/docker/buildx/driver/docker" // required to get default driver registered
 	"github.com/docker/buildx/util/progress"
+	cliconfig "github.com/docker/cli/cli/config"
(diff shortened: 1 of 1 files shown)


local/compose/build.go
```diff
@@ -36,6 +35,8 @@ import (
 	"github.com/docker/buildx/driver"
 	_ "github.com/docker/buildx/driver/docker" // required to get default driver registered
 	"github.com/docker/buildx/util/progress"
+	. "github.com/docker/cli/cli/config"
```
chris-crone, line 38: Not sure what we've done elsewhere but do we want to load this into the current namespace? It might be cleaner to import this as cliconfig
```suggestion
	cliconfig "github.com/docker/cli/cli/config"
```

chris-crone (approved):
Import change then LGTM

Issue #1524, opened 2021-04-11
Invalid mount config with "docker compose", but not with "docker-compose"


jochenw:
[docker-compose.txt](https://github.com/docker/compose-cli/files/6292633/docker-compose.txt)

The attached file works fine when running

        docker-compose up

If I use

        docker compose up

instead, I get the message

        Error response from daemon: invalid mount config for type "bind": field VolumeOptions must not be specified

Output of "docker version":

        λ docker version
        Client: Docker Engine - Community
         Cloud integration: 1.0.12
         Version:           20.10.5
         API version:       1.41
         Go version:        go1.13.15
         Git commit:        55c4c88
         Built:             Tue Mar  2 20:14:53 2021
         OS/Arch:           windows/amd64
         Context:           default
         Experimental:      true

        Server: Docker Engine - Community
         Engine:
          Version:          20.10.5
          API version:      1.41 (minimum version 1.12)
          Go version:       go1.13.15
          Git commit:       363e9a8
          Built:            Tue Mar  2 20:15:47 2021
          OS/Arch:          linux/amd64
          Experimental:     false
         containerd:
          Version:          1.4.4
          GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
         runc:
          Version:          1.0.0-rc93
          GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
         docker-init:
          Version:          0.19.0
          GitCommit:        de40ad0

Output of "docker context show"

        default

Output of "docker info"

λ docker info                                                                                                            
Client:                                                                                                                  
 Context:    default                                                                                                     
 Debug Mode: false                                                                                                       
 Plugins:                                                                                                                
  app: Docker App (Docker Inc., v0.9.1-beta3)                                                                            
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)                                                               
  scan: Docker Scan (Docker Inc., v0.6.0)                                                                                
                                                                                                                         
Server:                                                                                                                  
 Containers: 4                                                                                                           
  Running: 0                                                                                                             
  Paused: 0                                                                                                              
  Stopped: 4                                                                                                             
 Images: 5                                                                                                               
 Server Version: 20.10.5                                                                                                 
 Storage Driver: overlay2                                                                                                
  Backing Filesystem: extfs                                                                                              
  Supports d_type: true                                                                                                  
  Native Overlay Diff: true                                                                                              
 Logging Driver: json-file                                                                                               
[...]

PR #1531, opened 2021-04-13
ignore mount options which don't match type

ndeloof:
**What I did**
Ignore mount options that do not apply to the selected mount type
warn user about configuration mismatch

**Related issue**
close https://github.com/docker/compose-cli/issues/1524

### DIFF
--- a/local/compose/create.go
+++ b/local/compose/create.go
@@ -774,16 +774,48 @@ func buildMount(project types.Project, volume types.ServiceVolumeConfig) (mount.
 		}
 	}
 
+	bind, vol, tmpfs := buildMountOptions(volume)
+
 	return mount.Mount{
 		Type:          mount.Type(volume.Type),
 		Source:        source,
 		Target:        volume.Target,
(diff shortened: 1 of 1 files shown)


local/compose/create.go
```diff
@@ -774,18 +774,50 @@ func buildMount(project types.Project, volume types.ServiceVolumeConfig) (mount.
...
+
 	return mount.Mount{
 		Type:          mount.Type(volume.Type),
 		Source:        source,
 		Target:        volume.Target,
 		ReadOnly:      volume.ReadOnly,
 		Consistency:   mount.Consistency(volume.Consistency),
-		BindOptions:   buildBindOption(volume.Bind),
-		VolumeOptions: buildVolumeOptions(volume.Volume),
-		TmpfsOptions:  buildTmpfsOptions(volume.Tmpfs),
```
chris-crone, line 785: Are these functions in use elsewhere? Otherwise we can remove them

chris-crone (approved):
LGTM

Issue #1517, opened 2021-04-09
Docker scan advertising breaks output from `docker build -q`


RedwanFox:
- [ x] I have tried with the latest version of Docker Desktop
  - [ x] I have tried disabling enabled experimental features
  - [ ] I have uploaded Diagnostics
  - Diagnostics ID:

In latest version of Docker Desktop for windows `docker scan` advertising appears in stdout in every command output breaks automatisation pipelines

### Actual behavior

in git-bash
```
…/server-build-env
> x=$(docker build . -q) 
…/server-build-env 
> echo $x
sha256:95cde0989d09c3c13b36eda8b4415179e3631b47dd461e383b550945a3a2318e Use 'docker scan' to run Snyk tests against images to find vulnerabilities and learn how to fix them

```

### Expected behavior

```
…/server-build-env
> x=$(docker build . -q) 
…/server-build-env 
> echo $x
sha256:95cde0989d09c3c13b36eda8b4415179e3631b47dd461e383b550945a3a2318e
```

### Information

  - Windows Version: windows 10 20h2
  - Docker Desktop Version: 3.3.0
  - WSL2 or Hyper-V backend?  hyper-v
  - Are you running inside a virtualized Windows e.g. on a cloud server or a VM: no

### Steps to reproduce the behavior

See actual behaviour

slonopotamus:
Suggested fixes:

a. Stop outputting this ad text completely.
b. Write it to stderr instead of stdout. Though this still might break some automated scripts.

PR #1532, opened 2021-04-13
Fix scan message with quiet flag

chris-crone:
**What I did**
* Do not print the scan message when the quiet flag is specified
* Print the scan message to stderr instead of stdout

**Related issue**
Fixes: #1517 

/test-windows


ndeloof (approved):

chris-crone:
The Windows tests failing looks unrelated: https://github.com/docker/compose-cli/pull/1532/checks?check_run_id=2333539095#step:9:33

Can we merge without it?

PR #1539, opened 2021-04-14
[refactoring] package `compose` as a CLI plugin on local backend

ndeloof:
**What I did**
pass `compose.Service` to cobra command builder, so that this code doesn't depend on the backend mechanism and is easier to migrate into a compose cli plugin under docker/compose repository

This is based on earlier work on https://github.com/docker/compose-cli/pull/1397. Some of the commits have been cherry-picked.
Main difference is: compose commands have been refactored to adopt a "Dependency Injection by Constructor" pattern, i.e. :
- `FooCommand()` to create cobra compose command `foo` receive the backend's `service.Compose` implementation as parameter, and does not have to know about the context/backend logic. 
- From top-level `main.go` (CLI plugin) the `local.Backend` implementation is created by injecting Docker API client and config file (provided by the CLI plugin framework) as constructor parameters
- Compose cobra command are wrapped with an adaptor to manage exit code in a consistent way with metrics expected exit codes.

/test-aci
/test-ecs
/test-windows

### DIFF
--- a/cli/metrics/definitions.go
+++ b/cli/metrics/definitions.go
@@ -55,3 +55,26 @@ var (
 	// PullFailure failure while pulling image
 	PullFailure = FailureCategory{MetricsStatus: PullFailureStatus, ExitCode: 18}
 )
+
+//ByExitCode retrieve FailureCategory based on command exit code
+func ByExitCode(exitCode int) FailureCategory {
+	switch exitCode {
+	case 0:
+		return FailureCategory{MetricsStatus: SuccessStatus, ExitCode: 0}
+	case 14:
+		return FileNotFoundFailure
+	case 15:
+		return ComposeParseFailure
+	case 16:
+		return CommandSyntaxFailure
+	case 17:
+		return BuildFailure
+	case 18:
+		return PullFailure
+	case 130:
+		return FailureCategory{MetricsStatus: CanceledStatus, ExitCode: exitCode}
+	default:
+		return FailureCategory{MetricsStatus: FailureStatus, ExitCode: exitCode}
+	}
+
+}
--- a/utils/e2e/framework.go
+++ b/utils/e2e/framework.go
@@ -85,6 +85,13 @@ func newE2eCLI(t *testing.T, binDir string) *E2eCLI {
 		_ = os.RemoveAll(d)
 	})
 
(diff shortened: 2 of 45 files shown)


ndeloof:
test release on Docker Desktop: https://github.com/docker/pinata/pull/15646

cli/metrics/definitions.go
```diff
@@ -55,3 +55,26 @@ var (
...
+		return BuildFailure
+	case 18:
+		return PullFailure
+	case 130:
+		return FailureCategory{MetricsStatus: CanceledStatus, ExitCode: exitCode}
+	default:
+		return FailureCategory{MetricsStatus: FailureStatus, ExitCode: exitCode}
+	}
+
+}
```
ulyssessouza, line 80: This could be simplified by a const byExitCode of type map[int]FailureCategory.

utils/e2e/framework.go
```diff
@@ -85,6 +85,17 @@ func newE2eCLI(t *testing.T, binDir string) *E2eCLI {
 		_ = os.RemoveAll(d)
 	})
 
+	_ = os.MkdirAll(filepath.Join(d, "cli-plugins"), 0755)
+	composePluginFile := "docker-compose"
+	if runtime.GOOS == "windows" {
+		composePluginFile += ".exe"
+	}
+	composePlugin, _ := findExecutable(composePluginFile, []string{"../../bin", "../../../bin"})
```
ulyssessouza, line 93: filepath.Join

utils/e2e/framework.go
```diff
@@ -117,7 +128,7 @@ func SetupExistingCLI() (string, func(), error) {
 		return "", nil, err
 	}
 
-	bin, err := findExecutable([]string{"../../bin", "../../../bin"})
+	bin, err := findExecutable(DockerExecutableName, []string{"../../bin", "../../../bin"})
```
ulyssessouza, line 131: filepath.Join

cli/metrics/definitions.go
ndeloof: I'm not sure which one is "simpler". At least, not significantly more compact :P 
```
case x: 
    return y
default:
    ...
``` 
with 
```
map:= map[int]FailureCategory{
  x: y,
} 
if v, ok := map[x]; ok {
   return y
}
...
```

gtardif (approved):

PR #1541, opened 2021-04-14
don't set aliasses if no network binding was declared

ndeloof:
**What I did**
Don't build networkConfig if no network binding is declared
this is same as https://github.com/docker/compose/blob/master/compose/service.py#L817-L818

**Related issue**
close https://github.com/docker/dev-tooling-team/issues/301


ulyssessouza (approved):
LGTM

PR #1544, opened 2021-04-14
only start container once we have a wait channel set, to prevent race

ndeloof:
**What I did**
change API call order so we wait for container next exit _before_ we start it, otherwise a container to just run `/bin/true` might terminate before we invoke wait, resulting in daemon error "container xx is not running"

### DIFF
--- a/local/compose/convergence.go
+++ b/local/compose/convergence.go
@@ -337,16 +337,14 @@ func (s *composeService) connectContainerToNetwork(ctx context.Context, id strin
 }
 
 func (s *composeService) isServiceHealthy(ctx context.Context, project *types.Project, service string) (bool, error) {
-	containers, err := s.apiClient.ContainerList(ctx, moby.ContainerListOptions{
-		Filters: filters.NewArgs(
-			projectFilter(project.Name),
-			serviceFilter(service),
-		),
-	})
+	containers, err := s.getContainers(ctx, project.Name, oneOffExclude, false, service)
 	if err != nil {
 		return false, err
 	}
 
+	if len(containers) == 0 {
+		return false, nil
+	}
 	for _, c := range containers {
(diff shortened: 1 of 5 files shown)


local/compose/resize.go
```diff
@@ -28,13 +28,13 @@ import (
...
 )
 
-func (s *composeService) monitorTTySize(ctx context.Context, container string, resize func(context.Context, string, moby.ResizeOptions) error) error {
+func (s *composeService) monitorTTySize(ctx context.Context, container string, resize func(context.Context, string, moby.ResizeOptions) error) {
 	err := resize(ctx, container, moby.ResizeOptions{ // nolint:errcheck
 		Height: uint(goterm.Height()),
 		Width:  uint(goterm.Width()),
 	})
 	if err != nil {
-		return err
```
chris-crone, line 37: Curious why you dropped the error here?

local/compose/convergence.go
```diff
@@ -356,13 +354,15 @@ func (s *composeService) isServiceHealthy(ctx context.Context, project *types.Pr
 			return false, fmt.Errorf("container for service %q has no healthcheck configured", service)
 		}
 		switch container.State.Health.Status {
-		case "starting":
+		case moby.Starting:
 			return false, nil
-		case "unhealthy":
+		case moby.Unhealthy:
 			return false, nil
+		case moby.Healthy:
```
chris-crone, line 361: nit: If we only care about healthy, we could make this a simple if or a switch ... default

chris-crone (approved):
One question and a nit but LGTM

local/compose/resize.go
ndeloof: as we monitor resize signal, we don't know when container will terminate and resize would fail.
same as docker/cli@04dad42/cli/command/container/tty.go#L71 which always return nil

chris-crone (approved):

PR #1554, opened 2021-04-15
fix race condition with init containers

ndeloof:
**What I did**
replace use of WaitContainer with polling on Container.State
there's a risk container terminate before we run WaitContainer, and we never catch "next exit" event. This typically happen with e2e test to run `echo hello`.


chris-crone (approved):
Tested locally and this seems to fix it. Thanks!

gtardif:
I also had the issue with TestInitContainer on main, quite reproductible, and this PR fixes it. But for some reason the CI here is still failing

gtardif:
`TestInitContainer` seems to be fixed, butwith this PR `TestIPC` is now failing

PR #1581, opened 2021-04-21
Fix down not working if run from a different folder than up, because container labels stored relative paths

gtardif:
This is breaking one Desktop test, and I also ran into this when cleaning up test compose apps manually. 

**What I did**
Ensure we store absolute paths in container labels, and do not try to reconstruct composefile paths with unrealistic logic when reading labels

**Related issue**
https://ci-next.docker.com/teams-desktop/blue/organizations/jenkins/desktop%2Fdesktop-test-single-pr-mac/detail/desktop-test-single-pr-mac/84670/pipeline

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/awslabs/goformation/v4 v4.15.6
 	github.com/buger/goterm v1.0.0
 	github.com/cnabio/cnab-to-oci v0.3.1-beta1
-	github.com/compose-spec/compose-go v0.0.0-20210421142148-66b18e67d2c0
+	github.com/compose-spec/compose-go v0.0.0-20210422124345-d4bf0e1bfea5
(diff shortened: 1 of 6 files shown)


ndeloof (commented):
LGTM but IMHO it would make better sense to enforce `project.WorkingDir` being absolute in compose-go `normalize` so we only use absolute paths everywhere

gtardif:
This must also include the configfiles, that are resolved potentially by looking working dir and parent folder. I need to check if `normalize()` is executed with these config files already resolved or if the resolution happens later

gtardif:
Will use https://github.com/compose-spec/compose-go/pull/128

ndeloof (approved):

gtardif:
https://github.com/docker/compose-cli/pull/1581/checks?check_run_id=2411063172#step:9:22 already red on main, to be fixed in a separate PR... 😢

PR #1588, opened 2021-04-23
On default context, "compose" is implemented by CLI Plugin

ndeloof:
**What I did**
removed "compose" command for default context type, being implemented as a CLI plugin

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -218,9 +218,13 @@ func main() {
 
 	root.AddCommand(
 		run.Command(ctype),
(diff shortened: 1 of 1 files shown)


gtardif (approved):
makes sense, as a safety on top of compose removed from https://github.com/docker/compose-cli/pull/1588/files#diff-d4fceb1903b50aefe27a4159c53d34052bd7a2b43eaebc865ea120fa80a71e19R63. 
(And we already have e2e tests running compose CLI plugin without the compose-cli wrapper, and the other way around, compose-cli cloud integration without compose-cli)

PR #1594, opened 2021-04-26
Fix race in TestAttachRestart : wait for logs to appear rather than have a fixed TTL for the application

gtardif:

**What I did**
set the application to live forever, and wait for logs to appear, then down the app.

**Related issue**
Depending on other tests running in //, the previous condition was not met as the application was going down before having time to restart 3 times.
Seen almost always locally( (Desktop mac & windows), and rarely in the CI


ndeloof (approved):

local/e2e/compose/compose_test.go
```diff
@@ -192,15 +192,19 @@ func TestDownComposefileInParentFolder(t *testing.T) {
...
 	defer c.RunDockerOrExitError("compose", "-p", "attach-restart", "down")
-	output := res.Stdout()
 
-	exitRegex := regexp.MustCompile("another_1 exited with code 1")
-	assert.Equal(t, len(exitRegex.FindAllStringIndex(output, -1)), 3, res.Combined())
+	c.WaitForCondition(func() (bool, string) {
+		out := res.Stdout()
+		debug := res.Combined()
+		exitRegex := regexp.MustCompile("another_1 exited with code 1")
+		return len(exitRegex.FindAllStringIndex(out, -1)) == 3, fmt.Sprintf("'another_1 exited with code 1' not found 3 times in : \n%s\n", debug)
```
mat007, line 203: Do we need a regex for this? Can’t we use [strings.Count](https://golang.org/pkg/strings/#Count)?

gtardif: indeed, not sure why we had a regex initially there... fixed

PR #1603, opened 2021-04-27
Separate compose CLI plugin version from Cloud integration version

gtardif:

**What I did**
Separate compose CLI plugin version

**Related issue**
https://github.com/docker/dev-tooling-team/issues/285

### DIFF
--- a/cli/cmd/compose/version.go
+++ b/cli/cmd/compose/version.go
@@ -34,9 +34,10 @@ type versionOptions struct {
 func versionCommand() *cobra.Command {
 	opts := versionOptions{}
 	cmd := &cobra.Command{
-		Use:   "version",
-		Short: "Show the Docker Compose version information",
-		Args:  cobra.MaximumNArgs(0),
+		Use:    "version",
+		Short:  "Show the Docker Compose version information",
+		Args:   cobra.MaximumNArgs(0),
+		Hidden: true,
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			runVersion(opts)
 			return nil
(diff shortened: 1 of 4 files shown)


cli/cmd/compose/version.go
```diff
@@ -34,9 +34,10 @@ type versionOptions struct {
 func versionCommand() *cobra.Command {
 	opts := versionOptions{}
 	cmd := &cobra.Command{
-		Use:   "version",
-		Short: "Show the Docker Compose version information",
-		Args:  cobra.MaximumNArgs(0),
+		Use:    "version",
+		Short:  "Show the Docker Compose version information",
+		Args:   cobra.MaximumNArgs(0),
+		Hidden: true,
```
mat007, line 40: Why would we want to hide this command?

mat007 (approved):
Makes sense!

cli/cmd/compose/version.go
ndeloof: Should be hidden when used inside compose-cli but not for cli-plugin ...

ndeloof (approved):

gtardif:
> Should be hidden when used inside compose-cli but not for cli-plugin ...

Yes, not really sure what to do with this one, I'm not even sure users should use `docker compose version`, other commands don't have a version subcommand and we really added this as a way to get the compose version for the cli plugin & the redirect from `-compose`. 
Given it's been added just recently & is not very user facing, I prefer to not make this visible for now, it will be easier to change it to visible if we want to.

gtardif:
Happy to discuss/revisit the visibility in a later PR

PR #1563, opened 2021-04-16
Create mount source if it does not exist

aiordache:
Create a directory if a bind-mount source does not exist to align with docker-compose.
```
services:
  frontend:
    image: nginx
    volumes:
      - /tmp/doesnotexist:/test
```
```
[example]$ docker-compose up -d
Creating network "example_default" with the default driver
Creating example_frontend_1 ... done
```
```
[example]$ ls -al /tmp | grep doesnotexist
drwxr-xr-x  2 root root      40 Apr 16 10:45 doesnotexist

```
```
[example]$ docker compose up -d
[+] Running 4/4
 ⠿ Network "example_default"     Created                                                                                                           0.0s
 ⠿ Container example_frontend_1  Started                                                                                                           0.7s
```
```
[example]$ ls -al /tmp | grep doesnotexist
drwxr-xr-x  2 anca users     40 Apr 16 10:45 doesnotexist

```

Closes https://github.com/docker/compose-cli/issues/1543

PR #1604, opened 2021-04-27
when bind mount has CreateHostPath set, use bind API

### LINKED ISSUE #1543: invalid mount config for type "bind": bind source path does not exist:
**Description**

When running with docker compose:  Trying to mount a folder into a container throws an error if the folder does not already exist in the host.
When running with docker-compose: Trying to mount a folder into a container automatically creates the folder if the folder does not already exist

The expectation would be that both commands work the same way

**Steps to reproduce the issue:**
1. Add a valid volume to a service in your docker-compose file but do not create the folder in the host file system
2. `docker compose up`
3. Error is displayed
4. `docker-compose up`
5. Application comes up normally

**Describe the results you received:**

Got error : `Error response from daemon: invalid mount config for type "bind": bind source path does not exist: <path to folder that does not currently exist>`

**Describe the results you expected:**

Folder gets automatically created just like with docker-compose

**Output of `docker version`:**

Client: Docker Engine - Community
 Cloud integration: 1.0.9
 Version:           20.10.5
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        55c4c88
 Built:             Tue Mar  2 20:13:00 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.5
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       363e9a8
  Built:            Tue Mar  2 20:15:47 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

**Additional environment details (AWS ECS, Azure ACI, local, etc.):**

local

ndeloof:
**What I did**

Use `bind` when volume definition require host path to be created when missing, which isn't supported by `mount`

Closes #1543
Supersede #1563
https://docker.atlassian.net/browse/IL-79

### DIFF
--- a/local/compose/create.go
+++ b/local/compose/create.go
@@ -574,22 +574,25 @@ func (s *composeService) buildContainerVolumes(ctx context.Context, p types.Proj
 		return nil, nil, nil, err
 	}
 
-	// filter binds and volumes mount targets
 	volumeMounts := map[string]struct{}{}
 	binds := []string{}
+MOUNTS:
 	for _, m := range mountOptions {
-
-		if m.Type == mount.TypeVolume {
-			volumeMounts[m.Target] = struct{}{}
-			if m.Source != "" {
-				binds = append(binds, fmt.Sprintf("%s:%s:rw", m.Source, m.Target))
+		volumeMounts[m.Target] = struct{}{}
+		// `Bind` API is used when host path need to be created if missing, `Mount` is preferred otherwise
+		if m.Type == mount.TypeBind {
+			for _, v := range service.Volumes {
+				if v.Target == m.Target && v.Bind != nil && v.Bind.CreateHostPath {
+					mode := "rw"
+					if m.ReadOnly {
+						mode = "ro"
+					}
+					binds = append(binds, fmt.Sprintf("%s:%s:%s", m.Source, m.Target, mode))
+					continue MOUNTS
+				}
 			}
 		}
-	}
-	for _, m := range mountOptions {
-		if m.Type == mount.TypeBind || m.Type == mount.TypeTmpfs {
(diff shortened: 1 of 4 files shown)


local/compose/create.go
```diff
@@ -574,22 +574,25 @@ func (s *composeService) buildContainerVolumes(ctx context.Context, p types.Proj
...
+MOUNTS:
 	for _, m := range mountOptions {
-
-		if m.Type == mount.TypeVolume {
-			volumeMounts[m.Target] = struct{}{}
-			if m.Source != "" {
-				binds = append(binds, fmt.Sprintf("%s:%s:rw", m.Source, m.Target))
+		volumeMounts[m.Target] = struct{}{}
+		// `Bind` API is used when host path need to be created if missing, `Mount` is preferred otherwise
+		if m.Type == mount.TypeBind {
```
gtardif, line 583: To confirm: so we keep "bind" mounts only in the case of "createHostPath == true", otherwise default to the mount API even when `m.Type == mount.TypeBind`, is this correct ?

ndeloof: yes indeed, which will only happen if user did define an explicit mount with full syntax, 
... which is probably less than 1% usages, everybody just uses `source:target` short syntax

gtardif (approved):

Issue #1607, opened 2021-04-28
exec does not forward exit status


trkoch:
**Description**
`docker compose exec` does not return exit status of process. `run` seems to work as expected.

**Steps to reproduce the issue:**
docker compose exec <service> bash -c "false"; echo $?

**Describe the results you received:**
0

**Describe the results you expected:**
1 (i.e. same as `docker-compose exec <service> bash -c "false"; echo $?` mind the `-`).

**Additional information you deem important (e.g. issue happens only occasionally):**
Possibly related to https://github.com/docker/compose-cli/issues/1552.

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Cloud integration: 1.0.12
 Version:           20.10.5
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        55c4c88
 Built:             Tue Mar  2 20:13:00 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.5
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       363e9a8
  Built:            Tue Mar  2 20:15:47 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```
#
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  scan: Docker Scan (Docker Inc., v0.6.0)

Server:
 Containers: 3
  Running: 3
  Paused: 0
  Stopped: 0
 Images: 29
 Server Version: 20.10.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
[...]

PR #1609, opened 2021-04-28
docker compose exec to return command exit code

ndeloof:
**What I did**
Inspect exec after completion to return exitcode on `docker compose exec`

**Related issue**
close https://github.com/docker/compose-cli/issues/1607

### DIFF
--- a/cli/cmd/compose/exec.go
+++ b/cli/cmd/compose/exec.go
@@ -22,6 +22,7 @@ import (
 	"os"
 
 	"github.com/containerd/console"
+	"github.com/docker/cli/cli"
 	"github.com/spf13/cobra"
 
 	"github.com/docker/compose-cli/api/compose"
@@ -108,5 +109,13 @@ func runExec(ctx context.Context, backend compose.Service, opts execOpts) error 
 		execOpts.Writer = con
 		execOpts.Reader = con
 	}
-	return backend.Exec(ctx, project, execOpts)
+	exitCode, err := backend.Exec(ctx, project, execOpts)
+	if exitCode != 0 {
+		errMsg := ""
+		if err != nil {
+			errMsg = err.Error()
+		}
+		return cli.StatusError{StatusCode: exitCode, Status: errMsg}
+	}
+	return err
 }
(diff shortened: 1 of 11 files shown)


cli/cmd/compose/exec.go
```diff
@@ -108,5 +109,13 @@ func runExec(ctx context.Context, backend compose.Service, opts execOpts) error
...
 		execOpts.Reader = con
 	}
-	return backend.Exec(ctx, project, execOpts)
+	exitCode, err := backend.Exec(ctx, project, execOpts)
+	if exitCode != 0 {
+		errMsg := ""
+		if err != nil {
+			errMsg = err.Error()
+		}
+		return cli.StatusError{StatusCode: exitCode, Status: errMsg}
```
lorenrh, line 118: can it be possible to have a non 0 exit code with no error 🤔 ?

local/compose/exec.go
```diff
@@ -42,10 +42,10 @@ func (s *composeService) Exec(ctx context.Context, project *types.Project, opts
 		),
 	})
 	if err != nil {
-		return err
+		return 0, err
 	}
 	if len(containers) < 1 {
-		return fmt.Errorf("container %s not running", getContainerName(project.Name, service, opts.Index))
+		return 0, fmt.Errorf("container %s not running", getContainerName(project.Name, service, opts.Index))
```
lorenrh, line 48: do we really want this to be 0? 🤔

ndeloof: with another language I would have used `Either(ExitCode, Error)` but with golang we have to return both values for "everything went fine" and "oups" scenario, as a tuple.
and we can't use `nil` here until we make the signature to use `*int` which is not really better

ndeloof: as an additionl note, this is basically a copy/paste/adapt signature from https://github.com/docker/cli/blob/a32cd16160f1b41c1c4ae7bee4dac929d1484e59/cli/command/container/exec.go#L116
I'd like we can just rely on docker/cli for such interactions and don't have to duplicate code, maybe we can evolve in this direction making required changes to docker/cli once Compose 2.0 is a thing

lorenrh (approved):
LGTM 🌮

Issue #1602, opened 2021-04-27
docker compose build --build-arg ARG Not able to access ARG in the image while building


saikumar-everest:
My **Dockerfile.dev** has a command to install all gems.

**Usual command:**
docker-compose build --build-arg BUNDLE_GEMS__CONTRIBSYS__COM app

This is able to build the image successfully as it could access the sending build argument which is a secret to download a gem from server securely.

**Experimental command:**
docker compose build --build-arg BUNDLE_GEMS__CONTRIBSYS__COM app

But with this command replaced, I'm not able to build image. I get 401 unauthorised error while downloading a gem. I clearly know that we get this when we don't try to download a gem securely by providing a secret.

So I expect this experimental feature is not yet working as expected (docker-compose ≠ docker compose).

Liam-Johnston:
This same problem occurs when defining build args in `docker-compose.yml`

ndeloof:
actually a duplicate for https://github.com/docker/compose-cli/issues/1560

PR #1610, opened 2021-04-28
resolve build args without value from environment

ndeloof:
**What I did**
Resolve build args without value (`--build-arg FOO`) from environment

**Related issue**
close https://github.com/docker/compose-cli/issues/1602


cli/cmd/compose/build.go
```diff
@@ -85,11 +85,13 @@ func runBuild(ctx context.Context, backend compose.Service, opts buildOptions, s
 		return err
 	}
 
+	fmt.Println(opts.args)
```
mat007, line 88: Debug statement left over?

ndeloof: oups :)

lorenrh (approved):
LGTM 🎉

PR #1616, opened 2021-04-29
handle dependency involved by shared PID namespace

ndeloof:
**What I did**
detect service dependency involved by `pid: "service:foo`


lorenrh (approved):
🍬 ✨

gtardif (approved):

Issue #1553, opened 2021-04-15
Running docker compose down -v (--volumes) command does not remove created docker volumes


afedotov:
**Description**

When running `docker compose down` command with `--volumes` option, volumes created by `docker compose up` command does not removed.

**Steps to reproduce the issue:**

1. Take any **docker-compose.yml** file which contains 'volumes' section with volumes mounted to any compose services
2. Run `docker compose up -d`
3. Wait until containers are started
4. Run `docker compose down -v`
5. Run `docker volume ls` and see that volumes are not removed by previous compose down command

**Describe the results you received:**

```
$ docker volume ls
DRIVER    VOLUME NAME

$ docker compose up -d
[+] Running 3/3
 ⠿ Network "keycloak_default"       Created                                                              3.9s
 ⠿ Volume "keycloak_keycloak_data"  Created                                                              0.0s
 ⠿ Container keycloak_keycloak_1    Started                                                              3.1s

$ docker compose down -v
[+] Running 2/2
 ⠿ Container keycloak_keycloak_1  Removed                                                                1.8s
 ⠿ Network "keycloak_default"     Removed                                                                2.8s

$ docker volume ls
DRIVER    VOLUME NAME
local     keycloak_keycloak_data
```

**Describe the results you expected:**

Created volumes must be removed.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
$ docker version
Client: Docker Engine - Community
 Cloud integration: 1.0.12
 Version:           20.10.5
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        55c4c88
 Built:             Tue Mar  2 20:13:00 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.5
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       363e9a8
  Built:            Tue Mar  2 20:15:47 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
 Kubernetes:
  Version:          v1.19.7
[...]

PR #1618, opened 2021-04-30
down --volume to remove volume

ndeloof:
**What I did**
implement "down --volumes" to cleanup compose application named volumes

**Related issue**
close https://github.com/docker/compose-cli/issues/1553


lorenrh (approved):
It has tests! 💯😄

Issue #1391, opened 2021-03-04
consider introduction of `docker compose cp`


ndeloof:
see https://github.com/docker/compose/issues/5523 for context

this would align with `docker cp` targeting compose services

JulienTant:
i guess this can be closed

PR #1623, opened 2021-05-02
Add cp command

JulienTant:
**What I did**
Introducing a `docker compose cp` command on local context. It allows you to copy a file/folder between a service container and the local file system the same way `docker cp` does.

If you scaled your service, you can chose which container will receive the file using a `--index=X` as you would when using `docker compose exec`.

I also added a `--all`  flag as a second commit, allowing users to copy to all containers of a service. I added that as a separate commit so it can be discarded easily if we need more time to decide how to tackle this.

**Related issue**
https://github.com/docker/compose-cli/issues/1391

### DIFF
--- a/aci/compose.go
+++ b/aci/compose.go
@@ -80,6 +80,10 @@ func (cs *aciComposeService) UnPause(ctx context.Context, project string, option
 	return errdefs.ErrNotImplemented
 }
 
+func (cs *aciComposeService) Copy(ctx context.Context, project *types.Project, options compose.CopyOptions) error {
+	return errdefs.ErrNotImplemented
+}
+
 func (cs *aciComposeService) Up(ctx context.Context, project *types.Project, options compose.UpOptions) error {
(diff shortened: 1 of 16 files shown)


ndeloof (approved):
LGTM.
Would be nice to include an e2e test. I'm not a "test all the things" guy, so may sounds like "_do as I say, not as I do_" with me asking for it, but seems this can be easily tested with a simple `compose up` && `compose cp <toContainer>` && `compose cp <fromContainer>`
Can use https://github.com/docker/compose-cli/blob/main/local/e2e/compose/networks_test.go#L74-L90 as an example to adopt e2e framework

JulienTant:
> Would be nice to include an e2e test.
I'm on it, thanks for the example.

/offtopic: I discovered and learned docker watching your "Quoi d'neuf Docker ?", I'm thankful that you created this content.

ndeloof:
thanks again for your contribution

Issue #1620, opened 2021-04-30
Network aliases not work with `docker compose`


k-teruya26:
**Description:**
When using new `docker compose`, network aliases are not resolved.
But works with `docker-compose`.

**Steps to reproduce the issue:**
1. Create docker-compose.yml file.
```
version: "3.8"

services:
  container1:
    image: php:7.3.27-apache

  container2:
    image: php:7.3.27-apache
    networks:
      default:
        aliases:
          - alias-of-container2
```
2. Start services.
`docker compose up -d`
3. Send http request from container1 to container2 (alias).
`docker compose exec container1 curl http://alias-of-container2/`

**Describe the results you received:**
`curl: (6) Could not resolve host: alias-of-container2`

**Describe the results you expected:**
When using `docker-compose`, this result shown.
```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
...
```

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Cloud integration: 1.0.12
 Version:           20.10.5
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        55c4c88
 Built:             Tue Mar  2 20:13:00 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.5
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       363e9a8
  Built:            Tue Mar  2 20:15:47 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker context show`:**  

```
default
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  scan: Docker Scan (Docker Inc., v0.6.0)

Server:
 Containers: 2
  Running: 2
  Paused: 0
[...]

PR #1625, opened 2021-05-03
use highest priority network to declare network aliasses

ndeloof:
**What I did**
Fixed networkConfig computation to align with docker-compose
https://github.com/docker/compose/blob/aa45dedb3d5b18df35c5a9c9d54c06cee0cfafe0/compose/service.py#L816

**Related issue**
see https://github.com/docker/compose-cli/issues/1620

note: full-fix will require updates in compose-go to introduce service.networks.priority

### DIFF
--- a/local/e2e/compose/fixtures/network-alias/compose.yaml
+++ b/local/e2e/compose/fixtures/network-alias/compose.yaml
@@ -1,0 +1,15 @@
+services:
+
+  container1:
+    image: nginx
+
+  container2:
+    image: nginx
+    networks:
+      foo:
+        aliases:
+          - alias-of-container2
+
+networks:
+  foo:
+    name: bar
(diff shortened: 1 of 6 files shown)


local/e2e/compose/fixtures/network-alias/compose.yaml
```diff
@@ -0,0 +1,15 @@
+services:
+
+  container1:
+    image: nginx
+
+  container2:
+    image: nginx
```
gtardif, line 7: We could add a test case with network priority here as well

gtardif (approved):
LGTM, with an additional test case checking the network priority in addition to network aliases

gtardif:
Actually as you mentioned, I don't see how we can easily test behaviour change based on network priority... and compose-go has unit tests about these, so I suppose we're OK here

Issue #1560, opened 2021-04-16
`docker compose run -e FOO` does not set FOO


jamshid:
**Description**

`docker compose run` is not compatible with `docker run` and `docker-compose run` in passing the value of an exported environment variable with plain `-e FOO`. It requires the variable be set explicitly like `-e FOO=${FOO}`.

**Steps to reproduce the issue:**
1. Using following docker-compose.yml:
```
services:
  test:
    image: alpine
    command: env
```
2. See `docker compose` behave differently from `docker-compose`:
```
% export FOO=foo

% docker compose run -e FOO test       
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=50098a50e8f7
TERM=xterm
HOME=/root

% docker-compose run -e FOO test
Creating simple_test_run ... done
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=52b3e29bb74d
TERM=xterm
FOO=foo
HOME=/root
```

**Describe the results you received:**

above
**Describe the results you expected:**
above

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Cloud integration: 1.0.12
 Version:           20.10.5
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        55c4c88
 Built:             Tue Mar  2 20:13:00 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.5
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       363e9a8
  Built:            Tue Mar  2 20:15:47 2021
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
 Kubernetes:
  Version:          Unknown
  StackAPI:         Unknown
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
[...]

PR #1628, opened 2021-05-04
Support bypass envvars

ulyssessouza:
**What I did**
Support bypass envvars

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/1560

### DIFF
--- a/api/compose/api.go
+++ b/api/compose/api.go
@@ -245,22 +245,6 @@ func (e Event) String() string {
 
 }
 
-// EnvironmentMap return RunOptions.Environment as a MappingWithEquals
-func (opts *RunOptions) EnvironmentMap() types.MappingWithEquals {
-	environment := types.MappingWithEquals{}
-	for _, s := range opts.Environment {
-		parts := strings.SplitN(s, "=", 2)
-		key := parts[0]
-		switch {
-		case len(parts) == 1:
-			environment[key] = nil
-		default:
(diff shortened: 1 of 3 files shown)


api/compose/api.go
```diff
@@ -254,6 +255,9 @@ func (opts *RunOptions) EnvironmentMap() types.MappingWithEquals {
 		switch {
 		case len(parts) == 1:
 			environment[key] = nil
+			if v, ok := os.LookupEnv(key); ok {
```
ndeloof, line 258: don't use `os.LookupEnv` but `project.Environment` which combine all possible sources for variables (.env, os.Env, --env-file...) and can be extended/better controlled in the future

Also, consider using `MappingWithEquals.Resolve`

ndeloof (changes requested):

api/compose/api.go
ndeloof: see https://github.com/docker/compose-cli/pull/1610/files#diff-477fda25fb05d92fe2c8e3cb31fdc228a8f5597b9e26b22f7feeff8246404db7R45-R51 as an example you can reuse in `local/compose/run.go:RunOneOffContainer`

ndeloof: I just noticed this whole method could be simplified as it just reproduce existing `NewMappingWithEquals`

ndeloof (approved):

Issue #1629, opened 2021-05-04
`docker compose logs --follow` doesn't work


yordis:
**Steps to reproduce the issue:**
1. Run `docker compose logs --follow
2. Notice the issue

**Describe the results you received:**

```text
github.com/docker/compose-cli/local/compose.(*composeService).Logs.func1(0x0, 0x0)
	github.com/docker/compose-cli/local/compose/logs.go:54 +0x188
golang.org/x/sync/errgroup.(*Group).Go.func1(0xc000352cf0, 0xc0004b3860)
	golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57 +0x59
created by golang.org/x/sync/errgroup.(*Group).Go
	golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:54 +0x66
```

**Describe the results you expected:**

- To follow the logs

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
❯ docker version
Client:
 Cloud integration: 1.0.14
 Version:           20.10.6
 API version:       1.41
 Go version:        go1.16.3
 Git commit:        370c289
 Built:             Fri Apr  9 22:46:57 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.6
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       8728dd2
  Built:            Fri Apr  9 22:44:56 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```
❯ docker context show
default
```

**Output of `docker info`:**

```
❯ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
[...]

PR #1630, opened 2021-05-04
don't invoke r.Close on error as r == nil

ndeloof:
**What I did**
Prevent panic calling r.Close as r is nil

**Related issue**
https://github.com/docker/compose-cli/issues/1629


lorenrh (approved):
Makes sense 🍡

Issue #1582, opened 2021-04-22
docker compose up does not allocate static IP properly


LaendaEagle:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Description**

I have multiple applications managed in multiple projects. For some technical reason I need those applications to run on the same network with fixed IPs (as the IP is stored in some places).
For that I have a docker-compose.yml setting the network up with appropriate driver and range. Then I refer to this network from my other docker-compose.yml file where I assign the IP address to my application.
All this work fine with docker-compose up but the fixed IP is not allocated with docker compose up instead the next free IP of my named network is used.
Just to be clear I restate it: this is working with docker-compose, not with the docker compose new flavour available with Docker Desktop 3.3.1 (Mac).

**Steps to reproduce the issue:**
1. cd network
2. docker compose up -d
3. cd ../application
4. docker compose up -d
5. docker inspect <application-container-id>

**Describe the results you received:**
The IP of my application's container is 172.22.0.3

**Describe the results you expected:**
The IP of my application's container is 172.22.0.18

**Additional information you deem important (e.g. issue happens only occasionally):**
file: network/docker-compose.yml
```
version: "3.5"
services:
  web:
    image: nginx:latest
    networks:
        default:
            ipv4_address: 172.22.0.2

networks:
  default:
    name: testnetwork
    driver: bridge
    ipam:
      config:
        - subnet: 172.22.0.0/24
```

file: application/docker-compose.yml
```
version: "3.5"
services:
  application:
    image: nginx:latest
    networks:
        default:
            ipv4_address: 172.22.0.18
networks:
  default:
    name: testnetwork
```

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Cloud integration: 1.0.12
 Version:           20.10.5
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        55c4c88
 Built:             Tue Mar  2 20:13:00 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.5
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       363e9a8
  Built:            Tue Mar  2 20:15:47 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
[...]

PR #1633, opened 2021-05-04
configure service with ipAddress

ndeloof:
**What I did**
Setup container(s) with ip address as defined by service's network config

**Related issue**
close https://github.com/docker/compose-cli/issues/1582


lorenrh (approved):
🌮

PR #1642, opened 2021-05-05
Fix building compose files with absolute build context / Dockerfile

gtardif:

**What I did**
Allow for absolute build context path / Dockerfile path. These path are encountered using ddev, that generates composefiles.

**Related issue**
Fixes https://github.com/docker/dev-tooling-team/issues/313

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/awslabs/goformation/v4 v4.15.6
 	github.com/buger/goterm v1.0.0
 	github.com/cnabio/cnab-to-oci v0.3.1-beta1
-	github.com/compose-spec/compose-go v0.0.0-20210503135708-e8ee37c1478c
+	github.com/compose-spec/compose-go v0.0.0-20210505145624-6dcd3d18b38b
 	github.com/containerd/console v1.0.1
 	github.com/containerd/containerd v1.4.3
 	github.com/containerd/continuity v0.0.0-20200928162600-f2cc35102c2a // indirect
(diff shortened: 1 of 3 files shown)


local/compose/build.go
```diff
@@ -271,10 +272,14 @@ func (s *composeService) toBuildOptions(service types.ServiceConfig, imageTag st
 		plats = append(plats, p)
 	}
 
+	dockerfilePath := service.Build.Dockerfile
+	if filepath.IsAbs(service.Build.Context) && !filepath.IsAbs(dockerfilePath) {
```
ndeloof, line 276: Any idea why this isn't addressed by https://github.com/compose-spec/compose-go/blob/d07bfddf9d051cee60daf9e6f1b03671d620fb66/loader/normalize.go#L76 ?

ndeloof: I ran a quick test.
This code (which should have addressed this during loading) fails as `url.ParseRequestURI` doesn't return error when used with an absolute path!

I'm going to check how buildx distinguish a path vs a git URL

ndeloof: seems we will have to replicate https://github.com/docker/buildx/blob/ab09846df70df822fd32b9d0c715dc0605344618/bake/remote.go#L85-L93

gtardif: right, should be... I'll try to debug this (and adapt this PR depending on what I found)

ndeloof: proposed a (reasonable?) workaround as https://github.com/compose-spec/compose-go/pull/138 so we won't have to keep compose-go in sync with buildkif supported protocol (which might include some non-documented by the compose-spec)

gtardif: Updated with your fix in compose-go

gtardif: Thx !

ndeloof (approved):

Issue #1552, opened 2021-04-15
docker compose ps does not show the stopped/exited containers


arbesulo:
**Description**

`docker compose ps` only shows the running containers

**Steps to reproduce the issue:**
```yaml
#docker-compose.yml:
version: "3.7"

services:
  redis:
    image: redis
  nginx:
    image: nginx
```

1. docker compose up -d
2. docker compose ps
```bash
NAME                   SERVICE             STATUS              PORTS
compose_test_nginx_1   nginx               running             80/tcp
compose_test_redis_1   redis               running             6379/tcp
```
3. docker compose stop redis
4. docker compose ps
```bash
NAME                   SERVICE             STATUS              PORTS
compose_test_nginx_1   nginx               running             80/tcp
```

I would expect some information about the stopped container, as docker-compose does:
```bash
docker-compose ps
        Name                      Command               State    Ports
-----------------------------------------------------------------------
compose_test_nginx_1   /docker-entrypoint.sh ngin ...   Up       80/tcp
compose_test_redis_1   docker-entrypoint.sh redis ...   Exit 0
```

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Cloud integration: 1.0.12
 Version:           20.10.5
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        55c4c88
 Built:             Tue Mar  2 20:13:00 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.5
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       363e9a8
  Built:            Tue Mar  2 20:15:47 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

ndeloof:
This indeed has been implemented as `docker compose ps -a` for consistency with the `docker ps` command and flags, but make it backward incompatible with ̀`docker-compose`

PR #1649, opened 2021-05-06
compose ps list stopped containers. -a is used to include oneoff

ndeloof:
**What I did**
Fix behaviour of `compose ps -a` flag, so that `ps` always list stopped services, and `-a` is used to also include one_off containers (created by run command)

**Related issue**
close https://github.com/docker/compose-cli/issues/1552


lorenrh (approved):
yaay 🌮

Issue #1545, opened 2021-04-14
docker compose up doesn't use credentials to pull images from ghcr.io


Zenexer:
**Description**

Attempting to run `docker compose up -d` with a compose file that references a private image on ghcr.io fails with the following error:

> failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to create LLB definition: failed to authorize: failed to fetch anonymous token: unexpected status: 401 Unauthorized

`docker-compose up -d` works, and `docker compose up -d` works if I run `docker pull` for that image first.

**Steps to reproduce the issue:**

1. Create a docker-compose.yml file that references a private image on ghcr.io
2. Log into ghcr.io via `docker login ghcr.io`
3. Run `docker compose up -d`

**Describe the results you received:**

```
% docker --debug compose up -d
DEBUG serving grpc connection
DEBUG serving grpc connection
DEBUG serving grpc connection
[+] Building 0.7s (7/7)
 => [redacted1 internal] load build definition from Dockerfile 0.0s
 => => transferring dockerfile: 86B 0.0s
[+] Building 0.8s (9/9) FINISHED
 => [redacted1 internal] load build definition from Dockerfile 0.0s
 => => transferring dockerfile: 86B 0.0s
 => [elasticsearch internal] load build definition from Dockerfile 0.0s
 => => transferring dockerfile: 62B 0.0s
 => [phpmyadmin internal] load build definition from Dockerfile 0.0s
 => => transferring dockerfile: 69B 0.0s
 => [redacted1 internal] load .dockerignore 0.0s
 => => transferring context: 2B 0.0s
 => [elasticsearch internal] load .dockerignore 0.0s
 => => transferring context: 2B 0.0s
 => [phpmyadmin internal] load .dockerignore 0.0s
 => => transferring context: 2B 0.0s
 => ERROR [redacted1 internal] load metadata for ghcr.io/redacted2/redacted3:redacted4 0.1s
 => CANCELED [elasticsearch internal] load metadata for docker.io/library/elasticsearch:7.12.0 0.0s
 => CANCELED [phpmyadmin internal] load metadata for docker.io/library/phpmyadmin:5.1.0-fpm-alpine 0.0s
------
 > [redacted1 internal] load metadata for ghcr.io/redacted2/redacted3:redacted4:
------
The new 'docker compose' command is currently experimental. To provide feedback or request new features please open issues at https://github.com/docker/compose-cli
[...]

PR #1651, opened 2021-05-06
pass authprovider to buildkit session

ndeloof:
**What I did**
Aligned compose build code with https://github.com/docker/buildx/blob/master/commands/build.go

**Related issue**
https://github.com/docker/compose-cli/issues/1545


ndeloof:
@tonistiigi could you please confirm this is the correct way?
i.e. `GetDriver` already has an `Auth` parameter we pass docker config file, why do we also have to set this as `Session` option?

gtardif (approved):
LGTM, Just pending @tonistiigi confirmation

ndeloof:
go.mod is killing me ...

tonistiigi:
> i.e. GetDriver already has an Auth parameter we pass docker config file, why do we also have to set this as Session option?

Yes, both are needed. `GetDriver` is only used when pulling buildkit image itself for creating the builder container one time. That image is public but we pass auth to avoid registry side rate-limiting.

ndeloof:
dependency issue:
buildx `authprovider` import introduced in this PR relies on containerd 1.5.x
kube backend relies on helm -> deislabs/oras -> containerd 1.4.x

containerd updated `PushContent` signature to add a `platform` parameter. 
As Goland does not support polymorphism nor default parameter values (welcome into XXIth century) I can't find any possible workaround here 😱

go.mod
```diff
@@ -37,40 +39,49 @@ require (
...
+	k8s.io/client-go v0.20.6
 	rsc.io/letsencrypt v0.0.3 // indirect
 	sigs.k8s.io/kustomize/kyaml v0.10.5
 )
 
 // (for buildx)
 replace github.com/jaguilar/vt100 => github.com/tonistiigi/vt100 v0.0.0-20190402012908-ad4c4a574305
+
+// see https://github.com/helm/helm/pull/9675
+replace helm.sh/helm/v3 => github.com/ndeloof/helm/v3 v3.1.0-rc.1.0.20210507084602-d0ceda767d95
```
gtardif, line 87: A quick comment to explain this hack would be useful in the future...

gtardif (approved):
LGTM with nit (comment on go mod replace)

PR #1659, opened 2021-05-11
Add standalone execution support

ulyssessouza:
**What I did**
Detect when executed as a standalone binary and run commands independently of `docker/cli`


ndeloof:
obsolete

PR #1661, opened 2021-05-12
feat(ecs): add extension to specify Fargate capacity provider

### LINKED ISSUE #674: Allow user to select FARGATE_SPOT capacity provider
initially reported by https://github.com/docker/ecs-plugin/issues/143
see https://aws.amazon.com/fr/blogs/aws/aws-fargate-spot-now-generally-available/

We should provide some way for end-user to select capacity provider, either `FARGATE` (default) or `FARGATE_SPOT`.

tyronedd:
**What I did**
Add an `x-aws-cluster_fargate_type` extension to specify the default capacity provider for the generated cluster. Allows either `FARGATE` (default) or `FARGATE_SPOT`.

**Related issue**
Closes #674  

image


tyronedd:
@ndeloof I have ZERO experience in Go, but wanted to give it a Go (hehe) anyway. Obviously this code is far from prod/merge ready, but wanted to get your thoughts on the approach I took here.

ndeloof:
You don't need a new x-aws extension here (we already have too much of those...)
https://github.com/compose-spec/compose-spec/blob/master/deploy.md#placement can be used to declare criteria a node must match so the service get scheduled, can be used to define capacity provider affinities

Issue #1664, opened 2021-05-12
npipe binding not created when running docker compose up


luke-barnett:
**Description**

npipe bindings on windows are not created when running docker compose vs. directly running docker-compose.

**Steps to reproduce the issue:**
1. Create basic docker-compose.yaml definition using npipe binding
2. Run docker compose up
3. Inspect container to validate bindings are present

Example compose file:

```
version: "3.8"

services:

  traefik:
    image: traefik:v2.2.0-windowsservercore-1809
    command:
      - "--ping"
      - "--api.insecure=true"
      - "--providers.docker.endpoint=npipe:////./pipe/docker_engine"
      - "--providers.docker.exposedByDefault=false"
    ports:
      - "8079:8080"
    volumes:
      - type: npipe
        source: '\\.\pipe\docker_engine_windows'
        target: '\\.\pipe\docker_engine'
```

**Describe the results you received:**
Container created without npipe binding

**Describe the results you expected:**
Container created with npipe binding

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
Client:
 Cloud integration: 1.0.14
 Version:           20.10.6
 API version:       1.41
 Go version:        go1.16.3
 Git commit:        370c289
 Built:             Fri Apr  9 22:49:36 2021
 OS/Arch:           windows/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.6
  API version:      1.41 (minimum version 1.24)
  Go version:       go1.13.15
  Git commit:       8728dd2
  Built:            Fri Apr  9 22:45:40 2021
  OS/Arch:          windows/amd64
  Experimental:     false
```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```
[
    {
        "Name": "default",
        "Metadata": {
            "StackOrchestrator": "swarm"
        },
        "Endpoints": {
            "docker": {
                "Host": "npipe:////./pipe/docker_engine",
                "SkipTLSVerify": false
            }
        },
        "TLSMaterial": {},
        "Storage": {
            "MetadataPath": "\u003cIN MEMORY\u003e",
            "TLSPath": "\u003cIN MEMORY\u003e"
        }
    }
]
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
[...]

PR #1676, opened 2021-05-17
named pipe need to be mounted using the bind API

ndeloof:
**What I did**
Used the `Bind` API for named pipes mount, just like we do for bind mounts

**Related issue**
close https://github.com/docker/compose-cli/issues/1664


lorenrh (approved):
🎉

PR #1679, opened 2021-05-17
resolve build args on compose up --build

ndeloof:
**What I did**
Resolve build args from environment when running `docker compose up --build`

**Related issue**
close https://github.com/docker/compose-cli/issues/1665

### DIFF
--- a/local/compose/build.go
+++ b/local/compose/build.go
@@ -45,24 +45,21 @@ func (s *composeService) Build(ctx context.Context, project *types.Project, opti
 	opts := map[string]build.Options{}
 	imagesToBuild := []string{}
 
-	args := map[string]string{}
-	for k, v := range options.Args.Resolve(func(s string) (string, bool) {
+	args := flatten(options.Args.Resolve(func(s string) (string, bool) {
 		s, ok := project.Environment[s]
 		return s, ok
-	}).RemoveEmpty() {
-		args[k] = *v
-	}
+	}))
 
 	for _, service := range project.Services {
 		if service.Build != nil {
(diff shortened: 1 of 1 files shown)


lorenrh (approved):
🌮

gtardif (approved):

SamuelMS:
@ndeloof I did a little digging but couldn't find anything about Docker's release cycle – do you have any guidance you can share on when this'll get pushed up to Docker for Mac? I'm glad there's a workaround (use `docker-compose` instead), but a few of our developers have run into this issue now.

reagent:
Ran into this one today and upgraded Docker on Mac -- looks like this is fixed in `20.10.12` / `1.29.2`:

```
$ docker-compose --version
docker-compose version 1.29.2, build 5becea4c
$ docker --version
Docker version 20.10.12, build e91ed57
```

tucomel:
the problem still persist here 
already using latest versions and the command 
`docker-compose build --build-arg ASPNETCORE_ENVIRONMENT=Development`
or
`docker-compose build --build-arg ENV=Development`
still not works 

image

juan7732:
@tucomel That doesn't seem to be an issue with the problem mentioned here. You need to remove the whitespace after `ASPNETCORE_ENVIRONMENT` in your docker-compose file.
image
image

F8ER:
> the problem still persist here already using latest versions and the command `docker-compose build --build-arg ASPNETCORE_ENVIRONMENT=Development` or `docker-compose build --build-arg ENV=Development` still not works
> 
> image

My apologies, quite off-topic, but just to clarify, is that Vim?

PR #1691, opened 2021-05-19
validate and pre-process flags in PreRun

ndeloof:
**What I did**
Moved flags validation/processing into `PreRunE`, so that `RunE` is dedicated to execution.
Longer terms goal is to be able to directly set plain `compose.Service` method as command's `RunE`


cli/cmd/compose/run.go
```diff
@@ -109,11 +109,14 @@ func runCommand(p *projectOptions, backend compose.Service) *cobra.Command {
...
-		RunE: Adapt(func(ctx context.Context, args []string) error {
+		PreRunE: Adapt(func(ctx context.Context, args []string) error {
+			opts.Service = args[0]
 			if len(args) > 1 {
 				opts.Command = args[1:]
 			}
-			opts.Service = args[0]
+			return nil
+		}),
+		RunE: Adapt(func(ctx context.Context, args []string) error {
```
gtardif, line 119: Why keep `if len(opts.publish) > 0 && opts.servicePorts {` here and in PreRunE ?

gtardif (commented):
LGTM, appart from one validation (forgotten ?) in RunE

Issue #1685, opened 2021-05-18
restart <service> will restart all dependent services


BretFisher:
**Description**

In the old `docker-compose restart <service>` it only restarts that specific service.

In `docker compose restart <service>` it will first restart services that <service> depends on, including DB's. If healthchecks are used for `depends_on`, it'll then have to re-wait for dependent services to be ready again. 

**Steps to reproduce the issue:**

1. Create two services in a compose file where Service1 `depends_on` Service 2.
2. Restart Service1.

**Describe the results you received:**

Service2 will restart, and then Service1 will restart.

**Describe the results you expected:**

Only Service1 should be affected.

**Additional information you deem important (e.g. issue happens only occasionally):**

Using `Docker Compose version 2.0.0-beta.1` in Docker Desktop 3.3.3 (64133) on macOS

**Output of `docker version`:**

```
Client:
 Cloud integration: 1.0.14
 Version:           20.10.6
 API version:       1.41
 Go version:        go1.16.3
 Git commit:        370c289
 Built:             Fri Apr  9 22:46:57 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true
```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```
default
```

PR #1693, opened 2021-05-20
only (re)start selected services

ndeloof:
**What I did**
Only (re)start selected services when user explicitly pass service names on (re)start commands.
Otherwise consider all project services

**Related issue**
close https://github.com/docker/compose-cli/issues/1685

### DIFF
--- a/local/compose/start.go
+++ b/local/compose/start.go
@@ -20,6 +20,7 @@ import (
 	"context"
 
 	"github.com/docker/compose-cli/api/compose"
+	"github.com/docker/compose-cli/utils"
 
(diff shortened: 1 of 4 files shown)


local/compose/start.go
```diff
@@ -18,6 +18,7 @@ package compose
 
 import (
 	"context"
+	"github.com/docker/compose-cli/utils"
```
lorenrh, line 21: extra line to separate from context?

lorenrh (approved):
YESS! (well, after fixing the Lint error 🤭 )

PR #1704, opened 2021-05-24
Do not resolve mount linux abs path as windows path on windows 

gtardif:
(eg. /var/run/docker.sock => c:/var/run/docker.sock folder)

**What I did**
test win mount path with ` filepath.IsAbs()`  and also `HasPrefix("/")  for linux abs path...

**Related issue**
Fixes https://github.com/docker/compose-cli/issues/1591


ndeloof (approved):

ndeloof:
ok, so golang's standard library just does it wrong, nice to know.
image

Issue #1710, opened 2021-05-25
docker compose version is confusing


PMExtra:
For now, execute `docker compose version` will get `Docker Compose version 2.0.0-beta.1`

But the tag of this repo is `1.0.x`

ndeloof:
Definitively.
We are working on isolating "compose 2" as a separate codebase, vs compose-cli for ECS/ACI/Kube cloud integration.

PR #1714, opened 2021-05-27
split compose-cli "docker" and composeV2 cli-plugin release processes

ndeloof:
**What I did**
split the release process so `v1.x` will build compose-cli while `v2.x` will build the composev2 cli-plugin. Temporary solution until composev2 code is moved to a new home

see sample release on https://github.com/ndeloof/compose-cli/releases

**Related issue**
close https://github.com/docker/compose-cli/issues/1710
internally tracked as https://docker.atlassian.net/browse/IL-159

image


mat007 (approved):
Makes sense!

Issue #1712, opened 2021-05-26
docker compose exec blows up when config specifies an environment variable that does not exist


jakeonfire:
**Description**
`docker compose exec` does not work when the compose config specifies an environment variable without a value that does not exist.

**Steps to reproduce the issue:**
1. Start a container meeting the above conditions
2. Exec into that container (e.g. `docker compose exec app bash`)
3. 💥 

**Describe the results you received:**
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1c4f64e]

goroutine 1 [running]:
github.com/docker/compose-cli/local/compose.(*composeService).Exec(0xc00000c090, 0x20cbec0, 0xc000316040, 0xc0004888c0, 0x0, 0x0, 0x7ffeefbff75d, 0x3, 0xc000163690, 0x1, ...)
	github.com/docker/compose-cli/local/compose/exec.go:58 +0x76e
github.com/docker/compose-cli/api/compose.(*ServiceProxy).Exec(0xc00002e620, 0x20cbec0, 0xc000316040, 0xc0004888c0, 0x0, 0x0, 0x7ffeefbff75d, 0x3, 0xc000163690, 0x1, ...)
	github.com/docker/compose-cli/api/compose/proxy.go:273 +0x11f
github.com/docker/compose-cli/cli/cmd/compose.runExec(0x20cbec0, 0xc000316040, 0x20e14b8, 0xc00002e620, 0xc000321000, 0x7ffeefbff75d, 0x3, 0xc000163690, 0x1, 0x1, ...)
	github.com/docker/compose-cli/cli/cmd/compose/exec.go:113 +0x275
github.com/docker/compose-cli/cli/cmd/compose.execCommand.func2(0x20cbec0, 0xc000316040, 0xc000163680, 0x2, 0x2, 0xc00013e108, 0x12)
	github.com/docker/compose-cli/cli/cmd/compose/exec.go:62 +0xd8
github.com/docker/compose-cli/cli/cmd/compose.Adapt.func1(0xc0003cfb80, 0xc000163680, 0x2, 0x2, 0x0, 0x0)
	github.com/docker/compose-cli/cli/cmd/compose/compose.go:60 +0x12f
github.com/spf13/cobra.(*Command).execute(0xc0003cfb80, 0xc000454260, 0x2, 0x2, 0xc0003cfb80, 0xc000454260)
[...]

PR #1715, opened 2021-05-27
remove unresolved vars from env set by `exec`

ndeloof:
**What I did**
fixed exec environment computation to ignore unset env variable values

**Related issue**
close https://github.com/docker/compose-cli/issues/1712

image


mat007 (approved):
Looks good!

lorenrh (approved):

Issue #1732, opened 2021-06-02
`docker compose up -d` can't up container with image digest


ryan5500:
**Description**

`docker compose up -d` can't up container with image digest like `postgres:10-alpine@sha256:xxx...`.
But `docker-compose up -d` could up container.

**Steps to reproduce the issue:**

1. prepare docker-compose.yml like below.

```
version: '3'
services:
  postgres-10-alpine:
    image: postgres:10-alpine@sha256:7b4ccd2681f2b3615998c80eaaf0c1c1749f970757f89a651d9dacb2219943a3
    environment:
      POSTGRES_PASSWORD: secret
    volumes:
      - pgdata-10-alpine:/var/lib/postgresql/data
    ports:
      - 5432:5432

volumes:
  pgdata-10-alpine:
```

2. exec `docker compose up -d`

**Describe the results you received:**

```bash
$ docker compose up -d
[+] Building 0.1s (5/5) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                                                                   0.0s
 => => transferring dockerfile: 132B                                                                                                                                                                   0.0s
 => [internal] load .dockerignore                                                                                                                                                                      0.0s
 => => transferring context: 2B                                                                                                                                                                        0.0s
 => [internal] load metadata for docker.io/library/postgres:10-alpine@sha256:7b4ccd2681f2b3615998c80eaaf0c1c1749f970757f89a651d9dacb2219943a3                                                          0.0s
 => CACHED [1/1] FROM docker.io/library/postgres:10-alpine@sha256:7b4ccd2681f2b3615998c80eaaf0c1c1749f970757f89a651d9dacb2219943a3                                                                     0.0s
 => => resolve docker.io/library/postgres:10-alpine@sha256:7b4ccd2681f2b3615998c80eaaf0c1c1749f970757f89a651d9dacb2219943a3                                                                            0.0s
 => ERROR exporting to image                                                                                                                                                                           0.0s
 => => exporting layers                                                                                                                                                                                0.0s
 => => writing image sha256:7421baebe58d2aaf8131c1020d4a7fa7ca0869b68bfbf750c7627f9820b83f25                                                                                                           0.0s
 => => naming to docker.io/library/postgres:10-alpine@sha256:7b4ccd2681f2b3615998c80eaaf0c1c1749f970757f89a651d9dacb2219943a3                                                                          0.0s
------
 > exporting to image:
------
[...]

PR #1733, opened 2021-06-02
stop using buildkit hack to pull images

ndeloof:
**What I did**
Stop using Buildkit hack to get image pulled, first pull then build
TODO investigate how we could get those in parallel within a shared progress.writer

**Related issue**
close https://github.com/docker/compose-cli/issues/1732

### DIFF
--- a/local/compose/build.go
+++ b/local/compose/build.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/compose-spec/compose-go/types"
 	"github.com/containerd/containerd/platforms"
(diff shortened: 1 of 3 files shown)


gtardif:
This should also allow better error metrics, separating pull issues from build issues

local/compose/build.go
```diff
@@ -87,7 +86,18 @@ func (s *composeService) Build(ctx context.Context, project *types.Project, opti
 }
 
 func (s *composeService) ensureImagesExists(ctx context.Context, project *types.Project, observedState Containers, quietPull bool) error {
-	images, err := s.getImageDigests(ctx, project)
+	for _, service := range project.Services {
+		if service.Image == "" && service.Build == nil {
+			return fmt.Errorf("invalid service %q. Must specify either image or build", service.Name)
```
mat007, line 91: nit: I think go errors are not supposed to contain caps, what about
```suggestion
			return fmt.Errorf("invalid service %q: must specify either image or build", service.Name)
```
?

mat007 (approved):
Makes sense!

local/compose/build.go
ndeloof: AFAIk there's a few other places we didn't strictly applied this gopherextremism.
Considering this error was just moved by this PR, to be debated in a separate PR if we want to also apply this "rule"

PR #1735, opened 2021-06-03
Move progress writter into backend(s)

ndeloof:
**What I did**
moved `progress.Run` to the backends, to offer more flexibility
(long term plan is to get `Up` implemented by local backend)

image

### DIFF
--- a/cli/cmd/compose/kill.go
+++ b/cli/cmd/compose/kill.go
@@ -18,26 +18,21 @@ package compose
 
 import (
 	"context"
+	"os"
 
+	"github.com/compose-spec/compose-go/types"
 	"github.com/spf13/cobra"
 
 	"github.com/docker/compose-cli/api/compose"
 )
 
-type killOptions struct {
-	*projectOptions
-	Signal string
-}
-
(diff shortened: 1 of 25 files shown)


local/compose/remove.go
```diff
@@ -48,18 +51,36 @@ func (s *composeService) Remove(ctx context.Context, project *types.Project, opt
...
+	if options.Force {
+		fmt.Println(msg)
+	} else {
+		confirm, err := prompt.User{}.Confirm(msg, false)
+		if err != nil {
+			return err
+		}
+		if !confirm {
+			return nil
+		}
```
mat007, line 68: nit: could be
```suggestion
		if err != nil || !confirm {
			return err
		}
```

mat007 (approved):
Makes sense!

cli/cmd/compose/kill.go
```diff
@@ -18,26 +18,21 @@ package compose
 
 import (
 	"context"
+	"github.com/compose-spec/compose-go/types"
```
lorenrh, line 21: linting failure 😢

lorenrh (approved):

PR #1747, opened 2021-06-07
move progress to backend on "down"

ndeloof:
**What I did**
Moved progress rendering to backend
Moved check for supported options to backend

this makes `down` CLI command way simpler and independent from backend.

### DIFF
--- a/aci/compose.go
+++ b/aci/compose.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 
 	"github.com/compose-spec/compose-go/types"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	"github.com/docker/compose-cli/aci/convert"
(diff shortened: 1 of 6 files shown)


aci/compose.go
```diff
@@ -123,21 +125,29 @@ func (cs aciComposeService) warnKeepVolumeOnDown(ctx context.Context, projectNam
...
+	if options.Volumes {
+		return errors.Wrap(errdefs.ErrNotImplemented, "--volumes option is not supported on ACI")
 	}
-	if cg.IsHTTPStatus(http.StatusNoContent) {
-		return errdefs.ErrNotFound
+	if options.Images != "" {
+		return errors.Wrap(errdefs.ErrNotImplemented, "--rmi option is not supported on ACI")
 	}
+	return progress.Run(ctx, func(ctx context.Context) error {
+		logrus.Debugf("Down on projectName with name %q", projectName)
```
mat007, line 135: I think this was a previous renaming gone wrong and it should probably be
```suggestion
		logrus.Debugf("Down on project with name %q", projectName)
```
?

mat007 (approved):
Makes sense!

PR #1752, opened 2021-06-07
move up logic from CLI into local backend

ndeloof:
**What I did**
move `up` command logic on backend so that local backend do implement `compose.API#Up`

this removes reference to context type from `cli` package, so that this code is now independent from ACI/ECS/Kube backends


ulyssessouza (approved):
LGTM

PR #1767, opened 2021-06-09
automatically disable TTY if shell isn't a terminal

ndeloof:
**What I did**
- Make `compose exec` / `compose run` safer by automatically applying no-TTY option when active shell isn't a terminal
- Fix `compose exec` ran in a non-interactive terminal (see https://docker.atlassian.net/browse/IL-47)

inspired by https://github.com/docker/cli/blob/master/cli/command/container/exec.go
such a pain this code is all private and we can't just use it here.


lorenrh (approved):
LGTM 🎉

PR #1768, opened 2021-06-09
include packaging LICENSE in release

ndeloof:
**What I did**
release process to include binary distribution LICENSE


mat007 (approved):
Makes sense!

PR #1773, opened 2021-06-10
Completion

ulyssessouza:
**What I did**
Add dynamic completion to commands

**Related issue**
https://github.com/docker/compose-cli/issues/1338


cli/cmd/compose/completion.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+func completionCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "completion [bash|zsh|fish|powershell]",
```
thaJeztah, line 27: Wondering if, instead, we should have a `--completion` flag, to prevent the completion command itself being part of user's completions.


I think the _ideal_ would be for the `--completion` flag to be on the `docker` command, and (if plugins are installed) that flag to collect completions for the _cli_, _including_ the plugins that are installed).

See discussion on https://github.com/docker/cli/issues/1257 (there may be some other discussions about this)

thaJeztah: Basically, this command would not be something that users are using on a daily base (it would mostly be, either a "set and forget" option, or something that packagers (including Docker Desktop) could use to export and install the scripts

ndeloof: I fully agree this would make more sense for completion script to be generated fro docker/cli and installed plugins

thaJeztah: The `docker-cli-plugin-metadata` command currently implements "v0.0.1" of the cli-plugin spec. Perhaps we should do a `v0.0.1` of that spec that (when implemented by a plugin) means that the plugin has a `--completion=<bash|fish|auto|whatever>` flag, or a (hidden?) subcommand (`docker-cli-plugin-completion`). In that case, the CLI could iterate through all plugins that implement v0.0.2 and collect / concatenate the completion scripts.

I'm not 100% sure if concatenating works of course (perhaps?) something we could try.

thaJeztah: That said; "conceptually" I think this feature is nice to have (and something I wanted to implement in the `docker` CLI as well to allow more easily package everything), just that it "should be there, but not _too much_ in plain sight" as it's not a command users want to be using as part of their daily work (and to reserve the possibility to remove/rename/change)

ndeloof: I don't see any reason we would like to hide a `completion` command. That owuld just follow the standard cobra completion pattern https://github.com/spf13/cobra/blob/master/shell_completions.md

We can make this a hidden command if you don't want users to be disturbed by yet another verb on docker CLI, but really that sounds a minor issue to me vs the benefits

thaJeztah: I definitely think it clutters the UX. w.r.t. Cobra "conventions", not sure we need to follow those (i.e. have a framework dictate what our UX will look like); the cobra convention also would have this as a "top level" subcommand (in our case `docker completion`), not as a command on a subcommand.

As mention, this is _not_ a command that users would be typing (interactively) as part of their workflow; either it would be in (e.g.) `~/.profile` or `~/.bashrc` (or .. some other hidden file), or it would be used by packagers to automatically install.

In addition, if we _do_ add support for this to the main `docker` command, it means that this subcommand (or flag) would be a temporary / transitional thing, and I'd rather not have it  clutter the UX in the meantime (at least not as a command, more preferably as a `--flag`).

ndeloof: I would not consider `docker completion` to be a command to be hidden, even it's not designed to be ran on a daily basis (or we should remove most of the existing commands :P). As long as I'll have to run it any time I install / update a CLI plugin, it would still make sense to me this is a standard command on the CLI. I also hate having flags being used as psuedo-commands, as this just negates the concept to "tweak a command". Maybe `docker system completion` would make sense then, if you don't want this top-level.

Anyway I don't really care how it get exposed, but mostly interested we could get a completion script generated for CLI+plugins

cmd/compose/completion.go
```diff
@@ -0,0 +1,113 @@
...
+  # you will need to enable it.  You can execute the following once:
+
+  $ echo "autoload -U compinit; compinit" >> ~/.zshrc
+
+  # To load completions for each session, execute once:
+  $ docker compose completion zsh > "${fpath[1]}/_docker_compose"
+
+  # You will need to start a new shell for this setup to take effect.
+
+fish:
```
mat007, line 54:
```suggestion
Fish:
```
?

cmd/compose/completion.go
```diff
@@ -0,0 +1,113 @@
...
+			case "zsh":
+				err = cmd.Root().GenZshCompletion(os.Stdout)
+			case "fish":
+				err = cmd.Root().GenFishCompletion(os.Stdout, true)
+			case "powershell":
+				err = cmd.Root().GenPowerShellCompletion(os.Stdout)
+			}
+			return err
+		},
+		Hidden: true,
```
mat007, line 86: Discussing with @ulyssessouza it seems this `completion` command is only needed when using the plugin standalone.
We’re thinking we may be able to remove this, as users will not use the plugin without going through the docker CLI.

ulyssessouza: Done

lorenrh (approved):
🌮 🎉

PR #1779, opened 2021-06-11
define compose labels within the compose API

ndeloof:
**What I did**
Avoid duplicate definition of Tags and Labels
compose API now defines all compose v2 labels as public constants in the API
This prepares moving compose v2 into a new home with public labels we can use from compose-cli


mat007 (approved):
Makes sense!

PR #1782, opened 2021-06-11
fix volume inheritance and conflict with trailing '/'

ndeloof:
**What I did**

Avoids conflict when volumes are declared on same path but some use a trailing `/` by normalizing mount target using `path.Clean`

Also fix inherited volume computation (`ImageInspect.ContainerConfig` is not set with Dockerfile VOLUME, we have to use `ImageInspect.Config`)

**Related issue**
close https://github.com/docker/compose-cli/issues/1775

### DIFF
--- a/local/compose/create.go
+++ b/local/compose/create.go
@@ -19,6 +19,7 @@ package compose
 import (
 	"context"
 	"fmt"
+	"path"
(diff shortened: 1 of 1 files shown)


ulyssessouza (approved):
LGTM

Issue #1786, opened 2021-06-14
Introduce `docker compose logs --since`


ndeloof:
`docker logs` defines `--since` option to limit the amount of logs to retrieve from container's history. This would make sense to align and define a comparable option for `docker compose logs`

see https://github.com/moby/moby/blob/master/api/types/client.go#L75

afshinpaydar-binary:
@ndeloof A simple implementation by passing `--since` & `--until` parameters to docker engine https://github.com/docker/compose-cli/pull/1791

PR #1791, opened 2021-06-14
Add docker compose logs --since option

afshinpaydar-binary:
**What I did**

Add docker compose logs --since option

**Related issue**
[#1786](https://github.com/docker/compose-cli/issues/1786)


local/compose/logs.go
```diff
@@ -48,6 +48,7 @@ func (s *composeService) Logs(ctx context.Context, projectName string, consumer
 				ShowStdout: true,
 				ShowStderr: true,
 				Follow:     options.Follow,
+				Since:      options.Since,
```
ndeloof, line 51: docker API `Since` is documented as a Unix Timestamp (https://docs.docker.com/engine/api/v1.41/#operation/ContainerLogs)
but according to [docker/ci code](https://github.com/docker/cli/blob/master/cli/command/container/logs.go#L61) just passing to the engine seems to be ok.
Confirmed by actual engine code https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/api/types/time/timestamp.go#L20-L26
That's a bit weird, but ¯\\_(ツ)_/¯

ndeloof (approved):
LGTM, maybe would make sense to also implement `--until` on the same basis for consistency?

Issue #1802, opened 2021-06-16
interoperability issue


ndeloof:
docker-compose up fails when same compose file was first started by docker compose up

```
Building with native build. Learn about native build in Compose here: https://docs.docker.com/go/compose-native-build/
Traceback (most recent call last):
  File "docker-compose", line 3, in <module>
  File "compose/cli/main.py", line 80, in main
  File "compose/cli/main.py", line 192, in perform_command
  File "compose/metrics/decorator.py", line 18, in wrapper
  File "compose/cli/main.py", line 1165, in up
  File "compose/cli/main.py", line 1161, in up
  File "compose/project.py", line 675, in up
  File "compose/project.py", line 752, in _get_convergence_plans
  File "compose/service.py", line 408, in convergence_plan
  File "compose/service.py", line 434, in _containers_have_diverged
  File "compose/container.py", line 307, in has_legacy_proj_name
  File "distutils/version.py", line 52, in __lt__
  File "distutils/version.py", line 337, in _cmp
TypeError: '<' not supported between instances of 'str' and 'int'
[1494986] Failed to execute script docker-compose
```

ndeloof:
docker-compose expect compose version to be `x.y.z` while docker compose use internal.Version
https://github.com/docker/compose/blob/318181b21e65b4c711ae83b937a78c660541323f/compose/container.py#L298

PR #1803, opened 2021-06-16
better interoperability with docker-compose on version label

ndeloof:
**What I did**
Set compose.version label as `<int>.<int>.<int>` for interoperability with docker-compose (python distutils.LooseVersion)

**Related issue**
close https://github.com/docker/compose-cli/issues/1802

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,6 +37,7 @@ require (
 	github.com/google/go-cmp v0.5.5
 	github.com/hashicorp/go-multierror v1.1.0
 	github.com/hashicorp/go-uuid v1.0.2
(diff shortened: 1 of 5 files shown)


lorenrh (approved):
🎉

mat007 (approved):
Makes sense!

mat007 (commented):
Any test that already exists by the way? 😁

ndeloof:
@mat007 interesting suggestion. We discussed running docker-compose test suite with `docker compose` but it's mostly based on implementation details, not features. Sort of "compatibility test suite" would be nice, we tried to define one by compose-spec, but so far limited to a few generic usages.

PR #1821, opened 2021-06-21
Don't apply restart_policy on `compose run`

ndeloof:
Compose run is used to execute a "one off" container, restart_policy doesn't make sense in this context (should only apply to long running services)
This also fix incompatibility with auto-remove HostConfig option

Also fix earlier one-off containers being started by `compose up`

**Related issue**
close https://github.com/docker/compose-cli/issues/1694

image


lorenrh (approved):
🎉

PR #1829, opened 2021-06-22
compose ps to include container command

ndeloof:
**What I did**
Added COMMAND column to `compose ps` to align with `docker ps`

**Related issue**


lorenrh (approved):
Looks good!

ulyssessouza (approved):
LGTM

Issue #1253, opened 2021-02-05
Verbose not available via "docker compose" (and why's it different from -l debug?)


kengruven:
**Description**

With the old(?) `docker-compose`, there are two ways to get lots of debugging output: `--log-level=debug` and `--verbose`.  It's not clear what the difference between them is.  I was trying to debug some "bind mounts", for example, and those appear in `--verbose` but not in `--log-level=debug` (the most verbose logging level?).

With the new(?) `docker compose`, `--log-level debug` was fixed (#1226), but there is no `--verbose` flag.

The verbose/debug discrepancy is pretty confusing.  Why are there two ways to turn up the logging verbosity, with different feature sets?

Anyway, what I want is a way to see what "bind mounts" Docker Compose thinks there are, when running `docker compose`.  One way might be to bring `--verbose` to `docker compose`, but it'd be even nicer if `--verbose` just went away and `--log-level=debug` took over those duties, too.

**Steps to reproduce the issue:**
1. Make a docker-compose.yml file with a bind mount, like:

```
    my_service:
      volumes:
       - ./foo:/foo/
```

2. Run `docker-compose --verbose up`, and observe that it prints lines like:

    compose.cli.verbose_proxy.proxy_callable: docker create_container <- ...[redacted]... 'Binds': ['/my/project/folder/foo:/foo:rw'], ...[redacted]...

3. Try other ways to run Docker Compose verbosely, like `docker-compose --log-level=debug up`, `docker compose --verbose up`, etc.  (Note that you need to run "down" between these.)

**Describe the results you received:**

The only way to see info about bind mounts is `docker-compose --verbose`.  `--log-level=debug` doesn't print that info.  The `--verbose` flag doesn't exist for `docker compose`.

**Describe the results you expected:**

Running `--log-level=debug` should give me the maximum verbosity in all logging.  A separate `--verbose` flag is just confusing, and I don't know why it exists.

The new `docker compose` command should support the maximum level of logging, somehow.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Cloud integration: 1.0.7
 Version:           20.10.2
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        2291f61
 Built:             Mon Dec 28 16:12:42 2020
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.2
[...]

PR #1830, opened 2021-06-22
add --verbose option for compatibility

ndeloof:
**What I did**
Introduced `--verbose` command line option to enforce command line compatibility with `docker-compose`.
This just set logging to TRACE level
full compatibility would require to tweak the Docker API client so it dump API calls on console, but AFAIK this is not feasible within a CLI plugin as this would have to apply during client setup.

**Related issue**
close https://github.com/docker/compose-cli/issues/1253

### DIFF
--- a/cmd/compose/compose.go
+++ b/cmd/compose/compose.go
@@ -25,6 +25,8 @@ import (
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


mat007 (approved):
Makes sense!

Issue #1815, opened 2021-06-18
v1 incompatibility: anonymous volumes are lost when image changes


jamshid:
**Description**

Anonymous volumes are lost (!) when `up`ing a service and the container has to be recreated (e.g. because the image changes). This one seems like a serious incompatibility from v1, have to downgrade until this is fixed.

**Steps to reproduce the issue:**
1. `docker-compose up -d`
```
services:
  test:
    image: "centos:${CENTOS_TAG-7}"
    command: sleep 900
    volumes:
      - /var/testvolume
```
2. Write a file into the volume.
```
% docker-compose exec test bash
[root@cd3b0ec831eb /]# touch /var/testvolume/thisshouldremain
```
3. Change the image and `up` again, the file is gone!
```
% CENTOS_TAG=8 docker-compose up -d
% docker-compose exec test bash    
[root@77935ec85770 /]# ls /var/testvolume/
[root@77935ec85770 /]# 
```

**Describe the results you received:**

anonymous volume not preserved when container recreated

**Describe the results you expected:**

it should be preserved like it was with docker-compose v1

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  compose: Docker Compose (Docker Inc., 2.0.0-beta.3)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 58
  Running: 51
  Paused: 0
  Stopped: 7
 Images: 181
 Server Version: 20.10.7
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: d71fcd7d8303cbf684402823e425e9dd2e99285d
 runc version: b9ee9c6314599f1b4a7f497e1f1f856fe433d3b7
 init version: de40ad0
 Security Options:
  seccomp
[...]

PR #1831, opened 2021-06-22
inherit anoymous volumes

ndeloof:
**What I did**
Inherit anonymous volume from previous container
includes test case

**Related issue**
close https://github.com/docker/compose-cli/issues/1815


mat007 (approved):
Makes sense!

Issue #1151, opened 2021-01-12
Control-C issues duplicate signals to container


ryanuber:
- [x] I have tried with the latest version of Docker Desktop
  - [x] I have tried disabling enabled experimental features
  - [x] I have uploaded Diagnostics
  - Diagnostics ID: BD253A61-8FD5-49F8-BC42-41EFE6A5AC2A/20210112212415

### Expected behavior

Running a container in the foreground via `docker run` and pressing ctrl-c sends exactly one signal (defined by STOPSIGNAL in the Dockerfile) to the running process.

### Actual behavior

Two signals are sent to the running container, causing undesired behavior in the running application. Running with `-it` (interactive, TTY) produces a single signal, as expected.

### Information
  - Is it reproducible? Yes
  - Is the problem new? Yes - this seems to have been a behavior change at Docker Desktop 2.4.0. Prior versions send a single signal as expected.
  - Did the problem appear with an update? Versions from 2.4.0 onward exhibit the unexpected multiple signals behavior.
  - macOS Version: 10.15.7

### Steps to reproduce the behavior

1. Create a `Dockerfile` with the following content:
```Dockerfile
FROM ubuntu:20.04

COPY sig.sh /sig.sh
RUN chmod +x /sig.sh

STOPSIGNAL SIGINT

ENTRYPOINT ["/sig.sh"]
``` 
2. Create a script file `sig.sh` with the following content:
```bash
#!/bin/bash

trap "register" INT

register() {
    echo "INT"
    [ -z "$SIG" ] && SIG=1 || exit
}

while true; do
    sleep 100 &
    wait
done
```
3. Build the container with `docker build -t signal-test:latest .`
4. Run the container with `docker run signal-test:latest`
5. Press ctrl+c once. Note that `INT` is printed twice, thus indicating that two signals were sent to the process.

stephen-turner:
I'm just going to move the ticket to that repo.

ndeloof:
Internaly tracked as https://docker.atlassian.net/browse/IL-268

jakeonfire:
i'm not seeing _any_ signals hit the running process.

running:
```
docker compose run --rm app my_script.sh
```

`my_script.sh`:
```
#!/usr/bin/env bash

echo sleeping
sleep 30 &
PID=$!

function cleaupQUIT() {
  echo got SIGQUIT
}
trap cleanupQUIT SIGQUIT

function cleaupTERM() {
  echo got SIGTERM
}
[...]

PR #1833, opened 2021-06-22
prevent signals to be sent twice to child process

ndeloof:
**What I did**
fix signal forwarding to child `com.docker.cli` process
child process is ran in a distinct process group so it won't receive the same signal as parent process from a Ctrl+C in terminal
also filter out some signal as copy/paste from docker/cli [ForwardAllSignals](https://github.com/docker/cli/blob/74a6e2cf627be262d1ba2d55e0f66856038fddd8/cli/command/container/signals.go#L16)

**Related issue**
close https://github.com/docker/compose-cli/issues/1151

### DIFF
--- a/cli/mobycli/exec.go
+++ b/cli/mobycli/exec.go
@@ -21,16 +21,19 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"os/signal"
+	gosignal "os/signal"
 	"path/filepath"
 	"regexp"
+	"syscall"
+
+	"github.com/docker/docker/pkg/signal"
+
+	"github.com/spf13/cobra"
 
 	apicontext "github.com/docker/compose-cli/api/context"
 	"github.com/docker/compose-cli/api/context/store"
 	"github.com/docker/compose-cli/cli/metrics"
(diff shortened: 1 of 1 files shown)


thaJeztah:
Do you think we should move the `ForwardAllSignals` utility to a different location inside docker/cli so that it can be consumed here?

ndeloof:
We already have a list of func we'd like to reuse from docker/cli to avoid such duplications

thaJeztah:
> We already have a list of func we'd like to reuse from docker/cli to avoid such duplications

👍 happy to discuss; we should probably look what's the most suitable location (docker/cli, or if it's more related to the API client, perhaps in moby/moby/client (e.g.))

cli/mobycli/exec.go
```diff
@@ -117,6 +132,7 @@ func RunDocker(childExit chan bool, args ...string) error {
 				if isRuntimeSig(sig) {
 					continue
 				}
+				fmt.Println("forward signal to process " + sig.String())
```
thaJeztah, line 135: Was this for debugging? Do we want this to print output?

ndeloof:
Closing, doesn't work :'(

PR #1834, opened 2021-06-22
signal are forwarded to child processes as same process group

### LINKED ISSUE #1151: Control-C issues duplicate signals to container

ndeloof:
**What I did**
fix duplicated signal being forwarded to child com.docker.cli process
child process is ran in same process group so it receives the same signal as parent process from a Ctrl+C in terminal

**Related issue**
close #1151

afaik this also fix https://github.com/docker/for-mac/issues/5501


ndeloof:
doesn't work on windows :'(

PR #1839, opened 2021-06-24
automation to remove pending label on new comment

ndeloof:
Adds automation to remove issues `pending` label as a new comment is created.
This allows to use this label for issues where we need reporter to provide more details, and exclude them from triage using `-label:pending` as issue search filter.
As user create a new label providing more details, `pending` label will be removed, so this issue will come back under our radar

### DIFF
--- a/.github/workflows/pending_answered.yml
+++ b/.github/workflows/pending_answered.yml
@@ -1,0 +1,17 @@
+name: Continuous integration
+
+on:
+  issue_comment:
+    types: [ created ]
+
+jobs:
+  pending:
+    if: ${{ !github.event.issue.pull_request && contains(github.event.issue.labels.*.name, 'pending') }}
+    name: Remove pending label
+    runs-on: ubuntu-latest
+    steps:
+      - name: removelabel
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -88,3 +88,5 @@ replace github.com/jaguilar/vt100 => github.com/tonistiigi/vt100 v0.0.0-20190402
 // with incompatible signature changes. This fork adopt oras-project/oras-go which is up-to-date on containerd 1.5
 // see https://github.com/helm/helm/pull/9675
 replace helm.sh/helm/v3 => github.com/ndeloof/helm/v3 v3.1.0-rc.1.0.20210507084602-d0ceda767d95
+
+replace github.com/compose-spec/compose-go => github.com/ndeloof/compose-go v0.0.0-20210623084147-f0ea5c285885
```
mat007, line 92: Is this intentional in this PR?

ndeloof: nope, my bad

ndeloof: TooMuchParallelTasksException :P

lorenrh (approved):
🎖️

PR #1847, opened 2021-06-25
Move compose e2e tests into pkg

ndeloof:
**What I did**
moved compose e2e tests into `pkg` so all the compose.v2 code is now isolated from compose-cli


pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+			panic(err)
+		}
+		// We don't need a functional scan plugin, but a valid plugin binary
+		err = CopyFile(composePlugin, filepath.Join(d, "cli-plugins", scanPluginFile))
+		if err != nil {
+			panic(err)
+		}
+	}
+
+	return &E2eCLI{binDir, d, t}
```
mat007, line 103: Are we ignoring `err` if it’s different from «is not exist»?

pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+func dirContents(dir string) []string {
+	res := []string{}
```
mat007, line 107: nit:
```suggestion
	var res []string
```

pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+			t.Errorf("Config: %s\n", string(conf))
+			t.Error("Contents of config dir:")
+			for _, p := range dirContents(d) {
+				t.Errorf(p)
+			}
+		}
+		_ = os.RemoveAll(d)
+	})
+
+	_ = os.MkdirAll(filepath.Join(d, "cli-plugins"), 0755)
```
mat007, line 80: Are we ignoring the error?

pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+	_ = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		res = append(res, filepath.Join(dir, path))
+		return nil
+	})
+	return res
+}
+
+func findExecutable(executableName string, paths []string) (string, error) {
+	for _, p := range paths {
+		bin, err := filepath.Abs(path.Join(p, executableName))
```
mat007, line 117: It should probably use `filepath.Join`?
```suggestion
		bin, err := filepath.Abs(filepath.Join(p, executableName))
```

pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+
+		if _, err := os.Stat(bin); os.IsNotExist(err) {
+			continue
+		}
+
+		return bin, nil
+	}
+
+	return "", errors.Wrap(os.ErrNotExist, "executable not found")
```
mat007, line 129: nit: adding context text to wrapped errors are suggested to be of the form
```suggestion
	return "", errors.Wrap(os.ErrNotExist, "looking for executable")
```
as there is usually a good chance the error already including something around «not found» and this would just add some repetition

pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// CopyFile copies a file from a sourceFile to a destinationFile setting permissions to 0755
+func CopyFile(sourceFile string, destinationFile string) error {
```
mat007, line 133: nit:
```suggestion
func CopyFile(sourceFile, destinationFile string) error {
```

pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+	src, err := os.Open(sourceFile)
+	if err != nil {
+		return err
+	}
+	// nolint: errcheck
+	defer src.Close()
+
+	dst, err := os.OpenFile(destinationFile, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0755)
```
mat007, line 141: Can’t this just be
```suggestion
	dst, err := os.Create(destinationFile)
```
?

pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// WaitForCondition wait for predicate to execute to true
+func (c *E2eCLI) WaitForCondition(predicate func() (bool, string), timeout time.Duration, delay time.Duration) {
+	checkStopped := func(logt poll.LogT) poll.Result {
+		pass, description := predicate()
+		if !pass {
+			return poll.Continue("Condition not met: %q", description)
+		}
+		return poll.Success()
```
mat007, line 228: We could invert the condition (arguably it’s usually easier to read)
```suggestion
		if pass {
			return poll.Success()
		}
		return poll.Continue("Condition not met: %q", description)
```

pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+
+// WaitForCmdResult try to execute a cmd until resulting output matches given predicate
+func (c *E2eCLI) WaitForCmdResult(command icmd.Cmd, predicate func(*icmd.Result) bool, timeout time.Duration, delay time.Duration) {
+	assert.Assert(c.test, timeout.Nanoseconds() > delay.Nanoseconds(), "timeout must be greater than delay")
+	var res *icmd.Result
+	checkStopped := func(logt poll.LogT) poll.Result {
+		fmt.Printf("\t[%s] %s\n", c.test.Name(), strings.Join(command.Command, " "))
+		res = icmd.RunCmd(command)
+		if !predicate(res) {
+			return poll.Continue("Cmd output did not match requirement: %q", res.Combined())
```
mat007, line 214: Is the string used to create an error? If yes maybe it should be all lower case?

pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+			return poll.Continue("reaching %q: Error %s", endpoint, err.Error())
+		}
+		if r.StatusCode == expectedStatus {
+			return poll.Success()
+		}
+		return poll.Continue("reaching %q: %d != %d", endpoint, r.StatusCode, expectedStatus)
+	}
+	poll.WaitOn(t, checkUp, poll.WithDelay(retryDelay), poll.WithTimeout(timeout))
+	if r != nil {
+		b, err := ioutil.ReadAll(r.Body)
```
mat007, line 262: There is probably a `defer r.Body.Close()` missing?
It’s supposed to be done as soon as the error returned by `client.Get` is nil.

pkg/e2e/scan_message_test.go
```diff
@@ -0,0 +1,78 @@
...
+		assert.Assert(t, !strings.Contains(res.Combined(), "docker scan"), res.Combined())
+		res = c.RunDockerCmd("rmi", "-f", "scan-msg-test-quiet_nginx")
+		assert.Assert(t, !strings.Contains(res.Combined(), "No such image"))
+
+		res = c.RunDockerCmd("compose", "-f", "fixtures/simple-build-test/compose.yaml", "-p", "scan-msg-test-q", "build", "-q")
+		defer c.RunDockerOrExitError("rmi", "-f", "scan-msg-test-q_nginx")
+		assert.Assert(t, !strings.Contains(res.Combined(), "docker scan"), res.Combined())
+	})
+
+	_ = c.RunDockerOrExitError("rmi", "scan-msg-test_nginx")
```
mat007, line 55: Are we ignoring the error?

mat007: Why is that one outside a `t.Run`?

pkg/e2e/scan_message_test.go
```diff
@@ -0,0 +1,78 @@
...
+	})
+
+	t.Run("do not display on compose up if no image built", func(t *testing.T) { // re-run the same Compose aproject
+		res := c.RunDockerCmd("compose", "-f", "fixtures/simple-build-test/compose.yaml", "-p", "scan-msg-test", "up", "-d")
+		defer c.RunDockerOrExitError("compose", "-f", "fixtures/simple-build-test/compose.yaml", "-p", "scan-msg-test", "down", "--rmi", "all")
+		assert.Assert(t, !strings.Contains(res.Combined(), "docker scan"), res.Combined())
+	})
+
+	t.Run("do not display if scan already invoked", func(t *testing.T) {
+		_ = os.MkdirAll(filepath.Join(c.ConfigDir, "scan"), 0755)
```
mat007, line 70: Any reason for ignoring the error here?

pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+		composePluginFile += ".exe"
+		scanPluginFile += ".exe"
+	}
+	composePlugin, err := findExecutable(composePluginFile, []string{"../../bin", "../../../bin"})
+	if os.IsNotExist(err) {
+		fmt.Println("WARNING: docker-compose cli-plugin not found")
+	}
+	if err == nil {
+		err = CopyFile(composePlugin, filepath.Join(d, "cli-plugins", composePluginFile))
+		if err != nil {
```
mat007, line 93: nit:
```suggestion
		if err := CopyFile(composePlugin, filepath.Join(d, "cli-plugins", composePluginFile)); err != nil {
```

pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+var (
+	// DockerExecutableName is the OS dependent Docker CLI binary name
+	DockerExecutableName = "docker"
+)
+
+func init() {
+	if runtime.GOOS == "windows" {
+		DockerExecutableName = DockerExecutableName + ".exe"
+	}
+}
```
mat007, line 48: Maybe:
```suggestion
	DockerExecutableName = "docker" + binExt()
)

func binExt() string {
	if runtime.GOOS == "windows" {
		return ".exe"
	}
	return ""
}
```
It can be re-used further down.

pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+	})
+
+	_ = os.MkdirAll(filepath.Join(d, "cli-plugins"), 0755)
+	composePluginFile := "docker-compose"
+	scanPluginFile := "docker-scan"
+	if runtime.GOOS == "windows" {
+		composePluginFile += ".exe"
+		scanPluginFile += ".exe"
+	}
+	composePlugin, err := findExecutable(composePluginFile, []string{"../../bin", "../../../bin"})
```
mat007, line 87: nit: could be
```suggestion
	composePlugin, err := findExecutable(composePluginFile, "../../bin", "../../../bin")
```
with
```
 func findExecutable(executableName string, paths ...string) (string, error) {
```

pkg/e2e/framework.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+	}
+
+	return &E2eCLI{binDir, d, t}
+}
+
+func dirContents(dir string) []string {
+	res := []string{}
+	_ = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		res = append(res, filepath.Join(dir, path))
```
mat007, line 109: Why are we ignoring errors?

ndeloof:
@mat007 thanks for the review, but the intent here is not to review the existing tests, but to get them moved into compose.v2 `pkg`, isolated from compose-cli
We can create a "test cleanup" task if you think this makes sense

lorenrh (approved):
I'm in on creating a follow-up task for test cleanup 😅

PR #1859, opened 2021-06-29
Adds s390x support

vibhutisawant:
Signed-off-by: vibhutisawant <Vibhuti.Sawant@ibm.com>

**What I did**
Added support for s390x.
Tested the binary on s390x arch and binary works as expected. Also, all tests have passed successfully.

**Related issue**
This PR is related to https://github.com/docker/compose/issues/7785


ndeloof (approved):

ndeloof:
Makes sense, as we also do for other CLI plugins like https://github.com/docker/buildx/releases/tag/v0.5.1

Just a note that we don't have CI infrastructure for s390x support, so this will be maintained on a "best-effort" basis

vibhutisawant:
Hi @ndeloof . Thanks for adding s390x support.
Could you please provide more info on the CI infrastructure and whether it will be possible to add s390x? We can provide s390x VM if needed.

ndeloof:
We use Github Actions to run CI builds. Would make sense we can run builds on s390x, but as we plan to split this repo and move Compose.v2 to a new home, maybe better to look into this improvement in a few.

PR #1860, opened 2021-06-29
add support for detach keys on compose run|exec

ndeloof:
**What I did**
introduced support for detach keys on `compose run` and `compose exec`

**Related issue**
close https://github.com/docker/compose-cli/issues/1709

### DIFF
--- a/pkg/compose/attach.go
+++ b/pkg/compose/attach.go
@@ -27,6 +27,7 @@ import (
 	"github.com/docker/compose-cli/pkg/api"
 	moby "github.com/docker/docker/api/types"
 	"github.com/docker/docker/pkg/stdcopy"
+	"github.com/moby/term"
 
 	"github.com/docker/compose-cli/pkg/utils"
 )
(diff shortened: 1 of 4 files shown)


pkg/compose/attach.go
```diff
@@ -105,12 +106,18 @@ func (s *composeService) attachContainerStreams(ctx context.Context, container s
...
+			return restore, detached, err
+		}
+		restore = func() {
+			term.RestoreTerminal(in.FD(), state)
 		}
 		go func() {
-			io.Copy(stdin, in) //nolint:errcheck
+			_, err := io.Copy(stdin, r) //nolint:errcheck
+			if _, ok := err.(term.EscapeError); ok {
+				detached <- true
```
mat007, line 119: Can it be detached several times? Otherwise it could be
```suggestion
				close(detached)
```
?

mat007 (commented):
Any tests? 😁

pkg/compose/attach.go
ndeloof: I expect we can throw away all this code to just adopt docker/cli run/exec commands

ndeloof:
Writting a test case for this feature which requires a terminal is pretty hard without some major refactoring

mat007 (approved):
Code makes sense!

PR #1877, opened 2021-07-02
Issue template should ask for docker-compose version

rfay:
**What I did**

Just added docker-compose version to issue template.


mat007 (approved):
Makes sense, thanks!

Issue #1873, opened 2021-07-01
`docker-compose exec` stderr output goes to stdout


rfay:
**Description**

Use this docker-compose.yaml:

```yaml
services:
  web:
    container_name: web
    image: busybox
    user: "$UID:$GID"
    command: "tail -f /dev/null"
```

Now 

```bash
docker-compose exec -T web ls junker99
```

You'll see the expected `ls: junker99: No such file or directory` but unfortunately it's on stdout, not stderr, it should be on stderr (where the `ls` put it)

```bash
docker-compose exec -T web ls junker99 >/dev/null
````

shows no output, it should show `ls: junker99: No such file or directory`

docker-compose v1 does this correctly

PR #1878, opened 2021-07-02
distinguish stdout and stderr

ndeloof:
**What I did**
Distinguish stdout and stderr when running exec in non-TTY mode

This code would better re-use https://github.com/docker/cli/blob/master/cli/command/container/exec.go#L74, need to make this a public func.

**Related issue**
closes https://github.com/docker/compose-cli/issues/1873


pkg/utils/writer.go
```diff
@@ -21,7 +21,7 @@ import (
 	"io"
 )
 
-// GetWriter creates a io.Writer that will actually split by line and format by LogConsumer
+// GetWriter creates a io.Stdout that will actually split by line and format by LogConsumer
```
mat007, line 24: This looks suspicious as well.

pkg/utils/writer.go
```diff
@@ -34,7 +34,7 @@ type splitWriter struct {
 	consumer func(string)
 }
 
-// Write implements io.Writer. joins all input, splits on the separator and yields each chunk
+// Write implements io.Stdout. joins all input, splits on the separator and yields each chunk
```
mat007, line 37: This too.

mat007 (approved):
Looks good but for the changes in the comments (a bit too aggressive search and replace? 😁)

lorenrh (approved):

cli/server/proxy/streams/logs.go
```diff
@@ -24,7 +24,7 @@ import (
 	containersv1 "github.com/docker/compose-cli/cli/server/protos/containers/v1"
 )
 
-// Log implements an io.Writer that proxies logs over a gRPC stream
+// Log implements an io.Stdout that proxies logs over a gRPC stream
```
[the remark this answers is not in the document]
ndeloof: nope. Golang refactoring trying to be clever :P

PR #1879, opened 2021-07-02
signals

ndeloof:
**What I did**
removed home-made signal forwarding. Signals are sent from console to the whole process group, so child processes also receive SIGINT.

note: tested on my Mac, Linux and Windows boxes.
works when used to run compose-cli => com.docker.cli => signal_test container as described by #1151
but doesn't work if a compose service is defined to use signal_test, and use `compose run`: Ctrl+C on console cancels the context and we can't see `INT` logged on console.

so this is at least a partial fix, will require further investigations

**Related issue**
expected to fix https://github.com/docker/compose-cli/issues/1151 🤞 


mat007:
I don’t see the initial issue on Windows with beta 4. Was it only a problem on Linux?

ndeloof:
Haven't tested initial issue on Windows, but have reproduced on Linux and Mac.

cli/mobycli/exec.go
```diff
@@ -63,10 +62,8 @@ func mustDelegateToMoby(ctxType string) bool {
 }
 
 // Exec delegates to com.docker.cli if on moby context
-func Exec(root *cobra.Command) {
-	childExit := make(chan bool)
-	err := RunDocker(childExit, os.Args[1:]...)
-	childExit <- true
+func Exec() {
```
ulyssessouza, line 65: If I understood correctly the signature change is just for code cleaning, right?

ndeloof: yes, parameter is now unused (lint)

lorenrh (approved):

mat007:
@ndeloof it seems ` /generate-artifacts `  doesn’t work here?
Maybe because the PR had been created before it’s introduced?
Or was it deactivated because we’re in the process of moving to ` docker/compose ` ?

ndeloof:
generate-artifacts is for compose.v2 only, we haven't define a comparable workflow for compose-cli

mat007 (approved):
Looks good!
Tested on both Mac and Windows (using the CLI from https://github.com/docker/cli/pull/3302).

The OP repro does not work entirely properly on Windows, but it’s an issue with Docker CLI as bypassing compose-cli still exhibits it. I’ll have a look at fixing it there.
In a nutshell the first ctrl-c writes a single INT to the console, but then it appears to be all frozen and a second ctrl-c doesn’t show up. It does work properly if run with `-it` though.

Issue #1876, opened 2021-07-02
`docker-compose exec -T` doesn't properly handle stdin + EOF in compose-cli/docker-compose v2


rfay:
**Description**

If you try to use `docker-compose exec -T` with stdin, it doesn't properly exit when stdin gets to EOF. 

**Steps to reproduce the issue:**

Set up a service; `docker-compose up` with this docker-compose.yaml:

```yaml
services:
  test:
    image: busybox
    command: tail -f /dev/null
```

Now run `echo "hi there" | docker-compose exec -T test cat` 

**Describe the results you received:**

It outputs "hi there" and then hangs

**Describe the results you expected:**

It should output "hi there" and then exit/complete.

On docker-compose v1 this behaves correctly. 

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
Client:
 Cloud integration: 1.0.17
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.16.4
 Git commit:        f0df350
 Built:             Wed Jun  2 11:56:23 2021
 OS/Arch:           darwin/arm64
 Context:           desktop-linux
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.7
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       b0f5bc3
  Built:            Wed Jun  2 11:55:36 2021
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.4.6
  GitCommit:        d71fcd7d8303cbf684402823e425e9dd2e99285d
 runc:
  Version:          1.0.0-rc95
  GitCommit:        b9ee9c6314599f1b4a7f497e1f1f856fe433d3b7
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```
desktop-linux
```

**Output of `docker info`:**

```
Client:
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  compose: Docker Compose (Docker Inc., 2.0.0-beta.4)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 9
  Running: 9
  Paused: 0
  Stopped: 0
 Images: 223
 Server Version: 20.10.7
[...]

PR #1880, opened 2021-07-02
close sdtin stream on EOF to propagate exec process' stdin

ndeloof:
**What I did**
close hijacked http stdin stream on EOF to enforce container's exec process to get stdin closed and terminate accordingly.

**Related issue**
close https://github.com/docker/compose-cli/issues/1876

### DIFF
--- a/pkg/compose/exec.go
+++ b/pkg/compose/exec.go
@@ -72,7 +72,7 @@ func (s *composeService) Exec(ctx context.Context, project *types.Project, opts 
 	if err != nil {
 		return 0, err
 	}
-	defer resp.Close()
+	defer resp.Close() //nolint:errcheck
 
 	if opts.Tty {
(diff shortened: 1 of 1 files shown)


pkg/compose/exec.go
```diff
@@ -118,11 +118,13 @@ func (s *composeService) interactiveExec(ctx context.Context, opts api.RunOption
 			_, err := stdcopy.StdCopy(opts.Stdout, opts.Stderr, stdout)
 			outputDone <- err
 		}
+		defer stdout.Close() //nolint:errcheck
```
mat007, line 121: Any reason this is in a `defer` as it’s the last statement in the function?

ndeloof: nope, I just did the same as `stdin` without actually checking :P

mat007 (approved):

PR #1883, opened 2021-07-03
[RFC] Rename module to "v2" to fix broken v2.x tags.

thaJeztah:
This project created v2.x tags, but did not update the module paths to match.
As a result, the module is broken when using those tags (and will be broken
once v2.0.0 is tagged):

```bash
$ mkdir myproject && cd myproject
$ go mod init myproject
go: creating new go.mod: module myproject
```

A regular "go get" still works, because go modules do not consider pre-releases
of modules (alpha, beta, rc, or any SemVer version with a `-suffix`).

```bash
$ go get github.com/docker/compose-cli
go: downloading github.com/docker/compose-cli v1.0.17

$ go get github.com/docker/compose-cli@main
go: downloading github.com/docker/compose-cli v1.0.18-0.20210703113911-8d2ea5feb16f
go get: added github.com/docker/compose-cli v1.0.18-0.20210703113911-8d2ea5feb16f
```

However, trying to use any of the v2.x beta versions will fail:

```bash
$ go get github.com/docker/compose-cli@v2.0.0-beta.4
go get: github.com/docker/compose-cli@v2.0.0-beta.4: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

*This means that this module will be broken if `v2.0.0` (non-beta) is tagged.*

If I understand correctly, the `v2.x` tags are currently used for the `compose-cli`
_binary_ releases, and not (yet) for the module.

If the intent is to separate module releases (which *must* follow SemVer, so may
follow a different stability than binary releases) from binary releases, then
alternatives are:

1. Use tags for the binary that are non-SemVer, for example
    - `binary-v2.0.0-beta.4`
    - `docker-compose-v2.0.0-beta.4`
    - `docker-compose/v2.0.0-beta.4` (but, see below, this means there can
      never be a package/directory named `docker-compose` in the repository)
2. Make `cli/cmd` a separate module;
    - put a `go.mod` inside the `cli/cmd` with all dependencies for the command-line.
    - the module must then be named `github.com/docker/compose-cli/cli/cmd`. This
      may be a somewhat ugly name, so perhaps the path should be updated.
    - tag binary releases as `cli/cmd/vX.Y.Z`
    - add CI validation that none of the other code is importing `cli/cmd` to
      prevent introducing circular dependencies (which will be difficult to resolve)
    - given that the CLI will use the "main" module, make sure that the `go.mod`
      is updated to reflect the latest release of the main repository, and always
      tag binary and module "in tandem".
    - add `replace` rules to test the CLI code against the latest changes of the
      main module in the same branch.
3. Combination of the above. This assumes that code inside `cmd` is not intended
  to be used as a module by others.
    - create a `go.mod` in `cli/cmd` (or `cli`)
    - include dependencies that are specific to the cli
    - for "local" dependencies (main module), use `replace` rules
    - add a README outlining that code in cli is not for external consumption
    - use "non-SemVer" tags (as outlined in 1.)
    - CI validation must still be put in place to prevent the main module from
      importing the "cli" modules.
4. (This PR). Release the code as v2, and "retract" the existing v2.x-beta tags
   to mark them as "broken". This option means that binary releases and go module
   releases must follow the same release-cadence, including (SemVer) major version
   updates if the go code introduces a breaking change.
     - additional releases for the `v1` module (if desirable) must be released
       from a release branch
     - or (the official recommendation, which is _horrible_) v1 and v2 code live
       in the same branch, but separate copies of the code exist (root directory
       for "v1" and v1, v2, v3, v4 directories containing the code for each major
       release)

Options 1-3 allow binary releases to follow a different release schedule (and
"stability") than the "module" releases. This can help in situations where
the module introduces a breaking change (requiring it to increment the "major"
version (v1 -> v2 -> v3), without requiring the "binary" release to also be
released as a new major version.

From the options above, option 1 is the easiest to implement; go modules will
ignore the tags (as they're not SemVer), so from a "Go" perspective, those tags
do not exist. External users are still able to (not sure if recommended though)
consume code from `cli/cmd/**`, but must refer to it using the module version
(`v1.xx` currently).

Option 2 and 3 have the advantage that module dependencies that are *only* used
by the cli do not end up as a dependency for the main module, but option 2. will
be cumbersome ("submodules" are a pain if there's a chance of having inter-
dependencies in the same repository).

Option 3 combines the advantages of 1 and 2, but (as described) with the assumption
that the "cli" module must _never_ be used externally as a module, and that code
in the repository can _never_ import code from that directory.

Of course, if the intent is for the _module_ to also be released as `v2.x`, then
this pull request (Option 4.) is the right approach, with the caveats outlined
in the option.

Signed-off-by: Sebastiaan van Stijn <github@gone.nl>

**What I did**

**Related issue**


import-restrictions.yaml
```diff
@@ -4,13 +4,22 @@
     - github.com/docker/compose-cli/cli
     - github.com/docker/compose-cli/ecs
     - github.com/docker/compose-cli/local
+    - github.com/docker/compose-cli/v2/cli
+    - github.com/docker/compose-cli/v2/ecs
+    - github.com/docker/compose-cli/v2/local
```
thaJeztah, line 9: FWIW; looks like this configuration only looks at _literal_ paths, not subdirectories? (e.g. `aci/e2e` imports the CLI); https://github.com/docker/compose-cli/blob/7f18b47a9acfece3411bfada46d9565a7035d2bd/aci/e2e/e2e-aci_test.go#L53

And I think I saw some other paths as well

thaJeztah:
@mat007 @ndeloof @gtardif ptal: opened this for discussion

ndeloof:
The initial plan was to move the `cmd` and `pkg` content into github.com/docker/compose/v2 by the time we are confident to release as "compose V2 G.A.) and keep the compose-cli and cloud backends on this repo with the v1 tags.
But we can debate another approach that would be compliant with go.mod constraints

ndeloof:
closed as https://github.com/docker/compose-cli/pull/2055 has been merged

thaJeztah:
Does that work as expected? I see the module is named `github.com/docker/compose/v2`, but the github repository is of course named `github.com/docker/compose-cli` (with `-cli`);

https://github.com/docker/compose-cli/blob/v2/go.mod#L1

```go
module github.com/docker/compose/v2
```

ndeloof:
code will enventually be moved to github.com/docker/compose/v2 (this is an intermediate step, splitting codebase). So far compose-cli uses a `replace` rule to resolve the adequate github repo

PR #1897, opened 2021-07-06
Fix help text when for context create and kube backend

mikesir87:
**What I did**

Small fix to the CLI help text to align with the commands actually being used. Took me a while to figure out why it wasn't working! 😆


mat007 (approved):
Oh good catch, thanks!

PR #1901, opened 2021-07-07
Fix races

mat007:
**What I did**

I made the code consistent regarding loop variables capture for gorountines (ultimately we should probably use functions to split the code?). Also some variable copies before goroutines were missing.
I also serialized accesses to a map following crashes with «concurrent map read and map write» panics.

**Related issue**


lorenrh (approved):
LGTM, but wouldn't mind another pair of eyes!

ulyssessouza (approved):
LGTM

Issue #1893, opened 2021-07-06
docker compose ps xxx error behavior changed between v1 and v2


kinkerl:
**Description**

It looks like the behavior and exit codes of failure states changed with the docker compose ps command when looking for specific services. Also, the command help text now does not show the service as a potential parameter. 

**Steps to reproduce the issue:**

v1 CLI:

```
$ /usr/local/bin/docker-compose-old ps IDONTEXIST
ERROR: No such service: IDONTEXIST
$ echo $?
1

$ /usr/local/bin/docker-compose-old ps --help
List containers.

Usage: ps [options] [SERVICE...]
```

v2 CLI
```
$ docker compose ps IDONTEXIST
NAME                COMMAND             SERVICE             STATUS              PORTS
$ echo $?
0

$ docker compose ps --help
Usage:  docker compose ps
List containers
```

**Describe the results you received:**

We now see no error message and also more importantly, the error code of the command changed to a zero.

**Describe the results you expected:**

I would assume to get at least the same error code. This change in behavior actually broke our scripts and does not behave like a drop in replacement. 

**Output of `docker version`:**

v2 CLI
```
Docker version 20.10.7, build f0df350
Docker Compose version v2.0.0-beta.5
```

kinkerl:
:heart:

PR #1912, opened 2021-07-12
report error on `ps` when no container found for service(s)

ndeloof:
**What I did**
`compose ps` to return error when no container has been found for requested service.

**Related issue**
close https://github.com/docker/compose-cli/issues/1893


lorenrh (approved):
🎉

PR #1916, opened 2021-07-12
Allow unspecified protocol on exposed port in kube backend

mikesir87:
The Compose spec doesn't provide a default value, but the code
was defaulting to TCP in the service protocol, just not in the
name. If no protocol was specified, it would cause an invalid
service name (eg, "80-")

**What I did**

I updated the service name to use the same `toProtocol` function to remove the option for an invalid service name.


mat007 (approved):
Makes sense, thanks!
@aiordache does that look good to you?

Issue #1908, opened 2021-07-09
cannot use control characters in terminals via 'run' that work via 'up' + 'exec'


jakeonfire:
**Description**
When running a terminal shell via `docker compose run`, control characters like control, option/meta, and arrow keys do not work.

**Steps to reproduce the issue:**
1. write a simple `docker-compose.yml` like:
```
version: "3.8"
services:
  app:
    image: ubuntu:latest
```
2. start a shell with `docker compose run app bash`
3. (type a few characters, then) try any control keys, like ←, CTRL+A, META+←

**Describe the results you received:**
control characters are printed to the console, like `^[[D^A^[b`.

**Describe the results you expected:**
the commands should work as they do when following the same steps but with `docker compose up` and `docker compose exec app bash` for step # 2.

**Additional information you deem important (e.g. issue happens only occasionally):**
i suspect this may be related to https://github.com/docker/compose-cli/issues/1151

**Output of `docker compose version`:**
```
Docker Compose version v2.0.0-beta.6
```
(BTW this step should be updated in the template from "**Output of `docker-compose --version`: ***"

**Output of `docker version`:**
```
Client:
 Cloud integration: 1.0.17
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.16.4
 Git commit:        f0df350
 Built:             Wed Jun  2 11:56:22 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.7
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       b0f5bc3
  Built:            Wed Jun  2 11:54:58 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.6
  GitCommit:        d71fcd7d8303cbf684402823e425e9dd2e99285d
 runc:
  Version:          1.0.0-rc95
  GitCommit:        b9ee9c6314599f1b4a7f497e1f1f856fe433d3b7
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker context show`:**  
```
default
```

**Output of `docker info`:**
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  compose: Docker Compose (Docker Inc., v2.0.0-beta.6)
[...]

PR #1934, opened 2021-07-20
Restore support for control sequence in `compose run`

ndeloof:
**What I did**
Restore support for control sequence in `compose run`

**Related issue**
close https://github.com/docker/compose-cli/issues/1908

### DIFF
--- a/pkg/compose/attach.go
+++ b/pkg/compose/attach.go
@@ -24,11 +24,11 @@ import (
 
 	"github.com/compose-spec/compose-go/types"
 	"github.com/docker/cli/cli/streams"
-	"github.com/docker/compose-cli/pkg/api"
 	moby "github.com/docker/docker/api/types"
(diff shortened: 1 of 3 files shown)


nightpool:
Just ran into this after upgrade Docker Desktop. This completely breaks IRB, is there any way to get the fixed version working locally?

mat007:
> is there any way to get the fixed version working locally?

You would need to build it, see [BUILDING.md](https://github.com/docker/compose-cli/blob/main/BUILDING.md).
Or if you can wait a few days there is going to be an RC release of this project and you’ll be able to download it from the releases section. Otherwise it will be shipped in the next Docker Desktop release which we’re thinking is going to be mid-week next week.

nightpool:
okay, thanks, i can probably stumble along with docker-compose v1 until mid-week.

PR #1937, opened 2021-07-21
Print services in dependency order

milas:
**What I did**
Currently, `compose config --services` outputs the services in
a random/non-deterministic order.

In Compose v1, this was implicitly topologically sorted because
the project services were pre-sorted. With `compose-go`, the
services are unordered, and the `WithServices()` helper can be
used to iterate in dependency order.

**Tests**
There wasn't a great way to unit test this without further refactoring
as `runServices` uses `convertOptions::toProject()` which will
attempt to load a real Compose file. (AFAICT there aren't really any
existing CLI unit tests.)

If desirable, I can split the function in two so that the project
conversion from opts is separate from the actual logic which can
just take a `*types.Project` and so would be trivially testable.

image

### DIFF
--- a/cmd/compose/convert.go
+++ b/cmd/compose/convert.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/cnabio/cnab-to-oci/remotes"
 	"github.com/compose-spec/compose-go/cli"
+	"github.com/compose-spec/compose-go/types"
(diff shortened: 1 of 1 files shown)


milas:
Thanks for the quick turnaround! ❤️

PR #1942, opened 2021-07-22
stop using helm fork for oras dependency conflict

ndeloof:


thaJeztah:
Looks like the linter is picky

thaJeztah:
Possibly gRPC and some of the other deps may need to be pinned to older versions 🤔 they're the usual suspects, and often are incompatible 

Or ignore the linting errors, but possibly some changes would still be needed in the local code and/or elsewhere (haven't looked closely)

mat007 (approved):
Makes sense!

ulyssessouza (approved):
LGTM

PR #1943, opened 2021-07-22
Add a post-login message to promote personal access tokens

amineck:
**What I did**
- Add a post-login message to promote Personal Access Tokens https://docs.docker.com/docker-hub/access-tokens/
- Message will only appear if:
    1- User is using default registry
    2- User successfully logged in
    3- User is not using a Personal Access Token

```sh
> docker login -u amineck
Password:
Login Succeeded

Logging in with your password grants your terminal complete access to your account.
For better security, log in with a limited-privilege personal access token. Learn more at https://docs.docker.com/docker-hub/access-tokens/
````

### DIFF
--- a/cli/mobycli/exec.go
+++ b/cli/mobycli/exec.go
@@ -83,7 +83,7 @@ func Exec(root *cobra.Command) {
 		utils.DisplayScanSuggestMsg()
 	}
 	if command == "login" && !metrics.HasQuietFlag(commandArgs) {
-		utils.DisplayPATSuggestMsg(commandArgs)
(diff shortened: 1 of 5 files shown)


ndeloof (changes requested):
As this is a compose-cli only "feature", i.e. not intended to move into compose.v2, pat_suggest.go should not be part of `pkg` and could just be hosted within `cli` package

amineck:
Sure @ndeloof, I was trying to follow https://github.com/docker/compose-cli/blob/main/pkg/utils/scan_suggest.go

ndeloof (approved):

PR #1953, opened 2021-07-26
Post-login msg: Ignore if password is empty

amineck:
Signed-off-by: Amine Chouki <amine.chouki@docker.com>

**What I did**
Do not display the post-login message (to promote Personal Access Tokens) if the password is empty. This is more of a future proof check in case we introduce SSO or equivalent.


mat007 (approved):
Makes sense!

Issue #1917, opened 2021-07-12
Docker v2 Beta can't handle env file without "=" after variable name


potiuk:
**Description**

When you specify --env-file filename.env for the new docker-compose and the env file contains just variable names without "=" afterwards, the docker-compose command fails with "Can't separate key from value"

**Steps to reproduce the issue:**

1. Create `_docker.env` file with this content in a directory where you have a docker-compose file:
```
VARIABLE
```
2. Run `docker-compose up --env-file ${PATH}/_docker.env`

**Describe the results you received:**

The command fails with `Can't separate key from value` error

**Describe the results you expected:**

* It should not fail
* If VARIABLE is not defined in host, it should not be present in the container
* If VARIABLE is defined in host, it should be set in the container to the same value

**Additional information you deem important (e.g. issue happens only occasionally):**

Seems that adding `=` after the variable fixes the problem:

```
VARIABLE=
```

This issue was first discovered in  https://github.com/apache/airflow/issues/16949 and workarounded in https://github.com/apache/airflow/pull/16950 by @oyarushe for Apache Airflow. 

**Output of `docker-compose --version`: ***
```
v2.0.0-beta.6
```

**Output of `docker version`:**
```
Docker Desktop: 3.5.2 on MacOS
```

potiuk:
This workaround with adding `=` also has proven to break regular `docker --env-file` command. Previously, the same env file without `=` could be used for both `docker-compose` an `docker` but when you use the same pattern for docker env-file those variables get EMPTY:

```
VARIABLE=
```

potiuk:
Because of this docker-compose and docker different behaviour this is a major issue.

ndeloof:
need to check how https://pypi.org/project/python-dotenv/ and https://github.com/joho/godotenv differ for this specific syntax, and propose a fix

ndeloof:
python-dotenv explicitely consider lack of an equal sign as "value: none"
https://github.com/theskumar/python-dotenv/blob/master/src/dotenv/parser.py#L156-L160

ruby's dotenv parser don't declare the equal sign as optional
https://github.com/theskumar/python-dotenv/blob/master/src/dotenv/parser.py#L156-L160

Based on this I'm not sure godotenv would accept a PR to ignore a missing separator?

[...]

PR #1963, opened 2021-07-29
Fix OS's environment variable inheritance

ulyssessouza:
**What I did**
This is meant to be an intermediate a solution for OS inherited environment variables in https://github.com/joho/godotenv

This fix relies on a fork of https://github.com/joho/godotenv on my repository https://github.com/ulyssessouza/godotenv
Also in an alternative branch on [compose-go](https://github.com/compose-spec/compose-go/tree/inherited-dot-env-variables) to point to this fork and implement the changes needed in the file parsing.

**Related issue**
Resolves https://github.com/docker/compose-cli/issues/1917

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/awslabs/goformation/v4 v4.15.6
 	github.com/buger/goterm v1.0.0
 	github.com/cnabio/cnab-to-oci v0.3.1-beta1
-	github.com/compose-spec/compose-go v0.0.0-20210722130045-6e1e1c2b26de
+	github.com/compose-spec/compose-go v0.0.0-20210729053941-ef8509c7a5eb
 	github.com/containerd/console v1.0.2
 	github.com/containerd/containerd v1.5.0
 	github.com/distribution/distribution/v3 v3.0.0-20210316161203-a01c71e2477e
@@ -82,3 +82,5 @@ require (
 
 // (for buildx)
 replace github.com/jaguilar/vt100 => github.com/tonistiigi/vt100 v0.0.0-20190402012908-ad4c4a574305
+
+replace github.com/joho/godotenv => github.com/ulyssessouza/godotenv v1.3.1-0.20210720203245-d3051d55ab34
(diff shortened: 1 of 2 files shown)


mat007 (commented):
Makes sense, is there a way we can hand binaries to the OP in #1917 to have confirmation it addresses his issue?

mat007:
Also PR is still a draft.

ulyssessouza:
@mat007 We don't have a proper way to hand a binary to contributor by now. What I can propose is to checkout and build it.
(The easy way would be using the release page as a pre-release, but it's not meant to serve this propose)

mat007:
>  Obs: Just notice that the bump on compose-go is not to it's main branch so if it's bumped after that it will miss the fix

Ah, we would need to use our fork of godotenv fork in compose-go, do we want a PR there first?

mat007:
> replace don't apply on transitive dependencies, so this fix can't be applied on compose-go "only"

Yes, we need it in both compose-go and compose-cli, but in this PR if I understand correctly compose-cli uses a fork of compose-go. I don’t think we want that, do we?

ulyssessouza:
I just made a [PR](https://github.com/compose-spec/compose-go/pull/165) on `compose-go` to remove the `replace` and use the fork explicitly. Once it's merged in `compose-go` I will update this one to use it.

mat007:
But can’t we use a replace in both projects?

mat007:
Oh right, no we can’t see https://github.com/golang/go/issues/30354
That makes sense, and it’s what @ndeloof was actually saying 😬

go.mod
```diff
@@ -19,7 +19,7 @@ require (
 	github.com/awslabs/goformation/v4 v4.15.6
 	github.com/buger/goterm v1.0.0
 	github.com/cnabio/cnab-to-oci v0.3.1-beta1
-	github.com/compose-spec/compose-go v0.0.0-20210722130045-6e1e1c2b26de
+	github.com/compose-spec/compose-go v0.0.0-20210729053941-ef8509c7a5eb
```
mat007, line 22: This needs to be updated now that https://github.com/compose-spec/compose-go/pull/165 has been merged.

ulyssessouza:
Close that in favor of https://github.com/docker/compose-cli/pull/2002

PR #1965, opened 2021-07-29
Fix issue `docker compose rm -s` not removing containers

motemen:
**What I did**

When `docker compose rm` is run with `-s` (`--stop`), it actually does not remove containers, just stops them.
This PR fixes this issue.


mat007 (approved):
Oh good catch, thanks!

Issue #1950, opened 2021-07-24
Docker Compose v2 - "expose" value is ignored


gorbunovav:
**Description**

I have `node.js` service configured via `docker-compose.yml` based on stock `node:12.19.0-alpine` image with the `expose` parameter specified. With `Compose v2` I don't see exposed ports listed in the `docker inspect` output for the running container.

**Steps to reproduce the issue:**
1. Create `docker-compose.yml`, add some service based on the image without exposed ports specified inside the image.
2. Add `expose` parameter with some ports listed.
3. Launch the service.
4. Call `docker inspect` for the running container and look for the `ExposedPorts` entry.

**Describe the results you received:**

No exposed ports are defined for the running container.

**Describe the results you expected:**

Exposed ports are defined for the running container.

**Additional information you deem important (e.g. issue happens only occasionally):**

It works with the first version of the `docker-compose`.

**Output of `docker-compose --version`: ***
```
Docker Compose version v2.0.0-beta.6
```

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Cloud integration: 1.0.17
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        f0df350
 Built:             Wed Jun  2 11:56:47 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.7
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       b0f5bc3
  Built:            Wed Jun  2 11:54:58 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.6
  GitCommit:        d71fcd7d8303cbf684402823e425e9dd2e99285d
 runc:
  Version:          1.0.0-rc95
  GitCommit:        b9ee9c6314599f1b4a7f497e1f1f856fe433d3b7
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```
default
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker Application (Docker Inc., v0.8.0)
  buildx: Build with BuildKit (Docker Inc., v0.2.2-10-g3f18b65-tp-docker)
[...]

PR #1967, opened 2021-07-30
include exposed ports in container ports config

ndeloof:
**What I did**
added support for `expose` config as we compute container ports

**Related issue**
close https://github.com/docker/compose-cli/issues/1950


mat007 (approved):
Makes sense, but do we need a test?

Issue #1955, opened 2021-07-26
`docker compose run` brings down the cluster after one-off container finishes execution.


vintlucky777:
**Description**

**Steps to reproduce the issue:**
1. Bring up local cluster with `docker-compose up --abort-on-container-exit`
2. Run a one-off job with `docker-compose run`, e.g. for a Django setup `docker-compose run --rm web python ./manage.py migrate`

**Describe the results you received:**

Docker compose now detects a new container join the cluster. After one-off completes execution it terminates, which causes docker compose to bring down the whole thing.

**Describe the results you expected:**

Expected the `docker compose up` cluster to keep running, and one-off container to not interfere with it via log output or exit code.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker-compose --version`:**
```
$ docker-compose --version
Docker Compose version v2.0.0-beta.6
```

**Output of `docker version`:**

```
$ docker version
Client:
 Cloud integration: 1.0.17
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.16.4
 Git commit:        f0df350
 Built:             Wed Jun  2 11:56:22 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.7
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       b0f5bc3
  Built:            Wed Jun  2 11:54:58 2021
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          1.4.6
  GitCommit:        d71fcd7d8303cbf684402823e425e9dd2e99285d
 runc:
  Version:          1.0.0-rc95
  GitCommit:        b9ee9c6314599f1b4a7f497e1f1f856fe433d3b7
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```
$ docker context show
default
```

**Output of `docker info`:**

```
$ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  compose: Docker Compose (Docker Inc., v2.0.0-beta.6)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 5
  Running: 5
  Paused: 0
[...]

PR #1987, opened 2021-08-05
ignore one-off container events

ndeloof:
**What I did**
Filter-out one-off container watching events. This prevents `up` to consider a one-off container to be part of the app and attach to it, then exit when container stops

**Related issue**
close https://github.com/docker/compose-cli/issues/1955


mat007 (approved):
Makes sense!

Issue #1993, opened 2021-08-06
ps command not filtering container with `-q` or `--services` flags


amoncusir:
**Description**

The filtering (`--status` and `--filter`) parameter in `ps` command, only works as expected without the `-q` or `--services` flags. 

**Steps to reproduce the issue:**
1. Up `docker-compose` stack
2. Execute the ps command: `docker-compose ps --status 'exited' -q` and compare with the same command without the `--status 'exited'` argument.

**Describe the results you received:**

```
✗ docker-compose ps -q
4e345fe68ff303a5f21fe3b81e17762bbcf1529a0ae515d292d7852412659df2
9c853627faf7d84db874cfa1da7d4f8f811f94042a2b566d53b59f8f64802c46
82c7e25bcf230af5e68846f1605fcf7915e345fcba0b9d290c948fb62425a8c9
```

```
✗ docker-compose ps -q --status 'exited'
4e345fe68ff303a5f21fe3b81e17762bbcf1529a0ae515d292d7852412659df2
9c853627faf7d84db874cfa1da7d4f8f811f94042a2b566d53b59f8f64802c46
82c7e25bcf230af5e68846f1605fcf7915e345fcba0b9d290c948fb62425a8c9
```

**Describe the results you expected:**

```
✗ docker-compose ps -q
4e345fe68ff303a5f21fe3b81e17762bbcf1529a0ae515d292d7852412659df2
9c853627faf7d84db874cfa1da7d4f8f811f94042a2b566d53b59f8f64802c46
82c7e25bcf230af5e68846f1605fcf7915e345fcba0b9d290c948fb62425a8c9
```

```
✗ docker-compose ps -q --status 'exited'
```

And the right variant without '-q' flag:

```
✗ docker-compose ps --status 'exited' 
NAME                COMMAND             SERVICE             STATUS              PORTS
```

```
✗ docker-compose ps                   
NAME                COMMAND                  SERVICE             STATUS              PORTS
httpbin             "gunicorn -b 0.0.0.0…"   httpbin             running             0.0.0.0:80->80/tcp, :::80->80/tcp
memcached           "docker-entrypoint.s…"   memcached           running (healthy)   0.0.0.0:11211->11211/tcp, :::11211->11211/tcp
mysql               "docker-entrypoint.s…"   mysql               running (healthy)   0.0.0.0:3306->3306/tcp, :::3306->3306/tcp
```

**Additional information you deem important (e.g. issue happens only occasionally):**

[...]

PR #1995, opened 2021-08-06
apply filter before formatter

ndeloof:
**What I did**
apply filter before formatter is selected

**Related issue**
close https://github.com/docker/compose-cli/issues/1993


mat007 (approved):
Looks good!

Issue #1983, opened 2021-08-02
Compose V2 container `Image` attribute is inconsistent with V1 and prevents Watchtower


amarcu5:
**Description**

*Compose V2 beta 6*

The Image attribute of images started by docker compose is a hash in V2, but is the image name on V1. Although similar to #1972, this is important as it currently prevents Watchtower (https://github.com/containrrr/watchtower/issues/1019) and other auto container update systems from working.

**Steps to reproduce the issue:**
1. `docker compose up` the following Compose file:
```
version: "3.9"

services:
  nginx:
    image: nginx
```
2. Inspect the container.
3. Notice that the Image attribute is a sha256 hash.

**Describe the results you received:**
`"Image": "sha256:08b152afcfae220e9709f00767054b824361c742ea03a9fe936271ba520a0a4b"`

**Describe the results you expected:**
`"Image": "nginx"`

**Output of `docker version`:**
```
Docker version 20.10.7, build f0df350
```

**Output of `docker info`:**
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  compose: Docker Compose (Docker Inc., v2.0.0-beta.6)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 1
  Running: 1
  Paused: 0
  Stopped: 0
 Images: 1
 Server Version: 20.10.7
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e25210fe30a0a703442421b0f60afac609f950a3
 runc version: v1.0.1-0-g4144b63
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.4.0-77-generic
 Operating System: Ubuntu 20.04.2 LTS
 OSType: linux
 Architecture: x86_64
[...]

PR #1997, opened 2021-08-06
introduce ImageDigestLabel to track image built for service

ndeloof:
**What I did**
don't override service.image with local digest, better use a new label to track the image used to create service, and detect service need to be re-created after image has been re-built

**Related issue**
fix https://github.com/docker/compose-cli/issues/1983


ulyssessouza (approved):
LGTM

github-afsaneh:
> **What I did**
> don't override service.image with local digest, better use a new label to track the image used to create service, and detect service need to be re-created after image has been re-built
> 
> **Related issue**
> fix #1983

#1934

PR #2004, opened 2021-08-09
Ensure creation of Labels before assigning values

lorenrh:
Signed-off-by: Lorena Rangel <lorena.rangel@docker.com>

**What I did**

Ensure that labels are created before assigning values. 

image


mat007 (approved):
Makes sense, thanks!

mat007:
Closing as we actually reverted the initial PR which broke the build

PR #2006, opened 2021-08-09
Revert "introduce ImageDigestLabel to track image built for service"

lorenrh:
Reverts docker/compose-cli#1997 as it wasn’t actually passing CI.


mat007 (approved):
Thanks!

PR #2039, opened 2021-08-24
Error on unsupported flags

ulyssessouza:
**What I did**
Add checks to ECS, ACI, and Kubernetes backends.

/test-kube
/test-aci
/test-ecs
/test-windows


ndeloof (approved):

ndeloof (approved):

utils/check.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/docker/compose-cli/api/context/store"
+	"github.com/docker/compose-cli/cli/config"
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+)
+
+var ErrUnsupportedFlag = errors.New("unsupported flag")
+
+func CheckUnsupported(errs *multierror.Error, toCheck, expectedValue interface{}, commandName, msg string) *multierror.Error {
```
mat007, line 32: We don’t have to pass ` errs `  in here, we can just
```suggestion
func CheckUnsupported(toCheck, expectedValue interface{}, commandName, msg string) error {
```
as usual.

aci/compose.go
```diff
@@ -86,11 +87,36 @@ func (cs *aciComposeService) Copy(ctx context.Context, project *types.Project, o
...
+		{len(o.Start.AttachTo), 0, "attach-dependencies"},
+		{len(o.Start.ExitCodeFrom), 0, "exit-code-from"},
+		{o.Create.Recreate, "", "force-recreate"},
+		{o.Create.QuietPull, false, "quiet-pull"},
+		{o.Create.RemoveOrphans, false, "remove-orphans"},
+		{o.Create.Inherit, true, "renew-anon-volumes"},
+		{o.Create.Timeout, nil, "timeout"},
+	}
+	for _, c := range checks {
+		errs = utils.CheckUnsupported(errs, c.toCheck, c.expected, "up", c.option)
```
mat007, line 115: This should just be
```suggestion
		if err := utils.CheckUnsupported(c.toCheck, c.expected, "up", c.option); err != nil {
			errs = multierror.Append(errs, err)
		}
```

aci/compose.go
```diff
@@ -198,7 +239,16 @@ func (cs *aciComposeService) Ps(ctx context.Context, projectName string, options
 	return res, nil
 }
 
+func checkUnsupportedPsOptions(o api.PsOptions) error {
+	var errs *multierror.Error
+	errs = utils.CheckUnsupported(errs, o.All, false, "ps", "all")
+	return errs.ErrorOrNil()
```
mat007, line 245:
```suggestion
	return utils.CheckUnsupported(errs, o.All, false, "ps", "all")
```

cli/main.go
```diff
@@ -324,7 +324,7 @@ func exit(ctx string, err error, ctype string) {
 
 	if errors.Is(err, api.ErrNotImplemented) {
 		name := metrics.GetCommand(os.Args[1:])
-		fmt.Fprintf(os.Stderr, "Command %q not available in current context (%s)\n", name, ctx)
+		fmt.Fprintf(os.Stderr, "Command %q not available in current context (%s). %q\n", name, ctx, err)
```
mat007, line 327: Did you mean
```suggestion
		fmt.Fprintf(os.Stderr, "Command %q not available in current context (%s). %v\n", name, ctx, err)
```
maybe? It’s kind of unusual to use ` %q `  for errors, I think 🤔

cli/config/flags.go
```diff
@@ -35,10 +35,10 @@ type ConfigFlags struct {
 
 // AddConfigFlags adds persistent (global) flags
 func (c *ConfigFlags) AddConfigFlags(flags *pflag.FlagSet) {
-	flags.StringVar(&c.Config, config.ConfigFlagName, confDir(), "Location of the client config files `DIRECTORY`")
+	flags.StringVar(&c.Config, config.ConfigFlagName, ConfDir(), "Location of the client config files `DIRECTORY`")
 }
 
-func confDir() string {
+func ConfDir() string {
```
mat007, line 41: Any reason this is now exported?

utils/check.go
```diff
@@ -0,0 +1,54 @@
...
+	ctype := store.DefaultContextType
+	currentContext := config.GetCurrentContext("", config.ConfDir(), nil)
+	s, _ := store.New(config.ConfDir())
+	cc, _ := s.Get(currentContext)
```
mat007, line 36: Any reason for ignoring errors here?

utils/check.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	// Fixes type for posterior comparison
+	switch toCheck.(type) {
+	case *time.Duration:
+		if expectedValue == nil {
+			var nilDurationPtr *time.Duration
+			expectedValue = nilDurationPtr
+		}
+	}
+	if toCheck != expectedValue {
```
mat007, line 48: I think this would be more reliable
```suggestion
	if (toCheck != nil || expectedValue != nil) && toCheck != expectedValue {
```

kube/e2e/kube-simple-demo/demo_sentences.yaml
```diff
@@ -1,13 +1,9 @@
 services:
   db:
-    build: aci-demo/db
```
mat007, line 3: Was this deletion intentional?

kube/compose.go
```diff
@@ -287,9 +350,28 @@ func (s *composeService) Remove(ctx context.Context, project *types.Project, opt
...
+		toCheck, expected interface{}
+		option            string
+	}{
+		{o.Index, 0, "index"},
+		{o.Privileged, false, "privileged"},
+		{o.WorkingDir, "", "workdir"},
+	}
+	for _, c := range checks {
+		errs = utils.CheckUnsupported(errs, c.toCheck, c.expected, "exec", c.option)
+	}
```
mat007, line 371: nit: honestly this wouldn’t be worse, would it?
```suggestion
	errs = utils.CheckUnsupported(errs, o.Index, 0, "exec", "index")
	errs = utils.CheckUnsupported(errs, o.Privileged, false, "exec", "privileged")
	errs = utils.CheckUnsupported(errs, o.WorkingDir, "", "exec", "workdir")
```

aci/compose.go
```diff
@@ -86,11 +87,36 @@ func (cs *aciComposeService) Copy(ctx context.Context, project *types.Project, o
...
+	if err := checkUnsupportedUpOptions(options); err != nil {
+		return err
+	}
 	return progress.Run(ctx, func(ctx context.Context) error {
 		return cs.up(ctx, project)
 	})
 }
 
+func checkUnsupportedUpOptions(o api.UpOptions) error {
+	var errs *multierror.Error
```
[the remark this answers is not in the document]
ulyssessouza: Good catch!

aci/compose.go
ulyssessouza: The problem with this approach is that by taking this `if` outside of the function I will have to repeat it all over the code. And that's the very same `if`. Leaving this mutualised in the `CheckUnsupported` avoids this repetition.

aci/compose.go
```diff
@@ -198,7 +238,16 @@ func (cs *aciComposeService) Ps(ctx context.Context, projectName string, options
 	return res, nil
 }
 
+func checkUnsupportedPsOptions(o api.PsOptions) error {
+	var errs error
+	errs = utils.CheckUnsupported(errs, o.All, false, "ps", "all")
+	return errs
```
mat007, line 244: Maybe
```suggestion
	return utils.CheckUnsupported(nil, o.All, false, "ps", "all")
```
?

aci/compose.go
mat007: Yes, OK, good point!

aci/compose.go
ulyssessouza: Ops... That's pipeline effect

ecs/ps.go
```diff
@@ -56,3 +61,9 @@ func (b *ecsAPIService) Ps(ctx context.Context, projectName string, options api.
 	}
 	return summary, nil
 }
+
+func checkUnsupportedPsOptions(o api.PsOptions) error {
+	var errs error
+	errs = utils.CheckUnsupported(errs, o.All, false, "ps", "all")
+	return errs
```
mat007, line 68:
```suggestion
	return utils.CheckUnsupported(nil, o.All, false, "ps", "all")
```

utils/check.go
ulyssessouza: The problem here is that in GO, a literal `nil` is different of a typed `nil` when coming from a function parameter typed `interface{}`. 
That's why we need to check the type. If the expected is `nil` (literal), we need to check the type of the `toCheck` and change the expected to the same type. Even if the value is also `nil`.

See https://play.golang.org/p/vG12aUX1kvk

mat007: Yes, which is what the `(toCheck != nil || expectedValue != nil)` was supposed to cover, but it’s a bit trickier than that.
I’m merely concerned that this isn’t very generic and will only correctly handle `time.Duration`.
Here is a generic variant: https://play.golang.org/p/mOf1bwAFgN0

cli/config/flags.go
ulyssessouza: Used in check.go

cli/main.go
ulyssessouza: `%v` would show a Golang formatted string. As this is user facing we just want the error string

kube/compose.go
ulyssessouza: Done

utils/check.go
ulyssessouza: Cool! This is what I did at the beginning! But I just considered that was too complicated to handle just one type.
But if you don't think so I agree to implement it like that! 😸

cli/config/flags.go
mat007: Oh, right, indeed! Sorry.

utils/check.go
```diff
@@ -0,0 +1,60 @@
...
+var ErrUnsupportedFlag = errors.New("unsupported flag")
+
+func CheckUnsupported(errs error, toCheck, expectedValue interface{}, commandName, msg string) error {
+	ctype := store.DefaultContextType
+	currentContext := config.GetCurrentContext("", config.ConfDir(), nil)
+	s, err := store.New(config.ConfDir())
+	if err != nil {
+		return err
+	}
+	cc, err := s.Get(currentContext)
```
mat007, line 39: Wouldn’t it be safer to load this (from disk) just once?

cli/main.go
mat007: An error is a string, really 😁 For formatting both `%v` and `%s` for an error do the same: https://play.golang.org/p/MesffycUwLV

mat007: Maybe this should just be
```suggestion
		fmt.Fprintf(os.Stderr, "Command %q not available in current context (%s): %s\n", name, ctx, err)
```
?

utils/check.go
mat007: I think it’s probably fine. We can move it to `utils` as well.

utils/check.go
ulyssessouza: I'll pass that in the context then

mat007 (approved):
Thanks, looks good!

mat007:
Ah sorry, I think I merged a bit hastily as they were some errors. 😬

PR #2041, opened 2021-08-25
publish cross platform artifacts for PRs

ndeloof:
**What I did**
Build cross platform binaries and publish artifacts as part of PR builds to make it easier testing changes

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -1,0 +1,58 @@
+name: Publish Artifacts
+on:
+  issue_comment:
+    types: [created]
+jobs:
+  publish-artifacts:
+    if: github.event.issue.pull_request != '' && contains(github.event.comment.body, '/generate-artifacts')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go 1.16
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+        id: go
+
+      - name: Checkout code into the Go module directory
+        uses: actions/checkout@v2
+
+      - uses: actions/cache@v2
+        with:
+          path: ~/go/pkg/mod
+          key: go-${{ hashFiles('**/go.sum') }}
+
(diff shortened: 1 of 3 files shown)


ulyssessouza (changes requested):
I think we should have any mechanism to trigger that and not just generate the artifacts for all the PRs.

Maybe using something like the "/test-aci" tag to enable tests on ACI, for example.
A `/generate-artifacts` would be nice!

ndeloof:
@ulyssessouza makes sense.
unfortunately, this makes it harder to retrieve the matching GA workflow run to get binaries

ndeloof:
Added an extra step to update the PR comment with a link to workflow run page, so that user can access built binaries
see sample run on https://github.com/ndeloof/compose-cli/pull/4

.github/workflows/pending_answered.yml
```diff
@@ -1,4 +1,4 @@
-name: Continuous integration
+name: Remove pending lable on answer
```
mat007, line 1:
```suggestion
name: Remove pending label on answer
```

.github/workflows/artifacts.yml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          name: docker-compose-windows-amd64.exe
+          path: ${{ github.workspace }}/bin/docker-compose-windows-amd64.exe
+
+      - name: Update comment
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+          comment-id: ${{ github.event.comment.id }}
+          body: |
+            This PR can be tested using [binaries](https://github.com/docker/compose-cli/actions/runs/${{ github.run_id }})
```
mat007, line 56:
```suggestion
            This PR can be tested using [binaries](https://github.com/docker/compose-cli/actions/runs/${{ github.run_id }}).
```

mat007 (approved):
Nice!

Issue #2027, opened 2021-08-18
compose build <service> also rebuilds dependencies


mastacheata:
**Description**

Running `docker compose build <service>` on compose v1 resulted in rebuilding just that service. As of compose v2 it also rebuilds all services defined in depends_on

**Steps to reproduce the issue:**
1. Set up a compose file with 2 services A, B.
2. Make B depend on A
3. Run docker compose build B

**Describe the results you received:**
Both service A and B are rebuilt.

**Describe the results you expected:**
I expect only Service B to be rebuilt, without having to go through the build for service A

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker-compose --version`:**

```
Docker Compose version v2.0.0-beta.6
```

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Cloud integration: 1.0.17
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        f0df350
 Built:             Wed Jun  2 11:56:47 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.7
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       b0f5bc3
  Built:            Wed Jun  2 11:54:58 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.6
  GitCommit:        d71fcd7d8303cbf684402823e425e9dd2e99285d
 runc:
  Version:          1.0.0-rc95
  GitCommit:        b9ee9c6314599f1b4a7f497e1f1f856fe433d3b7
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```
default
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker Application (Docker Inc., v0.8.0)
  buildx: Build with BuildKit (Docker Inc., v0.2.2-10-g3f18b65-tp-docker)
  compose: Docker Compose (Docker Inc., v2.0.0-beta.6)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 44
  Running: 15
  Paused: 0
  Stopped: 29
 Images: 311
 Server Version: 20.10.7
[...]

PR #2044, opened 2021-08-26
Only build requested services

ndeloof:
**Related issue**
close https://github.com/docker/compose-cli/issues/2027


tim-timman:
@ndeloof Thanks. My searches didn't find that issue: all these different repositories is a mess. And their "responsibilities" don't stay the same... Having been burned by updates breaking things before, I haven't kept it up-to-date; guess I'll have to take my changes this time.

Issue #2021, opened 2021-08-17
"invalid mode" error for "run" command with "volumes_from" service configuration


tejksat:
**Description**

`docker compose run` fails with "invalid mode" error when Compose configuration contains "volumes_from" section.

**Steps to reproduce the issue:**
1. Create docker-compose.yml file with the following content:

```
version: "2"
services:
  app:
    image: busybox:latest
    volumes_from:
      - "container:${CONTAINER_ID}:ro"
```

2. Create new container that will be used as the source in `volumes_from` section:

```
$ docker run -d --rm alpine:latest nc -l -p 80
7cbc11ab9d41c19cbc78ce288725c1021f74dbf64a212d2eeebe37f3c9fee34b
```

3. Export container id as environment variable:

```
export CONTAINER_ID=7cbc11ab9d41c19cbc78ce288725c1021f74dbf64a212d2eeebe37f3c9fee34b
```

4. Run `app` service:

```
docker compose run app
```

**Describe the results you received:**

```
Error response from daemon: invalid mode: 7cbc11ab9d41c19cbc78ce288725c1021f74dbf64a212d2eeebe37f3c9fee34b:ro
```

**Describe the results you expected:**

```
Creating docker_compose_v2_app_run ... done
/ #
```

**Additional information you deem important (e.g. issue happens only occasionally):**

The issue is neither reproduced with `docker compose up` command nor with `docker-compose run app` command when using Docker Compose V1.

**Output of `docker-compose --version`:**

```
Docker Compose version v2.0.0-beta.6
```

**Output of `docker version`:**

```
Client:
 Cloud integration: 1.0.17
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.16.4
 Git commit:        f0df350
 Built:             Wed Jun  2 11:56:22 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.7
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       b0f5bc3
  Built:            Wed Jun  2 11:54:58 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.6
  GitCommit:        d71fcd7d8303cbf684402823e425e9dd2e99285d
 runc:
  Version:          1.0.0-rc95
  GitCommit:        b9ee9c6314599f1b4a7f497e1f1f856fe433d3b7
[...]

PR #2047, opened 2021-08-30
prepare volume for service set by `run`

ndeloof:
**What I did**
invoke `prepareVolumes` on service targetted by `run` 
as the global prepare logic is only applied to dependencies by the `run` execution flow

**Related issue**
close https://github.com/docker/compose-cli/issues/2021


ulyssessouza (approved):
LGTM

mat007 (commented):
Wouldn’t a few tests make sense? To document the use cases?

Issue #2032, opened 2021-08-19
docker-compose vs docker compose   different behaviour    and config.json  proxies


bb980701:
**Description**

docker-compose up was taking into account the default proxies setup in the config.json   as stated in doc : https://docs.docker.com/network/proxy/#set-the-environment-variables-manually

since Release 3.6.0 Docker desktop, docker-compose and docker compose refer to the same as version 
Docker Compose version v2.0.0-rc.1

**Steps to reproduce the issue:**
Set your /.docker/config.json with the
![image](https://user-images.githubusercontent.com/71716559/130103249-5017ff42-28aa-44e6-bd81-f12adf58aa8a.png)

Create the files :
![image](https://user-images.githubusercontent.com/71716559/130103317-58c60c04-3d47-41b2-bef9-be21165c57e9.png)

![image](https://user-images.githubusercontent.com/71716559/130103358-faac9207-d93e-41aa-b29b-28e5b23e8c2a.png)

![image](https://user-images.githubusercontent.com/71716559/130103389-d6cca697-1443-44f9-a8cf-34357102c578.png)

**Describe the results you received:**

**With Docker desktop 3.3.3**

*  docker-compose

docker-compose version  

docker-compose version 1.29.1, build c34c88b2
docker-py version: 5.0.0
CPython version: 3.9.0
OpenSSL version: OpenSSL 1.1.1g  21 Apr 2020

`docker-compose up`

Result is as expected by the documentation, environment variables were injected.

![image](https://user-images.githubusercontent.com/71716559/130103485-07ad30a0-50e1-4110-82c2-fc45c6df602e.png)

*  docker compose 

docker compose version
Docker Compose version 2.0.0-beta.1

`docker compose up`

Result is KO , no HTTP_PROXY,HTTPS_PROXY,http_proxy and https_proxy environment variables found

![image](https://user-images.githubusercontent.com/71716559/130105739-e799fa7f-738c-4fac-9caa-ffbeabb2def2.png)

**With Docker desktop 3.6.0**

docker-compose up
or
docker compose up

both refer to Docker Compose version v2.0.0-rc.1

and for both ; 
no HTTP_PROXY,HTTPS_PROXY,http_proxy and https_proxy environment variables found and so the container cannot reach internet :-(

![image](https://user-images.githubusercontent.com/71716559/130103671-f13ffdb8-d978-4551-b178-271af0ae7a72.png)
[...]

PR #2053, opened 2021-08-31
set proxy env vars based on CLI config

ndeloof:
**What I did**
added support for XXX_PROXY env variables based on docker config file

**Related issue**
https://github.com/docker/compose-cli/issues/2032


ulyssessouza (approved):
LGTM

Issue #2057, opened 2021-09-01
docker-compose vs docker compose different behaviour when ps with no running containers


mdantonio:
**Description**

When no containers are running `docker compose ps` fails with exit code 1 and error `not found`
On the opposite `docker-compose ps` was not failing (exit code 0). Since I have to adapt my scripts I would like to know if this is a new wanted behavior or a bug that should be fixed to make compose-cli behave like docker-compose

**Steps to reproduce the issue:**

docker-compose.yml
```
services:
  mytest:
    image: alpine
    command: sleep infinity
```

```
$ docker compose ps
not found

$ echo $?
1

$ docker-compose ps
Name   Command   State   Ports
------------------------------

$ echo $?
0
```

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker-compose --version`:**

```
$ docker-compose --version
docker-compose version 1.29.2, build unknown

$ docker compose version
Docker Compose version v2.0.0-rc.2

```

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Cloud integration: 1.0.17
 Version:           20.10.8
 API version:       1.41
 Go version:        go1.16.6
 Git commit:        3967b7d
 Built:             Fri Jul 30 19:54:27 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.8
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.6
  Git commit:       75249d8
  Built:            Fri Jul 30 19:52:33 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.8
  GitCommit:        7eba5930496d9bbe375fdf71603e610ad737d2b2
 runc:
  Version:          1.0.0
  GitCommit:        v1.0.0-0-g84113ee
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```
default
```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.6.2)
  compose: Docker Compose (Docker Inc., v2.0.0-rc.2)
[...]

PR #2061, opened 2021-09-01
Don't exit with code != 0 on `compose ps` empty result

ndeloof:
Don't exit with code != 0 when we succeed listing container but just none match compose project
Also includes a fix to e2e test flackyness

**Related issue**
close https://github.com/docker/compose-cli/issues/2057


mat007 (approved):
Makes sense!

PR #2074, opened 2021-09-13
Fixed the INSTALL.md file to take into account artifact packaging cha…

mreferre:
Fixed the INSTALL.md file to take into account artifact packaging changes

Signed-off-by: Massimo Re Ferre <massimo@it20.info>


mat007 (approved):
Thanks!

Issue #1920, opened 2021-07-13
Extend exec command to support specific instance name


mikesir87:
**Description**

I've been thinking about the `docker compose exec` command, especially in the context of the additional backends. As an example, if I'm using the kube backend and do a ps, I might get output like the following:

```shell
docker compose -p votingapp ps
NAME                      COMMAND             SERVICE             STATUS              PORTS
db-677f6fd579-7nhsj       ""                  db                  Running             
redis-6c7f9657c-jbdkk     ""                  redis               Running             
result-66b4856489-4pqz7   ""                  result              Running             
vote-65d4d9594f-4tmfb     ""                  vote                Running             
vote-65d4d9594f-m7jwm     ""                  vote                Running             
worker-7497c578cb-9p498   ""                  worker              Running             
worker-7497c578cb-h9rxl   ""                  worker              Running  
```

Now, if I saw in logs that I'm having issues with a specific replica (let's pretend `vote-*-m7jwm`), how do I exec into that specific instance? Since the k8s deployments are managing the replicas, the normal `--index` parameter 1) doesn't make sense and 2) doesn't look like it's currently supported in the backend.

Looking at this from a user's perspective, I'd expect I could take one of the names listed here and use that as an argument for `docker compose exec`. For example... `docker compose -p votingapp exec vote-65d4d9594f-m7jwm`.

So... my proposal/request is if the `SERVICE` option in `docker compose exec` could be extended to support a specific instance name. If not found, it falls back to treating it as a service name (as it currently does). I'd be happy to submit a PR for the kube backend, but recognize the other backends could probably use this feature too. But, before I do anything, I figured I'd start a conversation first. Thoughts?

ndeloof:
Once you have container name, you can just use plain `docker exec <container>` and don't need a compose command.

mikesir87:
@ndeloof That works when using a local compose setup, but doesn’t work in the context of the other backends, such as kube (where Docker might not even be available at all). The names that show up in the `docker compose ps` output are the pod names, so yes… I could switch and run `kubectl exec <pod-name>`, but I’m trying to allow teams to stay within the compose tooling.

ndeloof:
sure, but compose-cli is not just about compose, and `docker exec` (and other verbs) could be implemented as well by the Kube backend. cc @aiordache

mikesir87:
[...]

PR #2076, opened 2021-09-15
Allow compose exec to accept a specific instance in kube backend

mikesir87:
**What I did**

This PR provides the ability to perform a `docker compose exec <pod-name|service-name>` in the kube context and serves as an implementation of #1920. It looks at the requested name to see if it matches a pod name and exec's to it if a match is found. Otherwise, it falls back to the current implementation where it finds a matching pod by service.

The big reason to do this is to allow teams to stay within the compose tooling and exec into specific instances. In a normal Docker context, a user can simply use `docker exec`. But, that doesn't work in a kube context. They can look at logs to potentially find a problem, but have to switch to another tool to exec to the specific instance they want. And having to switch to `kubectl` and making sure you're in the right context/namespace... would be nice to just make it easier. 😄 

**Outstanding questions**

- I noticed that the ACI and ECS integrations currently don't support exec. No need to update those. But, does it make sense to update the Docker context exec as well to let you specify a specific instance you get from `docker compose ps`?
- What CLI/web docs can/should be updated, especially if the main context is to be updated.

**Related issue**

Issue #1920

### DIFF
--- a/kube/client/client.go
+++ b/kube/client/client.go
@@ -85,15 +85,21 @@ func (kc KubeClient) GetPod(ctx context.Context, projectName, serviceName string
 	if pods == nil {
 		return nil, nil
 	}
-	var pod corev1.Pod
+
+	for _, p := range pods.Items {
+		if p.Name == serviceName {
+			return &p, nil
+		}
+	}
+
 	for _, p := range pods.Items {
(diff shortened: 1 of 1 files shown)


aiordache (approved):
Thank you! Sorry for the delay 😅

PR #2078, opened 2021-09-21
Fix default values on ACI flags

ulyssessouza:
**What I did**
"always-recreate-deps" and "force-recreate" default to 'api.RecreateDiverged'

/test-aci
/test-ecs
/test-kube


mat007 (approved):
Looks good, is it still draft?

PR #2083, opened 2021-10-05
Remove obsolete workflow

mat007:
**What I did**

I deleted an obsolete github workflow.

This was used to clean up generated artifacts for compose v2 when using /generate-artifacts in a PR. Now that compose v2 and the associated workflows have been move to docker/compose, we don’t need it anymore.

See https://github.com/docker/compose-cli/runs/3799357123

**Related issue**


lorenrh (approved):
LGTM!

PR #2095, opened 2021-11-03
Fix plugin command delegation for version

ulyssessouza:
**What I did**
This makes the distinction between version command in "docker version" and "docker compose version"
In the second case it needs to delegate to the plugin.

Maybe another approach would be better. For example, with the full command chain as command identifier.
I just don't know if that's worth for now... Cases like this should remain very rare.

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -101,10 +101,14 @@ func isContextAgnosticCommand(cmd *cobra.Command) bool {
 	if cmd == nil {
 		return false
 	}
-	if _, ok := contextAgnosticCommands[cmd.Name()]; ok {
+	if _, ok := contextAgnosticCommands[cmd.Name()]; ok && isFirstLevelCommand(cmd) {
 		return true
 	}
 	return isContextAgnosticCommand(cmd.Parent())
+}
+
+func isFirstLevelCommand(cmd *cobra.Command) bool {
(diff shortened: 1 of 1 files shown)


mat007 (commented):
So until this change `docker compose version` was actually returning the version of `compose-cli`?
Is there no way now to get the version of that component?

mat007:
Well, it is set in the build though, it’s [that `GIT_TAG` env variable](https://github.com/docker/compose-cli/blob/main/Makefile#L44) inserted in the source code [here](https://github.com/docker/compose-cli/blob/main/builder.Makefile#L31).

mat007:
Also downloading the windows version:
```
$ ./docker-windows-amd64.exe version
Client:
 Cloud integration: v1.0.18
…
```
so it seems that’s the right version?

mat007:
Right sorry, I think I was getting confused again, this is what we’re trying to fix:
```
$ ./docker-windows-amd64.exe compose version
Docker Compose version dev
```

mat007 (commented):
Should we add another check to [TestCheckOwnCommand](https://github.com/docker/compose-cli/blob/fa05d4397ad4c3323fa5d15fef14355e1833adee/cli/main_test.go#L31)?

PR #2098, opened 2021-11-04
Revert "signals"

mat007:
Reverts docker/compose-cli#1879

Reverting as it’s breaking some integration tests downstream.
We’ll take another look later.

@ndeloof FIY ☝️


ulyssessouza:
@ndeloof We are reverting that cuz looks like it poses some problems in Docker Desktop.
It's not catching kill signals anymore.

The idea is reverting and reopening the issues related to this.
We will also open a new issue with the test case that fails in Docker Desktop so that the PR could be reworked to also take this case in consideration.

ulyssessouza (approved):
LGTM

PR #2104, opened 2021-11-16
Build metrics

crazy-max:
This PR handles metrics for build commands. It also adds a new field `Metadata` to the `Command` struct for extensibility. It could be used for other commands in the future.

Detecting build backend used by buildx is a bit tedious in the current release because no formatter is available to handle a proper output with `buildx ls`. The `ls` command locks the buildx store and that _can_ take some time (top 20s) to complete because it needs to check the liveliness of the nodes so user cannot use buildx during this time (https://github.com/docker/buildx/issues/764). That will cause some bad user experience if we rely on this command for metrics. I opened a pull request to start make things easier for us in the future (https://github.com/docker/buildx/pull/830).

So for our purpose we directly read from the store without using an actual command which is more effective:

```
$ cat ~/.docker/buildx/current
{"Key":"unix:///var/run/docker.sock","Name":"builder","Global":false}
# "builder" is the current instance used. read the instance info and retrieve the current driver:
$ cat ~/.docker/buildx/instances/builder
{"Name":"builder","Driver":"docker-container","Nodes":[{"Name":"builder0","Endpoint":"unix:///var/run/docker.sock","Platforms":null,"Flags":["--allow-insecure-entitlement","security.insecure","--allow-insecure-entitlement","network.host"],"DriverOpts":{"env.JAEGER_TRACE":"localhost:6831","image":"moby/buildkit:latest","network":"host"},"Files":null}],"Dynamic":false}
# "docker-container" is the current driver
```

The step above is valid only if user doesn't provide the `--builder` flag to the build command (or `BUILDX_BUILDER` env var). If the builder is provided, we only have to read from the store in `~/.docker/buildx/instances/<builder>`.

For the `docker build` command it's easier. We just need to read the cli config and check `DOCKER_BUILDKIT` env var as well as daemon config:

```json
{
  "features": {
    "buildkit": true
  }
}
```

There is also a special case if a builder alias is being used:

```
$ cat ~/.docker/config.json
{
        "credsStore": "desktop.exe",
        "aliases": {
                "builder": "buildx"
        }
}
```

The following metadata will be send along the command:

```json
{
        "build": {
                "cli": "docker | buildx | <alias>",
                "builder": "legacy | buildkit | <driver>"
        }
}
```

* `build.cli.docker`: user type `docker build ...`.
* `build.cli.buildx`: user type `docker buildx build ...` or `docker buildx bake ...`.
* `build.cli.<alias>` : user type `docker build ...` but an alias is used in `~/.docker/config.json` so set this one instead.
* `build.builder.legacy`: legacy builder is used with `docker build` command.
* `build.server.buildkit`: backend buildkit builder is used with `docker build` command.
* `build.server.<driver>`: a driver is used (can be `docker`, `docker-container` or `kubernetes` atm). Only for buildx.

Metrics commands have also been regenerated and sorted in separate commits.

cc @chris-crone @tonistiigi @mat007 @gtardif


cli/metrics/client.go
```diff
@@ -32,10 +32,11 @@ type client struct {
...
 type Command struct {
-	Command string `json:"command"`
-	Context string `json:"context"`
-	Source  string `json:"source"`
-	Status  string `json:"status"`
+	Command  string `json:"command"`
+	Context  string `json:"context"`
+	Source   string `json:"source"`
+	Status   string `json:"status"`
+	Metadata []byte `json:"metadata,omitempty"`
```
mat007, line 39: Should this not be
```suggestion
	Metadata map[string]interface{} `json:"metadata,omitempty"`
```
if it’s JSON?

cli/metrics/metrics.go
```diff
@@ -89,3 +97,18 @@ func GetCommand(args []string) string {
 	}
 	return result
 }
+
+type Metadata struct {
+	Build BuildMetadata `json:"build,omitempty"`
```
mat007, line 102: Should it just be a more generic `map[string]interface{}` here to avoid knowing about the internals of the build metrics?

cli/metrics/metrics.go
```diff
@@ -89,3 +97,18 @@ func GetCommand(args []string) string {
...
+}
+
+// GetMetadata returns the metadata linked to the invoked command
+func GetMetadata(command string, args []string) Metadata {
```
mat007, line 106:
```suggestion
func GetMetadata(command string, args []string) map[string]interface{} {
```
?

cli/metrics/metrics.go
```diff
@@ -89,3 +97,18 @@ func GetCommand(args []string) string {
...
+	var m Metadata
+	if command == "build" || command == "buildx" {
+		if bm := getBuildMetadata(command, args); (BuildMetadata{}) != bm {
```
mat007, line 109: That looks a bit dangerous to compare a struct by value, this means at this location of the code we know that it’s safe (e.g. no pointer in the struct). Maybe a pointer would be safer?

cli/metrics/metrics_build.go
```diff
@@ -0,0 +1,196 @@
...
+	var builder string
+	fset := pflag.NewFlagSet("buildx", pflag.ContinueOnError)
+	fset.String("builder", "", "")
+	_ = fset.ParseAll(buildArgs, func(flag *pflag.Flag, value string) error {
+		if flag.Name == "builder" {
+			builder = value
+		}
+		return nil
+	})
+	if len(builder) == 0 {
```
mat007, line 192: nit:
```suggestion
	if builder != "" {
```
?

chris-crone: @mat007 I haven't reviewed the PR but the suggested logic is the opposite of the original logic

cli/metrics/client.go
crazy-max: I think it's better to send plain bytes so server doesn't care about a specific type. After all it's just JSON.

cli/metrics/metrics.go
crazy-max: It's internal yes but we could reuse this object in an underlying logic that will import this package, let's say on pinata? But I think I should move that to another package under metrics actually like `metadata`.

cli/metrics/metrics_build.go
crazy-max: Yes not the same logic also len is more efficient that string comp.

mat007: There is no difference in the generated code, it’s just in my opinion easier to read with `== ""` as you know right away it’s a string.
See https://stackoverflow.com/a/18595217/1195832

crazy-max: Yeah more a special taste then as they said. Back in the days `len(s) == 0` made sense.

cli/metrics/metrics.go
crazy-max: Ok metadata is now under a dedicated package so we can reuse these types on pinata if we want.

cli/metrics/metrics.go
crazy-max: Ok I changed this behavior so we don't need to rely on comparison now.

crazy-max:
Putting in draft. I will add additional tests in the `e2e/cli-only` pkg.

cli/metrics/metadata/build.go
```diff
@@ -0,0 +1,192 @@
...
+		if n, ok := obj["Name"]; ok {
+			builder = n.(string)
+			// `Name` will be empty if `default` builder is used
+			// {
+			//   "Key": "unix:///var/run/docker.sock",
+			//   "Name": "",
+			//   "Global": false
+			// }
+			if len(builder) == 0 {
+				builder = "default"
```
mat007, line 127: Can’t we just
```suggestion
				return "docker"
```
here?

cli/metrics/metadata/metadata.go
```diff
@@ -0,0 +1,37 @@
...
+	metadatatypes "github.com/docker/compose-cli/cli/metrics/metadata/types"
+)
+
+// Get returns the JSON metadata linked to the invoked command
+func Get(command string, args []string) []byte {
+	var m metadatatypes.Metadata
+	if command == "build" || command == "buildx" {
+		m.Build = getBuildMetadata(command, args)
+	}
+	if (metadatatypes.Metadata{}) != m {
```
mat007, line 31: It looks like it’s comparing structs here as well?

cli/metrics/client.go
mat007: With bytes it’s going to write an unreadable blob, is this really what we want?
If the server is in JS for instance that’s not going to be very handy.
Even if it’s in Go it’s likely much easier to unserialize JSON all the way?

cli/metrics/metrics.go
mat007: I don’t really think we’ll do that though, we’re probably more going to just read it as a `map[string]interface{}` and add that to the heartbeat, aren’t we?

cli/metrics/client.go
mat007: E.g. https://play.golang.org/p/EQWj-1dC7it

cli/metrics/metrics.go
crazy-max: Oh ok if that's the case it makes more sense to do it this way yes.

cli/metrics/client.go
crazy-max: Looking at your comment https://github.com/docker/compose-cli/pull/2104#discussion_r751300156 we could use a map if you just want to put specific field to the heartbeat.
Or plain json as a string field instead of a slice of bytes so we push everything in `Metadata` to the heartbeat without bothering decoding its content. I think that's what the `Features` one does today.

cli/metrics/metadata/build.go
crazy-max: Sure we can, just wanted to be explicit on this special case where builder name can be empty.

gtardif:
Sorry I didn't comment in earlier on this ; adding what we just discussed on zoom with @crazy-max 
Adding the metadata as a new field in metrics might be a bit trick, because of the aggregation that is done in pinata. Desktop aggregates counts of commands, and metrics are "bucketted" by status, context, command. 
If we add metadata in the struct, I don't know how this will affect how we aggregate metrics in Desktop, and it might not be easy to find your information on the other end, in Looker. 
An easier way to go with this is adding a few values (well defined, short list) as "source", like `cli-build-container`, etc. and use these values when the buildx command is invoked. 
This will automatically translate in things we can query in looker (no need to change code in pinata).

Or we need to set another specific metrics field and also separate metrics based on that field in pinata (more bucketting than just based on status, context, cmd)

crazy-max:
@gtardif Thanks for your feedback! Using the existing `source` field sgtm.

crazy-max:
PTAL @mat007 @gtardif 

Removed `metadata` field and instead extends the `source` field as suggested. Will look like this:

```json
{"command":"build","context":"moby","source":"cli-docker;buildkit","status":"success"}
{"command":"buildx build","context":"moby","source":"cli-buildx;docker-container","status":"success"}
{"command":"buildx --builder build","context":"moby","source":"cli-buildx;docker-container","status":"success"}
```

cli/metrics/generatecommands/main.go
```diff
@@ -19,20 +19,26 @@ package main
...
 
 	"github.com/docker/compose/v2/pkg/utils"
 )
 
-var managementCommands = []string{"ecs", "scan"}
-
-var commands = []string{}
+var (
+	managementCommands = []string{"ecs", "scan"}
+	commands           = []string{"bundle", "completion", "install", "merge", "render", "split", "status", "uninstall", "validate"}
```
gtardif, line 30: I'm not sure I understand why we need this... should we define buildx (& compose) as management commands so that we walk through subcommands ?

cli/metrics/metadata/build.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	// unmarshal and returns `Driver`
+	var obj map[string]interface{}
+	if err = json.Unmarshal(rawBuilder, &obj); err != nil {
+		return driver
+	}
+	if d, ok := obj["Driver"]; ok {
+		driver = d.(string)
+	}
+	// TODO(@crazy-max): include buildkit version being used by this driver (e.g., docker-container;0.9.2)
```
gtardif, line 173: again, be careful about adding dynamic information in source value, as this will map to different fields in looker (and different columns in the DB behind looker). if we need version information, that should probably be reported differently in the desktop heartbeat, not in CLI execution metrics.

local/e2e/cli-only/e2e_test.go
```diff
@@ -210,6 +210,22 @@ func TestContextMetrics(t *testing.T) {
...
+		_ = os.Setenv("BUILDX_CONFIG", "./testdata/buildx-config")
+		c.RunDockerOrExitError("build", "-t", "foo:bar", ".")
+		c.RunDockerOrExitError("buildx", "build", "-t", "foo:bar", ".")
+		c.RunDockerOrExitError("buildx", "--builder", "graviton2", "build", "-t", "foo:bar", ".")
+
+		usage := s.GetUsage()
+		assert.DeepEqual(t, []string{
+			`{"command":"build","context":"moby","source":"cli-docker;buildkit","status":"failure"}`,
+			`{"command":"buildx build","context":"moby","source":"cli-buildx;docker-container","status":"failure"}`,
+			`{"command":"buildx --builder build","context":"moby","source":"cli-buildx;docker-container","status":"failure"}`,
```
gtardif, line 226: Nit : why use "-" as a first separator and ";" as a second one in source values here ; it might be more readable (and avoid some possible corner cases in looker) to just use "-" everywhere

cli/metrics/metadata/build.go
crazy-max: ah yes indeed, I have not updated this part about our conversation. will remove that comment for now.

local/e2e/cli-only/e2e_test.go
crazy-max: that's because buildx drivers can be `docker-container`, `docker` or `kubernetes` so I wanted to make a clear distinction of the `cli` and the `builder` being used otherwise we would have smth like `cli-buildx-docker-container` which is not really obvious compared to `cli-buildx;docker-container`. WDYT?

cli/metrics/generatecommands/main.go
crazy-max: somehow these commands were missing after the generation. see the specific commit: https://github.com/docker/compose-cli/pull/2104/commits/a4ae60a39d763cb4938396960bc3f8a8db26fe1b

local/e2e/cli-only/e2e_test.go
gtardif: right ; that should be ok, I don't foresee any issue with these values in looker

gtardif (approved):

PR #2114, opened 2021-11-24
Bump compose to v2.2.0 and compose-go to v1.0.5

mat007:
**What I did**

Bumped compose to v2.2.0 and compose-go to v1.0.5


ulyssessouza (approved):
LGTM

PR #2127, opened 2022-01-25
Remove the call to com.docker.cli to get the default context

rumpl:
**What I did**

Removed the call to com.docker.cli to get the default context that was [added here](https://github.com/docker/compose-cli/pull/401). This is no longer needed since the `docker compose` command was moved to a cli plugin for moby contexts

image


mat007 (approved):
Nice catch!

PR #2141, opened 2022-04-06
Add Stop Grace Period to compatibility file.

WoodellJ:
**What I did**
Add support for sending stop grace period down the pipe. This is already supported downstream but the compatibility file stops it. Updated the docs to reflect both this change and the fact that service.labels works as well.


WoodellJ:
Anyone available to review?

PR #2151, opened 2022-05-23
build metrics compatibility for next 22.06

crazy-max:
follow-up https://github.com/docker/cli/pull/3314 and https://github.com/docker/compose-cli/pull/2104

Ensures build metrics are future proof for the next Docker major release.

To detect if buildx will be used while typing `docker build ...`, we need to check the current docker cli version. I don't think there is another simple way, besides creating an infamous custom dimension in Looker. :sweat_smile:

cc @tonistiigi @chris-crone @gtardif @mat007 

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>


cli/main.go
```diff
@@ -86,6 +87,13 @@ func init() {
 	if err := os.Setenv("PATH", appendPaths(os.Getenv("PATH"), path)); err != nil {
 		panic(err)
 	}
+
+	// init metrics client
+	metricsClient = metrics.NewClient()
```
mat007, line 92: nit: isn’t the comment sort of paraphrasing the code on the next line? 😁

cli/metrics/metadata/build.go
```diff
@@ -31,23 +31,31 @@ import (
 	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/docker/api/types"
 	dockerclient "github.com/docker/docker/client"
+	"github.com/hashicorp/go-version"
 	"github.com/spf13/pflag"
 )
 
 // getBuildMetadata returns build metadata for this command
-func getBuildMetadata(cliSource string, command string, args []string) string {
+func getBuildMetadata(cliSource string, cliVersion string, command string, args []string) string {
```
mat007, line 39: nit:
```suggestion
func getBuildMetadata(cliSource, cliVersion, command string, args []string) string {
```

mat007 (approved):

cli/metrics/metrics.go
```diff
@@ -24,23 +24,32 @@ import (
...
 	}
 }
 
+func (c *client) getMetadata(cliSource string, args []string) string {
+	if len(args) == 0 {
+		return cliSource
+	}
+	switch args[0] {
+	case "build", "buildx":
+		cliSource = metadata.BuildMetadata(cliSource, c.cliVersion(), args[0], args[1:])
```
rumpl, line 48: Calling this here will make build commands start slower (while we execute the cli to get the version), are we ok with this?

crazy-max: Yes that's my concern too :disappointed:. Not sure how we can reliably returns the docker cli version without invocation.

crazy-max:
Switch from shelling out to retrieve docker cli version by using the [`debug/buildinfo` module](https://pkg.go.dev/debug/buildinfo). Atm docker cli strips DWARF symbol table so will be empty for current 22.06 but will be solved by https://github.com/docker/cli/pull/3717.

PTAL @mat007 @rumpl @glours

nicksieger (approved):

PR #2171, opened 2022-07-08
Use Docker CLI socket from home

p1-0tr:
**What I did**

We should not rely on having a global path for the Docker CLI socket. On
macOS this forces Docker Desktop to access directories which require
raised privileges. Whereas on Linux we do not create sockets in that
location at all, currently. So look for the Docker CLI socket in the
User's home directory.

**Related issue**


cli/metrics/conn_other.go
```diff
@@ -1,7 +1,8 @@
-// +build !windows
+//go:build !windows,!darwin
+// +build !windows,!darwin
 
 /*
-   Copyright 2020 Docker Compose CLI authors
+   Copyright 2020, 2022 Docker Compose CLI authors
```
thaJeztah, line 5: I think the validation currently doesn't allow for the years to differ between files, so all should have the same years

cli/metrics/conn_darwin.go
```diff
@@ -0,0 +1,41 @@
...
+	"path/filepath"
+)
+
+var (
+	socket = "/var/run/docker-cli.sock"
+)
+
+func init() {
+	// Attempt to retrieve the Docker CLI socket for the current user.
+	if home, err := os.UserHomeDir(); err == nil {
```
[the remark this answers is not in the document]
p1-0tr: Good point :) thanks

cli/metrics/conn_other.go
p1-0tr: Yeah :( I even did a quick google for some more decent copyright header checker, but couldn't find one which does the right thing

thaJeztah: In https://github.com/docker/cli-docs-tool/, we've been trying Google addlicense, which seemed to work (a bit more "loosely" in checking); see https://github.com/docker/cli-docs-tool/pull/4

.gitignore
```diff
@@ -1,3 +1,24 @@
 bin/
 dist/
 /.vscode/
+
+## VIM
```
mat007, line 5: Shouldn’t all that be in your global gitignore on your machine?

p1-0tr: Sure, that will work as well. Putting those in project-specific configs is force of habit for me.

cli/metrics/conn_windows.go
```diff
@@ -30,6 +31,10 @@ var (
 	socket = `\\.\pipe\docker_cli`
 )
 
+func init() {
+	overrideSocket() // nop, unless built for e2e testing
```
mat007, line 35: Did you mean
```suggestion
	overrideSocket() // no-op, unless built for e2e testing
```
or something like that?

glours (approved):
Sounds good to me, I'm just waiting for an approval from someone of the Pos team to merge it

cli/metrics/conn_windows.go
p1-0tr: `nop` is the mnemonic for no-op in most assemblers (and tends to be used as a neologism for "do nothing"), but yeah that would be clearer.

p1-0tr: done :)

cli/metrics/conn_darwin.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+var (
+	socket = "/var/run/docker-cli.sock"
+)
+
+func init() {
+	// Attempt to retrieve the Docker CLI socket for the current user.
+	if home := homedir.Get(); home != "" {
+		socket = filepath.Join(home, "/Library/Containers/com.docker.docker/Data/docker-cli.sock")
```
djs55, line 36: Maybe we should use a shorter Linux-style path in `$HOME/.docker/desktop/docker-cli.sock` even on Darwin because we've had some problems where the maximum path in the sockaddr is 104 characters (even shorter than Linux's 108 characters). We work around this in some of our binaries by calling `os.Chdir("Library/Containers/com.docker.docker")` and then using a relative path... but I suspect this trick doesn't work for the CLI because changing the current directory probably breaks something.

WDYT?

p1-0tr: Hmm, this is a tough one.

The best thing I can think of from the top of my head is:

    pushd ...
    connect with relative path
    popd
    
but not sure that would be safe to do.

djs55: Reflecting on it a bit, I'm not as concerned as I was because the socket is best-effort (and it will work probably 99% of the time). So I'm happy with it as-is.

djs55 (approved):

PR #2177, opened 2022-07-25
Fall back to global CLI socket on Linux

p1-0tr:
**What I did**

DD relies on having the CLI socket available in /var/run when running in
WSL. So in case we can't find the CLI socket in the calling user's home
directory, fall back to /var/run.

**Related issue**


glours (approved):
LGTM

djs55 (approved):

PR #2178, opened 2022-07-26
cli binary path handling

nicksieger:
**What I did**

Update the way we choose which `com.docker.cli` binary executable to use.

1. Use path from env variable `$DOCKER_COM_DOCKER_CLI` if defined and non-empty
2. Use path formed from parent directory of current process executable joined with
   `com.docker.cli`
3. Look for `com.docker.cli` on $PATH

**Related issue**

Butterfly

### DIFF
--- a/cli/mobycli/exec.go
+++ b/cli/mobycli/exec.go
@@ -24,6 +24,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"regexp"
+	"runtime"
 
 	"github.com/docker/compose/v2/pkg/compose"
 	"github.com/docker/compose/v2/pkg/utils"
@@ -38,7 +39,13 @@ import (
 var delegatedContextTypes = []string{store.DefaultContextType}
 
 // ComDockerCli name of the classic cli binary
-const ComDockerCli = "com.docker.cli"
+var ComDockerCli = "com.docker.cli"
+
+func init() {
+	if runtime.GOOS == "windows" {
+		ComDockerCli += ".exe"
+	}
+}
 
 // ExecIfDefaultCtxType delegates to com.docker.cli if on moby context
 func ExecIfDefaultCtxType(ctx context.Context, root *cobra.Command) {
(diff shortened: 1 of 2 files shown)


glours (approved):
LGTM
Maybe we can add tests to be sure all the options are properly covered, wdyt?

gabriellavengeo:
Looks good to me too, +1 on the adding tests suggestion

gabriellavengeo (approved):

cli/mobycli/exec.go
```diff
@@ -131,6 +122,25 @@ func RunDocker(childExit chan bool, args ...string) error {
 	return cmd.Run()
 }
 
+func comDockerCli() string {
+	if v := os.Getenv("DOCKER_COM_DOCKER_CLI"); v != "" {
+		return v
+	}
+
+	execBinary := findBinary(ComDockerCli)
```
StefanScherer, line 130: Nasty detail on Windows. It seems that `findBinary()` needs to append the `.exe` extension to successfully find the binary in the directory of the `docker.exe` binary. As it only checked `com.docker.cli` without extension it failed and then the `LookPath()` is called, and then the `com.docker.cli.exe` could be found in any folder of my current PATH.

Screenshot 2022-07-27 at 12 38 03

StefanScherer: This is only needed for the Windows binary `docker.exe`, the Linux binary that we use in the WSL integration already works fine, I successfully tested it in a WSL distro on my Windows machine.

nicksieger: Good catch, thanks. I am still confused by which Go APIs work without having to append .exe and which don't. Apparently the exec-related API's and LookPath don't...

glours (approved):
Thanks for adding the tests @nicksieger

cli/mobycli/exec.go
nicksieger: Oh, I see, it's the os.Stat call that needs it. At any rate the current fix is probably best to make sure we only run `com.docker.cli.exe` and not a .bat, etc...

PR #2185, opened 2022-08-24
docs: applies wording suggested in #2154

### LINKED ISSUE #2154: /compose-cli/docs/aci-container-features.md typo
From: https://github.com/docker/docker.github.io/issues/12172
https://github.com/docker/compose-cli/blob/main/docs/aci-container-features.md

> A single container is executed in its own ACI container group, that will container only one container.

This line needs to be modified to

> A single container is executed in its own ACI container group, which will contain a single container.

actualben:
**What I did**
I applied a wording fix to `docs/aci-container-features.md` suggested in #2154, wherein the sentence:

> A single container is executed in its own ACI container group, that will container only one container.

becomes:

> A single container is executed in its own ACI container group, which will contain a single container.

**Related issue**

closes #2154

### DIFF
--- a/docs/aci-container-features.md
+++ b/docs/aci-container-features.md
@@ -6,7 +6,7 @@ keywords: Docker, Azure, Integration, ACI, container, cli, deploy, cloud
 # Azure Container Instances: running single containers
 
 Single containers can be executed on ACI with the `docker run` command.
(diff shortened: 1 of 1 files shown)


StefanScherer (approved):
Thanks @actualben!
LGTM

PR #2191, opened 2022-09-14
cli: fix warnings on 'docker context use'

nicks:
**What I did**
fix warnings on 'docker context use'

**Related issue**
Fixes https://github.com/docker/cli/issues/3765

PXL_20220806_232348931 MP

### DIFF
--- a/cli/cmd/context/use.go
+++ b/cli/cmd/context/use.go
@@ -17,12 +17,9 @@
 package context
 
 import (
-	"fmt"
-
 	"github.com/spf13/cobra"
 
-	"github.com/docker/compose-cli/api/config"
-	"github.com/docker/compose-cli/api/context/store"
+	"github.com/docker/compose-cli/cli/mobycli"
 )
 
 func useCommand() *cobra.Command {
@@ -31,22 +28,8 @@ func useCommand() *cobra.Command {
 		Short: "Set the default context",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runUse(args[0])
+			mobycli.Exec(cmd.Root())
+			return nil
 		},
 	}
-}
-
-func runUse(name string) error {
-	s := store.Instance()
-	// Match behavior of existing CLI
-	if name != store.DefaultContextName {
-		if _, err := s.Get(name); err != nil {
-			return err
-		}
-	}
-	if err := config.WriteCurrentContext(config.Dir(), name); err != nil {
-		return err
-	}
-	fmt.Println(name)
-	return nil
 }
--- a/local/e2e/cli-only/e2e_test.go
+++ b/local/e2e/cli-only/e2e_test.go
@@ -206,7 +206,7 @@ func TestContextMetrics(t *testing.T) {
 			`{"command":"context use","context":"moby","source":"cli","status":"success"}`,
 			`{"command":"ps","context":"local","source":"cli","status":"success"}`,
 			`{"command":"stop","context":"local","source":"cli","status":"failure"}`,
-			`{"command":"context use","context":"local","source":"cli","status":"success"}`,
+			`{"command":"context use","context":"moby","source":"cli","status":"success"}`,
 			`{"command":"ps","context":"local","source":"cli","status":"success"}`,
 			`{"command":"context ls","context":"moby","source":"cli","status":"success"}`,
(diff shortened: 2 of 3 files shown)


nicks:
this is more of a "question in the form of a PR". I don't really understand why these cobra Command implementations forked the upstream implementations - I didn't see anything that the fork was adding. I would also be open to trying to reconcile the differences, but this way seemed easier.

local/e2e/cli-only/e2e_test.go
```diff
@@ -203,10 +203,10 @@ func TestContextMetrics(t *testing.T) {
 		assert.DeepEqual(t, []string{
 			`{"command":"context create","context":"moby","source":"cli","status":"success"}`,
 			`{"command":"ps","context":"moby","source":"cli","status":"success"}`,
-			`{"command":"context use","context":"moby","source":"cli","status":"success"}`,
+			`{"command":"context use","context":"local","source":"cli","status":"success"}`,
 			`{"command":"ps","context":"local","source":"cli","status":"success"}`,
 			`{"command":"stop","context":"local","source":"cli","status":"failure"}`,
-			`{"command":"context use","context":"local","source":"cli","status":"success"}`,
+			`{"command":"context use","context":"moby","source":"cli","status":"success"}`,
```
nicks, line 209: this changed because now, we capture the context AFTER the command ran instead of before. i'm not sure if this matters.

nicksieger:
I can't tell much from the commit history, it looks like the context commands here evolved around the same time or a little bit after those in docker/cli. Originally maybe there were reasons for that? @rumpl @ulyssessouza?

cli/cmd/context/use.go
```diff
@@ -31,22 +28,8 @@ func useCommand() *cobra.Command {
 		Short: "Set the default context",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runUse(args[0])
+			mobycli.Exec(cmd.Root())
```
rumpl, line 31: @nicks did you check that the `mobycli` can `context use` a context that's created by compose-cli?

nicks: as far as i can tell, they're interoperable -- they read and write from the same metadata tree under `~/.docker/contexts/meta`. Here's the sequence of commands I ran:

```
nick@grumpy:~/src/compose-cli$ com.docker.cli context create ecs ecs-test --from-env 
unknown flag: --from-env
See 'docker context create --help'.
nick@grumpy:~/src/compose-cli$ go run ./cli/main.go context create ecs ecs-test --from-env 
Successfully created ecs context "ecs-test"
nick@grumpy:~/src/compose-cli$ com.docker.cli context list
NAME            DESCRIPTION                               DOCKER ENDPOINT                                 KUBERNETES ENDPOINT                                 ORCHESTRATOR
default *       Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                     https://kubernetes.docker.internal:6443 (default)   swarm
desktop-linux                                             unix:///home/nick/.docker/desktop/docker.sock                                                       
ecs-test        credentials read from environment                                                                                                             
nick@grumpy:~/src/compose-cli$ com.docker.cli context use ecs-test
ecs-test
Current context is now "ecs-test"
nick@grumpy:~/src/compose-cli$ com.docker.cli context use default
default
Current context is now "default"
nick@grumpy:~/src/compose-cli$ com.docker.cli context rm ecs-test
ecs-test
nick@grumpy:~/src/compose-cli$ com.docker.cli context list
NAME            DESCRIPTION                               DOCKER ENDPOINT                                 KUBERNETES ENDPOINT                                 ORCHESTRATOR
default *       Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                     https://kubernetes.docker.internal:6443 (default)   swarm
desktop-linux                                             unix:///home/nick/.docker/desktop/docker.sock
```

do you think it makes sense to add an integration test for it?

rumpl (approved):
LGTM, I was afraid docker/cli wouldn't be able to use or remove the context because compose-cli contexts have some additional data but it seems to work so 👍

cli/cmd/context/use.go
rumpl: I think it might make sense yes, I know that the context management code was changed recently in docker/cli, hopefully the change still handles contexts with additional metadata

Issue #2115, opened 2021-11-25
AWS ECS Fargate ARM64 support:


thorfi:
Description

AWS have just announced support for ARM64 in Fargate:

https://aws.amazon.com/about-aws/whats-new/2021/11/aws-fargate-amazon-ecs-aws-graviton2-processors/

It would be good to have docker compose auto detect platform linux/arm64 and include CpuArchitecture: ARM64 chunk in the CloudFormation template, and also to have the docker/ecs-secrets-sidecar and docker/ecs-searchdomain-sidecar containers built as multi-platform images.

```
  Resources:
    FooTaskDefinition:
      Properties:
        RuntimePlatform:
          CpuArchitecture: ARM64
```

thorfi:
Re-raised in this repository from docker/compose#8958 as instructed.

thorfi:
Bump

thorfi:
bump

thorfi:
bump

JohnPreston:
@thorfi 

This is something supported and implemented in ECS Compose-X

https://docs.compose-x.io/syntax/compose_x/ecs.details/ecs.html#cpuarchitecture

This is what "I call" a service level extension field, which will take care of setting everything in the rendered CloudFormation task/service definitions. If you set a multi-containers "family" then the value needs only setting once, or set everywhere with the same value.

The value is disabled **in cloudformation** if you change the launch type or capacity providers so that it only works with Fargate.

Hope this helps you in your journey with docker-compose and AWS ECS.

stephenhibbert:
Could I please clarify if I have an arm64 docker image am I able today to run this on ECS Graviton2 in any way whatsoever? Even if I manually specify that it should be run on CpuArchitecture: ARM64 or does compose-cli _**only**_ support x86 docker images? I haven't been able to find the answer in the documentation, sorry if I missed something.

KymPohl:
In my case the only item preventing an docker ecs based deployment to aws arm64 is the generated CloudFormation template's reference to the docker/ecs-searchdomain-sidecar:1.0 image with the TaskDefinition section(s). This docker/ecs-searchdomain-sidecar:1.0 image is only built for an amd64 platform. Seems that a simply remedy would just be for Docker to push an arm64 version of this component to docker hub (needs to have the 1.0 tag though unless they change the template generation logic...

PR #2195, opened 2022-09-23
Added Linux/ARM64 to ECS sidecar builds fixes #2190 and #2115

### LINKED ISSUE #2190: No ARM64 version of docker/ecs-searchdomain-sidecar which is needed for Docker ECS context
I'm using a docker ecs context to produce a CloudFormation template on a linux/arm64 platform, and targeting a linux/arm64 deployment in aws. Running my docker compose up (using the ecs context) produces the TaskDefinition below as part of the outputted CloudFormation template. This TaskDefinition makes reference to a docker/ecs-searchdomain-sidecar:1.0 image which apparently only supports an amd64 architecture. Consequently the "_ResolvConf_InitContainer" component of the resulting deployment fails to startup on the target arm64 platform (runtime error --> "exec /resolv: exec format error"). FWIW, if I remove the relevant section of the TaskDefinmition of the template everything deploys fine to aws and executes without issue on the target arm64 platform.

So it seems that I either need to somehow locate an arm64-compatible image of ecs-searchdomain-sidecar:1.0, or is there a different recommended approach for using docker compose using an ecs context to perform an arm64 deployment to aws fargate?

Docker Hub entry for ecs-searchdomain-sidecar ... https://hub.docker.com/r/docker/ecs-searchdomain-sidecar/tags

```
  ApiTaskDefinition:
    Properties:
      ContainerDefinitions:
        - Command:
            - us-west-2.compute.internal
            - buildingscience-analyzer.local
          Essential: false
          Image: docker/ecs-searchdomain-sidecar:1.0
          LogConfiguration:
            LogDriver: awslogs
            Options:
              awslogs-group:
                Ref: LogGroup
              awslogs-region:
                Ref: AWS::Region
              awslogs-stream-prefix: buildingscience-analyzer
          Name: Api_ResolvConf_InitContainer
        - DependsOn:
            - Condition: SUCCESS
              ContainerName: Api_ResolvConf_InitContainer
          Environment:
            - Name: SERVER_PORT
              Value: "80"
          Essential: true
          Image: <my-app-image>
          LinuxParameters: {}
          LogConfiguration:
            LogDriver: awslogs
            Options:
              awslogs-group:
                Ref: LogGroup
              awslogs-region:
                Ref: AWS::Region
              awslogs-stream-prefix: buildingscience-analyzer
          Name: api
          PortMappings:
            - ContainerPort: 80
              HostPort: 80
              Protocol: tcp
      Cpu: "256"
      ExecutionRoleArn:
        Ref: ApiTaskExecutionRole
      Family: buildingscience-analyzer-api
      Memory: "512"
      NetworkMode: awsvpc
      RequiresCompatibilities:
        - FARGATE
      RuntimePlatform:
        CpuArchitecture: ARM64
        OperatingSystemFamily: LINUX
    Type: AWS::ECS::TaskDefinition
```

mtelvers:
**What I did**
Updated `Makefile` to use `docker buildx` to build the ECS sidecar images for both AMD64 and ARM64.

**Related issue**
ECS Fargate supports both AMD64 and ARM64.  Running `docker --context ecs compose up` introduces sidecar images which are only built for AMD64.  This results in all deployments to ARM64 failing. Publishing ARM64 images would fix issue #2190 and #2115.

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,16 +116,16 @@ publish-aci-sidecar: build-aci-sidecar ## build & publish aci sidecar image with
 	docker pull docker/aci-hostnames-sidecar:$(tag) && echo "Failure: Tag already exists" || docker push docker/aci-hostnames-sidecar:$(tag)
 
 build-ecs-search-sidecar:  ## build ecs search sidecar image locally and tag it with make build-ecs-search-sidecar tag=0.1
-	docker build -t docker/ecs-searchdomain-sidecar:$(tag) ecs/resolv
+	docker buildx build --platform linux/amd64,linux/arm64 -t docker/ecs-searchdomain-sidecar:$(tag) ecs/resolv
 
 publish-ecs-search-sidecar: build-ecs-search-sidecar ## build & publish ecs search sidecar image with make publish-ecs-search-sidecar tag=0.1
-	docker pull docker/ecs-searchdomain-sidecar:$(tag) && echo "Failure: Tag already exists" || docker push docker/ecs-searchdomain-sidecar:$(tag)
+	docker pull docker/ecs-searchdomain-sidecar:$(tag) && echo "Failure: Tag already exists" || docker buildx build --push --platform linux/amd64,linux/arm64 -t docker/ecs-searchdomain-sidecar:$(tag) ecs/resolv
 
 build-ecs-secrets-sidecar:  ## build ecs secrets sidecar image locally and tag it with make build-ecs-secrets-sidecar tag=0.1
-	docker build -t docker/ecs-secrets-sidecar:$(tag) ecs/secrets
+	docker buildx build --platform linux/amd64,linux/arm64 -t docker/ecs-secrets-sidecar:$(tag) ecs/secrets
 
 publish-ecs-secrets-sidecar: build-ecs-secrets-sidecar ## build & publish ecs secrets sidecar image with make publish-ecs-secrets-sidecar tag=0.1
-	docker pull docker/ecs-secrets-sidecar:$(tag) && echo "Failure: Tag already exists" || docker push docker/ecs-secrets-sidecar:$(tag)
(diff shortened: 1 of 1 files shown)


djs55 (approved):

nicksieger (approved):

taco-chen:
@StefanScherer Really looking forward to have the multi-arch release, which will make the compose file more elegant and tidy. Thanks!

mfittko:
Is there an issue I could follow in order to be notified when the image will be available as multi arch? We'd like to deploy on ECS Fargate with Graviton 2, but the lack of a multi arch sidecar image kinda blocks us here.

mfittko:
While cloud formation overlays are generally a great thing, it's utter nonsense for task definitions as you need to overlay _all_ ContainerDefinitions, so you have to places in the compose file for defining the containers, including environment variables etc. It would be great if you could only overwrite the image in this case, but what we need is just a multi arch image version of [docker/ecs-searchdomain-sidecar:1.0](https://hub.docker.com/r/docker/ecs-searchdomain-sidecar/tags) on docker hub, no? Why is it so complicated to provide this? Any amd64 based deployments should still work while it's possible for AMD deployments to work without the overlay.

@StefanScherer so IMO it should be okay to overwrite the 1.0 tag like this when it's based on the exact same Dockerfile - you'd save many people from running into this.
```
docker buildx build . -f Dockerfile -t "docker/ecs-searchdomain-sidecar:1.0" -t "docker/ecs-searchdomain-sidecar:latest" --platform=linux/arm64,linux/amd64 --push
```

But feel free to push for a more sophisticated solution, if that's what you think we need here. It could definitely made much easier to run Graviton 2 deployments on AWS ECS Fargate when this would work out of the box.

mfittko:
Again, what is the reason not to push a multi arch build to docker hub already? I mean not even the docker compose ecs integration needs to be updated since the CloudFormation does not contain the sha256 but only the (1.0) tag name, so any docker compose generated CF deployment would be able to just keep working, no @StefanScherer ? Would be great if you could either do this or at least give an answer why it is delayed further? We'd really love to get our workloads out on ARM without that ugly workaround. Thanks 🙏

PR #2204, opened 2022-12-06
cli/mobycli: Exec(): simplify output

thaJeztah:
The command now was showing 2 informational messages (about PAT and Scan); the scan one had been there for a while so most users will know about it by now.


thaJeztah:
Thanks!

Issue #2213, opened 2023-01-10
ECS: Allow toggling AssignPublicIp


BackSlasher:
/!\ Docker Compose V2 has moved to [github.com/docker/compose](https://github.com/docker/compose/tree/v2), this repository is for "Cloud Integrations". You can report issues related to `docker compose` [here](https://github.com/docker/compose/issues/new/choose).

**Description**

Redo of #2135.
All tasks are assigned a public IP.
Combined with #1783, this creates a bit of a security gap.
Assume the following:
```
services:
  caddy:
    image: caddy
    ....
    ports:
      - 80
      - 443
    networks:
      - backend
  sensitive_backend:
    image: python
    ...
    networks:
      - backend
networks:
- backend: {}
```
While nginx is open to the public (by publishing ports and causing a LB to be attached), we don't want `sensitive_backend` to be exposed.
However, they're both assigned a public IP and being joined to this secgroup:

![image](https://user-images.githubusercontent.com/7194491/211659672-6ff9c31c-2580-46ee-b493-98db902098a6.png)

This effectively allows public access to the container.
The steps that IMO should be taken are:
1. Make public IPs opt-in (this)
2. Redo the security group assignment (#1783)

**Steps to reproduce the issue:**
See compose file above

**Describe the results you received:**
All services are assigned a public IP address

**Describe the results you expected:**
Only services asking for a public IP (if any) should be assigned one

**Additional information you deem important (e.g. issue happens only occasionally):**
I'm not sure any service should have a public IP considering access should be done via LBs, but it's cheap to allow an optin.

**Output of `docker-compose --version`:**

```
(paste your output here)
```

**Output of `docker version`:**

```
Docker version 20.10.22, build 3a2c30b63a
```

**Output of `docker context show`:**  
You can also run `docker context inspect context-name` to give us more details but don't forget to remove sensitive content.

```

    {
        "Name": "nitz-ecs",
        "Metadata": {
            "Type": "ecs"
        },
        "Endpoints": {
            "docker": {
                "SkipTLSVerify": false
            },
            "ecs": {
                "Profile": "nitz"
            }
        },
        "TLSMaterial": {},
        "Storage": {
            "MetadataPath": "/home/nitz/.docker/contexts/STUFF",
            "TLSPath": "/home/nitz/.docker/contexts/tls/STUFF"
        }
    }
]

```

**Output of `docker info`:**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
[...]

PR #2214, opened 2023-01-10
ECS: supporting x-aws-assign_public_ip

BackSlasher:
Allowing users to choose whether their instances should get a public IP address or not. Defaulting to "no".

**What I did**
Supporting `x-aws-assign_public_ip` in the service level.
If not present / set to a falsy value, the service should not get a public IP address.
If set to a truthy value, the current behavior exists (set by the provider's capability).
I'm not touching the EC2 section as it seems a bit messed up anyway, and I suspect it's not covered by tests.
**Related issue**

#2213

Created new "slightly complex" test to cover my change. I intend to reuse it later as well.

Called Dubi. A friendly street cat.
image


BackSlasher:
comment for me: 6be190299d5626070b07219bb1d0032e385c2bc2

BackSlasher:
From testing, it seems that tasks without a public IP are unable to reach docker to pull images:
image

I might change the default to be "true" then, or move to changing the security group structure insteaad

PR #2219, opened 2023-01-30
metrics: add debug implementation

milas:
Refactor the metrics client slightly to add the concept of a "reporter" that actually sends the metrics.

The primary implementation is HTTP-based as before, but a key difference now is that there's another variant that writes to a file for simplified debugging. This is activated by setting the `DOCKER_METRICS_DEBUG_LOG` environment variable to a file path to write events to as JSON lines.


glours (approved):
LGTM

PR #2223, opened 2023-02-06
metrics: move helper code from Compose

milas:
Compose doesn't actually currently use this code, it's only used here. Moving it inline so that we can drop it from Compose in the future or make changes as needed without worrying about this as a dependency.

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -256,7 +256,7 @@ func main() {
 	if err = root.ExecuteContext(ctx); err != nil {
 		handleError(ctx, err, ctype, currentContext, cc, root)
 	}
-	metricsClient.Track(ctype, os.Args[1:], compose.SuccessStatus)
+	metricsClient.Track(ctype, os.Args[1:], metrics.SuccessStatus)
 }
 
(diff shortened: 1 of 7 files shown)


cli/main.go
```diff
@@ -300,20 +300,21 @@ $ docker context create %s <name>`, cc.Type(), store.EcsContextType), ctype)
 
 func exit(ctx string, err error, ctype string) {
 	if exit, ok := err.(cli.StatusError); ok {
-		metricsClient.Track(ctype, os.Args[1:], compose.SuccessStatus)
+		// TODO(milas): shouldn't this use the exit code to determine status?
+		metricsClient.Track(ctype, os.Args[1:], metrics.SuccessStatus)
```
glours, line 304: I agree, we should use the exit code and send the associated status IHMO

glours (approved):
LGTM

cli/main.go
milas: Yeah, I believe this is a long-standing bug 😅 

I'll confirm today and open a quick PR that just adds the exit code -> status wrapper call here, we have all the info available, so I think this was a copy-paste mistake rather than intentional

PR #2225, opened 2023-02-06
metrics: initial logic for command execution duration

milas:
Add a timer around command invocations to be reported with metrics.
This isn't actually sent anywhere currently, as it's meant for
evented data which is forthcoming. (We could report it with the
current events, but it's not clear that there's any value in doing
so.)

The signature for `Track()` has been changed to take an object with
all the fields. This is both for sanity to keep the method from
getting ridiculously long, and to make it easier to unify the usage
(heartbeat) and event code paths by ensuring we have all the data
for both.

(Replaces #2224 which I accidentally closed and can't re-open 😅)


glours (approved):
LGTM

PR #2226, opened 2023-02-07
metrics: track and report non-aggregated events

milas:
Currently, "usage" reports are sent, which are aggregated.
This adds an "event" variant, which won't be aggregated.

For compatibility, the "usage" events remain and nothing has
changed in terms of how they parse the command args, for example.
This will ensure continued functioning of anything that relies
on these metrics.

For the "event" variants, the CLI parsing is slightly different
in an attempt to improve data analysis capabilities while still
being respectful of user privacy (i.e. only known values of Docker
CLI commands/flags are ever recorded). Additionally, execution
duration information is now reported with these events.


cli/metrics/docker_command.go
```diff
@@ -0,0 +1,191 @@
...
+		if arg == "--help" || arg == "-h" {
+			usage = true
+		}
+	}
+
+	event := &DockerCLIEvent{
+		Command:      cmdPath[1].name,
+		Subcommand:   subcommand,
+		ExitCode:     int32(cmd.ExitCode),
+		Usage:        usage,
```
roryv-docker, line 86: Wasn't quite sure how to interpret `usage` var?

cli/main.go
```diff
@@ -260,7 +260,7 @@ func main() {
 		handleError(ctx, err, ctype, currentContext, cc, root, start, duration)
 	}
 	metricsClient.Track(
-		metrics.CmdMeta{
```
milas, line 263: (Sorry for the rename - there's so many types with `command` or `event` in their name since we've got 2x very similar code paths here now, so I've been _trying_ to keep things clear)

glours (approved):

PR #2240, opened 2023-04-12
metrics: scout command usage

paco-valverde:
**What I did**
Add to the command whitelist the `scout` command and the `cves` subcommand for getting usage metrics


glours (approved):
LGTM

PR #2245, opened 2023-05-15
cli: deprecation notice

ndeloof:
https://docker.atlassian.net/browse/ENV-189

**What I did**
added a deprecation notice when compose-cli is used with a non-default context type

```
$ docker context create ecs toto
Cloud integration is DEPRECATED. Read more on https://docs.docker.com/cloud/ecs-integration/
? Create a Docker context using:  [Use arrows to move, type to filter]
..
```

**(not mandatory) A picture of a cute animal, if possible in relation with what you did**

PR #2247, opened 2023-05-18
cli: allow suppressing deprecation warning

milas:
**What I did**
Added `COMPOSE_CLOUD_EOL_SILENT` env var, which will suppress the EOL message if set to a truthy value.

**Related issue**
#2245


nicksieger (approved):
👍

PR #2249, opened 2023-06-01
metrics: adjust Compose event collection

milas:
* for per-invocation events, only look at Compose `alpha` commands
* for aggregated events, include Compose `alpha` commands


glours (approved):
LGTM

PR #2251, opened 2023-06-02
metrics: include integer exit code on daily

milas:
**What I did**
Include the exit code itself. The status field (which a limited string enum inferred from some exit codes) remains to not break anything.


glours (approved):
👍

cmrigney (approved):
Awesome, thank you!

PR #2252, opened 2023-06-06
feat: display docker scout hints on build and pull

eunomie:
**What I did**

On `docker build` and `docker pull` commands, display a hint to `docker scout quickview`.

Hints are enabled by default.
To disable them:
- use the `--quiet`/`-q` quiet flag
- set the `DOCKER_CLI_HINTS` environment variable with a false (according to Go) value
- add a `-x-cli-hints` configuration in `~/.docker/config.json` file under `{plugins: {-x-cli-hints: {enabled}}}`

In case of `build`, if `--push`, `--progress` or `--output` flag are used, hints are not displayed to keep it working on main (and less advanced) case.

There's no hint for `buildx build`.

In case of a `docker build` the `docker scout quickview` command doesn't need an argument as it will take the most recently built image by default.
In case of a `docker pull` the pulled image is extracted from the command arguments.

The output already handles plain/tty display:
image

**Related issue**

https://github.com/docker/team-ssc-dev-features/issues/306

bela-bako-CT0FXtDtYII-unsplash


cli/mobycli/exec.go
```diff
@@ -111,6 +111,18 @@ func Exec(_ *cobra.Command) {
 	}
 	commandArgs := os.Args[1:]
 	command := metrics.GetCommand(commandArgs)
+	if (command == "build" || command == "pull") && !metrics.HasQuietFlag(commandArgs) {
+		var image string
+		if command == "pull" {
+			for _, a := range commandArgs[1:] {
+				if !strings.HasPrefix(a, "--") {
+					image = a
```
thaJeztah, line 119: This probably won't work for `docker build` as it takes the build context as positional argument, not an image.

Also note that;

- build may not build an image (in case `--output` is used)
- build allows controlling the output format (plain/tty), ik which case we may have to adjust the output accordingly

eunomie: Yes, this part is only for pull, not build. In case of a build we don't care that much as we already have a shortcut to get the most recently built image.

But I'll check:
- the output so it's only displayed when we build an image
- I'll check the output format

eunomie: The hint already handle plain/tty, I'll just check if there's other way to enable/disable it for build for instance
image

eunomie:
I updated some things:
- to disable them from the `~/.docker/config.json` file
- to disable them with a more generic `DOCKER_CLI_HINTS` env variable
- to disable them in more advanced/unknown build scenario (use of progress, output. push flags)

eunomie:
Also added a `CliHintsEnabled()` that can be used if we introduce more hints

cli/mobycli/cli_hints.go
```diff
@@ -0,0 +1,68 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
```
ChrisChinchilla, line 2: Does this year need updating?

cli/mobycli/cli_hints_test.go
```diff
@@ -0,0 +1,168 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
```
ChrisChinchilla, line 2: Same comment here

cli/mobycli/scout_suggest.go
```diff
@@ -0,0 +1,73 @@
...
+		return
+	}
+	if len(image) > 0 {
+		image = " " + image
+	}
+	out := os.Stderr
+	b := color.New(color.Bold)
+	_, _ = fmt.Fprintln(out)
+	_, _ = b.Fprintln(out, "What's Next?")
+	_, _ = fmt.Fprintf(out, "  View summary of image vulnerabilities and recommendations → %s", color.CyanString("docker scout quickview%s", image))
```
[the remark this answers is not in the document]
eunomie: 🤔 I don't know, analysis is really more vague than "vulnerabilities and recommendations". Especially that's what will be displayed by the proposed command.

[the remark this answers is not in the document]
eunomie: I don't know, I think I'm not aware of those discussions and it's still what we are displaying in the `docker scout` help so I'd more in favour to keep it consistent with the CLI.

glours (approved):

cli/mobycli/exec.go
```diff
@@ -111,9 +111,18 @@ func Exec(_ *cobra.Command) {
 	}
 	commandArgs := os.Args[1:]
 	command := metrics.GetCommand(commandArgs)
-	if command == "login" && !metrics.HasQuietFlag(commandArgs) {
-		displayPATSuggestMsg(commandArgs)
+	if !metrics.HasQuietFlag(commandArgs) {
+		switch command {
+		case "build": // only on regular build, not on buildx build
```
crazy-max, line 116: `build` command redirects to `buildx build` since Docker 23 and I don't see in the `metrics.HasQuietFlag` anything checking the BuildKit progress quiet mode.

cli/mobycli/scout_suggest.go
```diff
@@ -0,0 +1,73 @@
...
+
+func displayScoutQuickViewSuggestMsgOnPull(args []string) {
+	image := pulledImageFromArgs(args)
+	displayScoutQuickViewSuggestMsg(image)
+}
+
+func displayScoutQuickViewSuggestMsgOnBuild(args []string) {
+	// only display the hint in the main case, build command and not buildx build, no output flag, no progress flag, no push flag
+	if utils.StringContains(args, "--output") || utils.StringContains(args, "-o") ||
+		utils.StringContains(args, "--progress") ||
```
crazy-max, line 37: Ah I see `--progress` but not `BUILDKIT_PROGRESS` env var being taken into account though.

crazy-max: I think we should just check `--progress` in `metrics.HasQuietFlag`.

eunomie: I'll add the env var 👍

cli/mobycli/scout_suggest.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/fatih/color"
+)
+
+func displayScoutQuickViewSuggestMsgOnPull(args []string) {
+	image := pulledImageFromArgs(args)
+	displayScoutQuickViewSuggestMsg(image)
+}
+
+func displayScoutQuickViewSuggestMsgOnBuild(args []string) {
+	// only display the hint in the main case, build command and not buildx build, no output flag, no progress flag, no push flag
```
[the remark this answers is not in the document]
eunomie: I'm not sure to understand.
Here I check the command flags the user is passing, not what it's done internally. So if it's a `buildx build` command hint will not be displayed, as well as if it's a `build` command with `--progress` or `--output` flags.
Then the `build` command can redirect to `buildx build` but it's more an implementation detail here, no?

gtardif (approved):

PR #2259, opened 2023-07-25
compose-cli: fix PAT detection for PAT suggestion

landism:
**What I did**

### Problem
When logging in with a PAT, `docker login` tells me... I should be using a PAT:
```
docker login
Authenticating with existing credentials...
Login Succeeded

Logging in with your password grants your terminal complete access to your account.
For better security, log in with a limited-privilege personal access token. Learn more at https://docs.docker.com/go/access-tokens/
```

compose-cli is looking for the prefix `dckrp_` but hub currently issues pats with the prefix `dckr_pat_`.

### Solution
Also check for the prefix `dckr_pat_`. (maybe we could just exclusively check for this prefix? I have not looked into the history of `dckrp_`)

**Related issue**


landism:
@milas apparently I don't have access to add reviewers in this repo

rumpl (approved):

milas (approved):

PR #2262, opened 2023-08-22
scout hint: tell user logging in is required

laurazard:
The `docker scout quickview` hint, displayed after a user pulls or builds an image, only works if the user is logged in to Hub.

**What I did**

Check if user is logged in, and make hint more explicit/call this requirement out if not.

Logged in (same behaviour as before):

image

Logged out (add message about logging in):

image

image

### DIFF
--- a/cli/mobycli/scout_suggest.go
+++ b/cli/mobycli/scout_suggest.go
@@ -20,8 +20,11 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
+	"github.com/docker/cli/cli/config"
 	"github.com/docker/compose/v2/pkg/utils"
+	"github.com/docker/docker/registry"
 
 	"github.com/fatih/color"
 )
@@ -53,10 +56,14 @@ func displayScoutQuickViewSuggestMsg(image string) {
 	}
 	out := os.Stderr
 	b := color.New(color.Bold)
-	_, _ = fmt.Fprintln(out)
-	_, _ = b.Fprintln(out, "What's Next?")
-	_, _ = fmt.Fprintf(out, "  View summary of image vulnerabilities and recommendations → %s", color.CyanString("docker scout quickview%s", image))
-	_, _ = fmt.Fprintln(out)
+
+	_, _ = b.Fprintln(out, "\nWhat's Next?")
+	if !hubLoggedIn() {
+		_, _ = fmt.Fprintln(out, "  1. Sign in to your Docker account → "+color.CyanString("docker login"))
+		_, _ = fmt.Fprintln(out, "  2. View a summary of image vulnerabilities and recommendations → "+color.CyanString("docker scout quickview"+image))
+	} else {
+		_, _ = fmt.Fprintln(out, "  View a summary of image vulnerabilities and recommendations → "+color.CyanString("docker scout quickview"+image))
+	}
 }
 
 func pulledImageFromArgs(args []string) string {
@@ -73,4 +80,28 @@ func pulledImageFromArgs(args []string) string {
 		}
 	}
 	return image
+}
+
+// hubLoggedIn checks whether the user has credentials configured
+// for Docker Hub. If it fails to get a result within 100ms, it
+// short-circuits and returns `true`.
+// This can be an expensive operation, so use it mindfully.
+func hubLoggedIn() bool {
+	result := make(chan bool)
+	go func() {
+		creds, err := config.LoadDefaultConfigFile(nil).GetAllCredentials()
+		if err != nil {
+			// preserve original behaviour if we fail to fetch creds
+			result <- true
+		}
+		_, ok := creds[registry.IndexServer]
+		result <- ok
+	}()
+	select {
+	case loggedIn := <-result:
+		return loggedIn
+	case <-time.After(100 * time.Millisecond):
+		// preserve original behaviour if we time out
+		return true
+	}
 }


thaJeztah:
So, looking at this hint:

- logging won't be needed until I run that command
- and it's unlikely that a user will now do a `docker login` _if they're not planning to run the related `docker scout` command_

I'm slightly wondering if this is "too much information" at this moment in time, and cluttering the hint (hiding away the _relevant_ information in the noise). Effectively, this is listing prerequisites to run the `docker scout` command, and we should not end up with "make sure you have a working internet connection: to connect to the internet ..."

So, silly idea:

- Can we make `docker scout` check if the user is authenticated before the command is executed?
- Or.. make it "try", and fallback to asking the user to authenticate?
- ^^^ we should actually consider doing the same for `docker pull` (etc)


That would also allow us to include some of the wording we added in https://github.com/docker/cli/pull/4478

Thinking of a flow like:

```bash
$ docker scout quickview hello-world
Log in with your Docker ID or email address to use the scout service. If you
don't have a Docker ID, head over to https://hub.docker.com to create one.
You can log in with your password or a Personal Access  Token (PAT). Using
a limited-scope PAT grants better security and is required for organizations
using SSO. Learn more at https://docs.docker.com/go/access-tokens/

Username:
Password:
```


(we should consider changing `Password` to `Passphrase` or `Passphrase or PAT`, but probably should do so across the board)

laurazard:
@eunomie thoughts on @thaJeztah's idea?

eunomie:
I can see why we want to add more context about the `docker scout` commands.
To add `(login required)` might be OK, let me think a bit more about it. But we probably don't need the second line.

Right now, that's what you have if you run a scout command without to be logged in:

```
$ docker scout qv alpine:latest
ERROR   Status: login using Docker Desktop or 'docker login' command: no login, Code: 1
```

That's not a really good user experience. However I'm not sure about the addition of the login _inside_ docker scout.
I think the way it's working everywhere right now, is it's up to the user to login (CLI or desktop for instance).
So I don't want to include that logic here.

But to improve the login message 💯

eunomie:
I can change the `docker scout` login error that way:

image

But I wonder if it isn't too much details for this command, as this message it's more for `docker login`.

cli/mobycli/scout_suggest.go
```diff
@@ -53,10 +55,14 @@ func displayScoutQuickViewSuggestMsg(image string) {
 	}
 	out := os.Stderr
 	b := color.New(color.Bold)
-	_, _ = fmt.Fprintln(out)
-	_, _ = b.Fprintln(out, "What's Next?")
-	_, _ = fmt.Fprintf(out, "  View summary of image vulnerabilities and recommendations → %s", color.CyanString("docker scout quickview%s", image))
-	_, _ = fmt.Fprintln(out)
+
+	_, _ = b.Fprintln(out, "\nWhat's Next?")
```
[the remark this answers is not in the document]
laurazard: it was already there! :p

cli/mobycli/scout_suggest.go
```diff
@@ -74,3 +80,25 @@ func pulledImageFromArgs(args []string) string {
...
 	return image
 }
+
+// hubLoggedIn checks whether the user has credentials configured
+// for Docker Hub. This can be an expensive operation, so use it
+// mindfully.
+func hubLoggedIn() bool {
+	result := make(chan bool)
+	go func() {
+		hubAuth, err := config.LoadDefaultConfigFile(nil).GetAuthConfig("index.docker.io")
```
rumpl, line 90: You should use credential listing here, i.e. `config.LoadDefaultConfigFile(nil).GetCredentialsStore("").GetAll()` and then check if the map has the index.docker.io key.

A call to the `get` "endpoint" could force a login in the future

neersighted: Some users `docker login registry-1.docker.io`; we may want to check for the 'alternate' domains :joy:

cli/mobycli/scout_suggest.go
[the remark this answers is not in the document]
eunomie: Sorry, that's me 😇 
(I often prefer that to a `//nolint:errcheck` but maybe there's a better/cleaner way)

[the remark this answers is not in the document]
laurazard: I didn't even pay it any mind at the time, but now that I think about it it's probably a smell - we should be checking if it errors, since if an earlier message fails to write we probably don't want to send the subsequent ones.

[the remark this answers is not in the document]
thaJeztah: Yeah, these are annoying; you can configure the linter to ignore these specific ones in the config, or add `//nolint` tags, but that's noisy as well.

Using `_` is the most "correct" option, and it _does_ indicate _"no, this was not a mistake: I heard you complain about error-handling, and I decided to ignore you 😂"_

That said, it's really cumbersome; especially because a lot of printing is happening when you work on a CLI; I was exploring some alternatives in https://github.com/docker/cli/pull/3797 (think I have an  "approach 2" branch somewhere), but haven't found a "great" way yet.

cli/mobycli/scout_suggest.go
```diff
@@ -53,10 +55,14 @@ func displayScoutQuickViewSuggestMsg(image string) {
...
 	b := color.New(color.Bold)
-	_, _ = fmt.Fprintln(out)
-	_, _ = b.Fprintln(out, "What's Next?")
-	_, _ = fmt.Fprintf(out, "  View summary of image vulnerabilities and recommendations → %s", color.CyanString("docker scout quickview%s", image))
-	_, _ = fmt.Fprintln(out)
+
+	_, _ = b.Fprintln(out, "\nWhat's Next?")
+	if !hubLoggedIn() {
+		_, _ = fmt.Fprintf(out, "  View summary of image vulnerabilities and recommendations (login required) → %s\n", color.CyanString("docker scout quickview%s", image))
+		_, _ = fmt.Fprint(out, "  To sign in, use the `docker login` command or sign in via Docker Desktop\n")
```
dvdksn, line 62: Since we know that login is a prerequisite for quickview in this scenario, we could consider formatting this as a step list.

```suggestion
		_, _ = fmt.Fprintf(out, "  1. Sign in to your Docker account → %s\n", color.CyanString("docker login"))
		_, _ = fmt.Fprintf(out, "  2. View a summary of image vulnerabilities and recommendations → %s\n", color.CyanString("docker scout quickview%s", image))
```

dvdksn: (I omitted "or sign in using Docker Desktop" here because etoomuchprose - the shorter the better

sam-thibault: I like this a lot as a first step for 4.23 release.

thaJeztah: Yup, I like that suggestion as well.

I still think we should also explore the other options; ask for log-in if a command needs it), but we should look at that more broadly (including `docker pull`, `docker run`, `docker build` etc. if they try to pull (or push) an image and are missing credentials, or don't have the right scope)

/cc @jalonsogo

laurazard: image

dvdksn (approved):

cli/mobycli/scout_suggest.go
```diff
@@ -53,10 +56,14 @@ func displayScoutQuickViewSuggestMsg(image string) {
...
-	_, _ = b.Fprintln(out, "What's Next?")
-	_, _ = fmt.Fprintf(out, "  View summary of image vulnerabilities and recommendations → %s", color.CyanString("docker scout quickview%s", image))
-	_, _ = fmt.Fprintln(out)
+
+	_, _ = b.Fprintln(out, "\nWhat's Next?")
+	if !hubLoggedIn() {
+		_, _ = fmt.Fprintf(out, "  1. Sign in to your Docker account → %s\n", color.CyanString("docker login"))
+		_, _ = fmt.Fprintf(out, "  2. View a summary of image vulnerabilities and recommendations → %s\n", color.CyanString("docker scout quickview%s", image))
+	} else {
+		_, _ = fmt.Fprintf(out, "  View summary of image vulnerabilities and recommendations → %s\n", color.CyanString("docker scout quickview%s", image))
```
sam-thibault, line 65: Nit: "View summary" line 65 or "View a summary"?  line 63

sam-thibault (approved):

cli/mobycli/scout_suggest.go
thaJeztah, line 65: If we're nit-picking; can save some trees here 🌴 🌳 🌲 

```suggestion
	if !hubLoggedIn() {
		_, _ = fmt.Fprintln(out, "  1. Sign in to your Docker account → "+color.CyanString("docker login"))
		_, _ = fmt.Fprintln(out, "  2. View a summary of image vulnerabilities and recommendations → "+color.CyanString("docker scout quickview"+image))
	} else {
		_, _ = fmt.Fprintln(out, "  View a summary of image vulnerabilities and recommendations → "+color.CyanString("docker scout quickview"+image))
```

dvdksn: Ah, I overlooked the 2nd occurrence when I changed the first case to "a summary"

neersighted (approved):

cli/mobycli/scout_suggest.go
laurazard: Woops, I copy pasted @dvdksn's suggestion without looking properly 😬

rumpl (approved):

thaJeztah (approved):
LGTM

